PR #2, opened 2020-10-20
routing

naomigreen:
**What**  
Basic routing to enable form creation.

**Why**  
APIs aren't completed yet and we need to start creating the forms

**Anything else?**
n/a

### DIFF
--- a/pages/adult-form.js
+++ b/pages/adult-form.js
@@ -1,0 +1,20 @@
+import Head from 'next/head';
+import { useRouter } from 'next/router';
+
+const Adult = () => {
+  const router = useRouter();
+  return (
+    <div>
+      <Head>
+        <title>Adult</title>
+        <link rel="icon" href="/favicon.ico" />
+      </Head>
+      <button onClick={() => router.back()}>Back</button>
+      <div className="hero">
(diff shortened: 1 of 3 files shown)


pages/adult-form.js
```diff
@@ -0,0 +1,20 @@
+import Head from 'next/head';
+import { useRouter } from 'next/router';
+
+const Adult = () => {
+  const router = useRouter();
+  return (
+    <div>
+      <Head>
+        <title>Adult</title>
+        <link rel="icon" href="/favicon.ico" />
```
sirLisko, line 10: Not a big deal but I would probably move the `favicon` in the `_document` page, my fault that I copy and pasted the index from the `create-next-app` 😅

pages/adult-form.js
```diff
@@ -0,0 +1,20 @@
...
+      </Head>
+      <button onClick={() => router.back()}>Back</button>
+      <div className="hero">
```
sirLisko, line 13: These `classNames` are superfluous as they were defined in the `create-next-app`, we can get rid of them in the next iteration :)

sirLisko (approved):
Looks good, two small comments that are more a note to self for the next iteration.

Good to go 👍

PR #7, opened 2020-10-23
adult form and styling

naomigreen:
**What**  
**Forms steps**
Created Client details, Referral details and Case notes.
**Styling**
Updated the and styled the components
**Extra**
Changed the header component to the hackney component library.
Change favicon

This will be updated again once we decide how to generate forms.
CilentDetails
CaseNotes
ReferralDetails
 
**Why**  
Early stages of creating the form
**Anything else?**

- Have you added any new third-party libraries? lbh-frontend-react


components/Layout/Header/Header.jsx
```diff
@@ -1,25 +1,7 @@
-import HackneyLogo from './HackneyLogo';
+import { Header } from 'lbh-frontend-react';
```
sirLisko, line 1: Is this loading the whole library in the final bundle?

sirLisko (commented):
Looks good, only one question related to the `lbh-frontend-react`, if it's only loading the Header is a massive improvement 👍

components/Layout/Header/Header.jsx
naomigreen: Yeah I double checked the .next build and it's not.

PR #20, opened 2020-10-28
Add adult person view component

LBHABEBIC:
Add Adult person view component as per most recent designs.

image


components/PersonView/PersonView.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+const PersonView = ({ personId }) => {
+  const [person, setPerson] = useState();
+
+  const getPerson = async (personId) => {
+    try {
+      const data = await getResident(personId);
+      setPerson(data);
+    } catch {
+      console.log('error');
```
sirLisko, line 12: We need to handle this error, for example showing a custom `ErrorMessage`

components/PersonView/PersonView.jsx
```diff
@@ -0,0 +1,109 @@
...
+          <div className="govuk-grid-column-one-half">
+            <dt className="govuk-summary-list__key">Gender</dt>
+            <dd className="govuk-summary-list__value">{person?.gender}</dd>
+          </div>
+        </div>
+
+        <div className="govuk-grid-row">
+          <div className="govuk-grid-column-one-half">
+            <dt className="govuk-summary-list__key">
+              <h4>Address</h4>
```
sirLisko, line 72: Probably this shouldn't be `h4`

components/PersonView/PersonView.jsx
```diff
@@ -0,0 +1,109 @@
...
+              {new Date(person?.dateOfBirth).toLocaleDateString('en-GB')}
+            </dd>
+          </div>
+          <div className="govuk-grid-column-one-half">
+            <dt className="govuk-summary-list__key">NHS number</dt>
+            <dd className="govuk-summary-list__value">{person?.nhsNumber}</dd>
+          </div>
+        </div>
+
+        <div className="govuk-grid-row">
```
sirLisko, line 55: as it seems that is some cases the nationality is not present we can conditionally render this
```
{person?.nationality && <div className="govuk-grid-row">
```

components/PersonView/PersonView.jsx
```diff
@@ -0,0 +1,109 @@
...
+  useEffect(() => {
+    getPerson(personId);
+  }, [personId]);
+
+  return (
+    <>
+      <h1>View person</h1>
+      <h2 className="personDetails">PERSON DETAILS</h2>
+      <hr className="personDetailsHr" />
+      <dl className="govuk-summary-list--no-border   lbh-summary-list">
```
sirLisko, line 24: So we have `lbh-summary-list`? It seems we don't have any style attached to it 🤔

sirLisko (commented):
The functionality works as expected, I added a few notes.
Let's see what to do regarding the style, I'd go with the plain govuk one that also simplifies a lot the markup.

components/PersonView/PersonView.jsx
LBHABEBIC: Do you mean like this:  ```
[{person?.nationality && <div className="govuk-grid-row">}
          <div className="govuk-grid-column-one-half">
            <dt className="govuk-summary-list__key">Nationality</dt>
            <dd className="govuk-summary-list__value">{person?.nationality}</dd>
          </div>
        </div>](`url`)
```

File: components/PersonView/PersonView.jsx (hunk 4 above)
Comment (LBHABEBIC) [reply components/PersonView/PersonView.jsx:24]: Not sure...I copied it from lbh frontend. Take it out?

File: components/PersonView/PersonView.jsx (hunk 2 above)
Comment (LBHABEBIC) [reply components/PersonView/PersonView.jsx:72]: I made it h4 as anything bigger made things not align properly. Should it be bigger or smaller?

File: components/PersonView/PersonView.jsx (hunk 1 above)
Comment (LBHABEBIC) [reply components/PersonView/PersonView.jsx:12]: I can see that other components use <ErrorMessage label ={error} />  But I'm not clear where we define the error.

File: components/PersonView/PersonView.jsx (hunk 3 above)
Comment (sirLisko) [reply components/PersonView/PersonView.jsx:55]: Like this:
```
		{person?.nationality && (
          <div className="govuk-grid-row">
            <div className="govuk-grid-column-one-half">
              <dt className="govuk-summary-list__key">Nationality</dt>
                {person?.nationality}
              </dd>
            </div>
          </div>
        )}
```

components/PersonView/PersonView.jsx
sirLisko: We are not using the `lbh-frontend` library, for a series of reasons, we are referencing to https://design-system.service.gov.uk/components/.

components/PersonView/PersonView.jsx
sirLisko: The problem is the design, that is not compatible (out of the box) with our component library :(

components/PersonView/PersonView.jsx
LBHABEBIC: Ok I'll take the lbh things out.

components/PersonView/PersonView.jsx
sirLisko: Something like https://github.com/LBHackney-IT/lbh-social-care/blob/develop/components/Cases/Cases.jsx#L20.

The catch should intercept the error returned from the server `catch(e)` and handle it.

sirLisko (approved):
👍

PR #26, opened 2020-11-02
nationality list

naomigreen:
**What**  
Nationality dropdown

Screenshot 2020-11-02 at 09 54 30
Screenshot 2020-11-02 at 09 53 50

**Why**  
Field required

**Anything else?**


components/Form/Select/Nationality.jsx
```diff
@@ -0,0 +1,245 @@
...
+    <select
+      className="govuk-select"
+      id={name}
+      name={name}
+      ref={register}
+      aria-describedby={hint && `${name}-hint`}
+      onChange={(e) => onChange && onChange(e.target.value)}
+    >
+      <option value="">-- select one --</option>
+      <option value="british">British</option>
```
sirLisko, line 37: Can we have this in an external file? Like in a JSON

components/Form/Select/Nationality.jsx
```diff
@@ -0,0 +1,58 @@
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
+import formData from '../../../data/formData';
```
sirLisko, line 4: This is not pointing to any file 😅 
Also, we have `baseUrl` in place, so if a file is in the root folder you can just use `data/`, I think it's more clear :)

ps. sorry for being pedantic 😅

sirLisko (approved):
Tested and everything looks good 👍

PR #31, opened 2020-11-04
Task/align data cases table

LBHMVillalobos:
**What**  
- Data aligned to the new design
- Loading text replaced with the spinner component
- Jest snapshots have been updated

**Anything else?**

- Style will need to be reviewed at later stage.


LBHMVillalobos:
@sirLisko I have pushed the changes. Tests are passed now. Let me know if there is something else that does not look right please. Thanks for your help

sirLisko (approved):
Code looks good 👍 

PR #39, opened 2020-11-10
custom header

naomigreen:
**What**  
Change date format and updated to custom designs.
Screenshot 2020-11-10 at 18 33 26

**Why**  
New designs

**Anything else?**

- Anything else in this PR that isn't covered by the what/why? We don't have the GP data to display


components/Cases/CasesTable.jsx
```diff
@@ -2,20 +2,37 @@ import PropTypes from 'prop-types';
 
 const onClick = (url) => window.open(url, '_blank');
 
+const formatTime = (time) => {
```
sirLisko, line 5: Can we have this function in a `util` file?

components/Cases/CasesTable.jsx
```diff
@@ -2,20 +2,37 @@ import PropTypes from 'prop-types';
...
 const CasesEntry = ({
   formName,
   caseFormUrl,
   officerEmail,
   caseFormTimestamp,
 }) => (
   <tr className="govuk-table__row">
-    <td className="govuk-table__cell">{caseFormTimestamp}</td>
+    <td className="govuk-table__cell timestamp">
+      {formName && caseFormTimestamp && formatTime(caseFormTimestamp)}{' '}
```
sirLisko, line 24: Do we need to check `fornName`? 🤔

stylesheets/all.scss
```diff
@@ -32,3 +32,19 @@ $screen-sm-min: 576px;
 .govuk-input:disabled {
   background: govuk-colour('light-grey');
 }
+
+.govuk-table__cell.timestamp {
```
sirLisko, line 36: Not a big deal, I think we should stick with `BEM` notation for the classes :)

sirLisko (approved):
Looks good 👍

PR #48, opened 2020-11-17
added summary step

sirLisko:
**What**  
- added summary step to the forms

Screenshot 2020-11-17 at 12 15 28

**How to test it**
- check the `adult-referral` form
- after the last step there will be the Summary
- the submission has to work as before

### DIFF
--- a/components/Summary/Summary.jsx
+++ b/components/Summary/Summary.jsx
@@ -44,7 +44,10 @@ export const SummarySection = ({
       <h3 className="govuk-heading-m">{title}</h3>
       {Summary}
 
-      <Link href={id} as={`${formPath}${id}`}>
+      <Link
+        href={`${formPath}${id}?fromSummary=true`}
+        as={`${formPath}${id}?fromSummary=true`}
+      >
         <a className="govuk-link">Change</a>
       </Link>
     </div>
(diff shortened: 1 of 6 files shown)


naomigreen:
Looks great! Just confirming the best method to get back to the summary page.

components/Summary/Summary.jsx
```diff
@@ -0,0 +1,101 @@
...
+            ? Object.values(formData[name]).filter(Boolean).map(MultiValue)
+            : typeof formData[name] === 'boolean'
+            ? JSON.stringify(formData[name])
+            : formData[name],
+        }))}
+    />
+  );
+  return (
+    <div className="govuk-!-margin-bottom-7">
+      <h2>{title}</h2>
```
naomigreen, line 44: Could you add the govuk-fieldset__legend--m className, so we get the correct font

sirLisko: Good idea, but it's not a `fieldset`, let me see if I can find something similar 🤔

sirLisko: I am going to use `govuk-heading-m and making the element `h3` so it's more semantic 👍

naomigreen (approved):

PR #52, opened 2020-11-18
save function

naomigreen:
**What**  
Save data to local storage between steps 
Screenshot 2020-11-18 at 08 56 54

**Why**  
We have large forms 

**Anything else?**


utils/saveData.js
```diff
@@ -0,0 +1,17 @@
...
+export const saveData = (formPath, formData, step) => {
+  const data = {
+    step: step,
+    formData: formData,
+  };
+  windowGlobal.localStorage.setItem(formPath, JSON.stringify(data));
+};
+
+export const getData = (path) => {
+  return JSON.parse(windowGlobal.localStorage.getItem(path));
```
sirLisko, line 12: localStorage iterations should be around `try/catch` as it could be that the data is corrupted or the user has the localstorage disabled.

stylesheets/all.scss
```diff
@@ -48,3 +48,10 @@ $screen-sm-min: 576px;
 .govuk-divider {
   border: solid 2px #01674e;
 }
+li {
```
sirLisko, line 51: Are you sure you want to override the whole `li` selector?

data/forms/adult-referral.jsx
```diff
@@ -173,12 +173,6 @@ export default {
             'Community/Other route',
           ],
         },
-        {
-          component: 'Radios',
-          name: 'presentingIssueGen',
-          label: 'Presenting Issue',
-          options: ['Female', 'Male', 'Unknown', 'Other'],
-        },
```
sirLisko, line 181: Don't we need this? 🤔

components/FormWizard/FormWizard.jsx
```diff
@@ -66,7 +71,7 @@ const FormWizard = ({ formPath, formSteps, onFormSubmit, title }) => {
               <Breadcrumbs
                 key={step.id}
                 label={step.title}
-                link={`/form/adult-referral/${step.id}`}
+                link={`${formPath}${step.id}`}
```
sirLisko, line 74: Good catch 👍

components/DynamicStep/DynamicStep.jsx
```diff
@@ -34,7 +34,7 @@ const DynamicStep = ({
         </div>
         <Button
           className="govuk-button"
-          label="Next"
+          label="Save and Continue"
```
sirLisko, line 37: Are you sure we want to save on each step? I guess is more a product question.

components/FormWizard/FormWizard.jsx
```diff
@@ -7,12 +7,13 @@ import { useBeforeunload } from 'react-beforeunload';
 
 import DynamicStep from 'components/DynamicStep/DynamicStep';
 import Breadcrumbs from 'components/Breadcrumbs/Breadcrumbs';
+import { getData, saveData, deleteData } from 'utils/saveData';
 
 const FormWizard = ({ formPath, formSteps, onFormSubmit, title }) => {
   Router.events.on('routeChangeComplete', () => {
     window.scrollTo(0, 0);
   });
-  const [formData, setFormData] = useState({});
```
sirLisko, line 15: I would keep the formData as state rather than move it in the localstorage. It's more efficient to keep it at component level rather than having to get it each time from the localstorage.

components/SaveFormData/SaveFormData.jsx
```diff
@@ -0,0 +1,29 @@
...
+export const SavedForms = () => {
+  const windowGlobal = typeof window !== 'undefined' && window;
+  const data = Object.fromEntries(
+    Object.entries({ ...windowGlobal.localStorage }).filter(([key]) =>
+      key.includes('form')
+    )
+  );
+
+  return (
+    <>
```
sirLisko, line 13: We are rendering the whole container even if no incomplete forms are available.

data/forms/adult-referral.jsx
naomigreen: No this is a copy and paste error from gender.

stylesheets/all.scss
jaysonhunter: You should have a BEM class for the li tag and not style it directly.

jaysonhunter: You should add a BEM style class name to the li if you want to style it.

sirLisko (approved):
👍

PR #53, opened 2020-11-19
Child form

LBHABEBIC:
Added child FAST form. WIll need changes to be dynamic. Also ma need changes to the heading hierarchy. 

image


data/forms/child-referral.jsx
```diff
@@ -0,0 +1,565 @@
...
+      id: 'client-details',
+      title: 'Client Details',
+      components: [
+        {
+          component: 'TextInput',
+          name: 'worker_email',
+          width: '30',
+          label: 'Email address',
+          hint: 'Valid email address',
+          rules: { required: true },
```
sirLisko, line 15: We should check that is a correct email

data/forms/child-referral.jsx
```diff
@@ -0,0 +1,565 @@
...
+          name: 'position_agency',
+          width: '30',
+          label: 'Position / agency (if relevant): ',
+        },
+        {
+          component: 'TextInput',
+          name: 'referrer_telephone',
+          width: '30',
+          label: 'Telephone Number',
+          rules: { required: true },
```
sirLisko, line 69: We should validate that is a phone number, or at least that is a number.
Potentially we could also define the `type` as `tel` (you can find more info here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)

data/forms/child-referral.jsx
```diff
@@ -0,0 +1,565 @@
...
+      ],
+    },
+    {
+      id: 'requestForSupport',
+      title: 'Request For Support',
+      components: [
+        <h3 key="text" className="govuk-body">
+          Hackney Child Wellbeing Framework
+        </h3>,
+        <p key="text" className="govuk-body">
```
sirLisko, line 447: We should avoid having list elements with the same `key`. (reference https://reactjs.org/docs/lists-and-keys.html#keys)

data/forms/child-referral.jsx
```diff
@@ -0,0 +1,565 @@
...
+          Please consult the Hackney Child Wellbeing Framework
+          https://tinyurl.com/yxzq446e for guidance around levels of
```
sirLisko, line 449: Should this url be a link or it's fine as text?

components/FormWizard/FormWizard.jsx
```diff
@@ -75,7 +75,7 @@ const FormWizard = ({
                   <Breadcrumbs
                     key={step.id}
                     label={step.title}
-                    link={`/form/adult-referral/${step.id}`}
+                    link={`${formPath}${step.id}`}
```
sirLisko, line 78: 👍

sirLisko (commented):
It's looking good, only one thing we can improve is having proper validation/type for inputs that are numbers, emails, and phone numbers :)

data/forms/child-referral.jsx
LBHABEBIC: Does this mean every piece of stand alone text should have a separate key or we leave the key out?

data/forms/child-referral.jsx
LBHABEBIC: I will ask. It's like this in the google form.

data/forms/child-referral.jsx
Miles-Alford: Added new email input component, test & storybook

data/forms/child-referral.jsx
Miles-Alford: Added generic number component, test & storyboard. Used number type instead as tel type allows letters unless you use a regular expression and considering the different types of phone numbers it would be quite a large regex

data/forms/child-referral.jsx
Miles-Alford: Added unique key for each standalone text tag

data/forms/child-referral.jsx
Miles-Alford: used Link component, currently opens in the same window, will check with UX if it needs to be a new tab

sirLisko: `Link` component is only for internal links (because it's working in conjunction with the Router).
If this is meant to be a link I would have it open in a new tab.

sirLisko (changes requested):
I'd split the creation of the two new components in a standalone PR.

Also, the 2 fields should extend the `TextInput` not duplicating its code and they need to implement validation (ideally testing it).

components/Form/EmailInput/EmailInput.jsx
```diff
@@ -0,0 +1,27 @@
+import PropTypes from 'prop-types';
```
sirLisko, line 1: Sorry to be so picky but we can simplify a lot these components:

```
import { TextInput } from '..';

const EmailInput = (props) => (
    <TextInput {...props} type="email"></TextInput>
);

export default EmailInput;
```

the wrapper `<div className={cx('govuk-form-group')}>` is defined in `TextInput`, if we keep both it will render:
```
<div className="govuk-form-group">
  <div className="govuk-form-group">
```

Same thing for the `propTypes` that are already checked into the `TextInput`.

Miles-Alford: Done, was curious about the type checking for prop types if it needed to be done in both components, thanks for the clarification.

sirLisko: if you check the warning in the console you will have this:
```
Failed prop type: The prop `label` is marked as required in `TextInput`, but its value is `undefined`.
    in TextInput (at PhoneInput.jsx:8)
    in div (at PhoneInput.jsx:7)
    in PhoneInput (at DynamicInput.jsx:29)
    in DynamicInput (at DynamicStep.jsx:17)
    in div (at DynamicStep.jsx:14)
    in form (at DynamicStep.jsx:13)
    in DynamicStep (at FormWizard.jsx:93)
    in fieldset (at FormWizard.jsx:60)
    in div (at FormWizard.jsx:53)
    in FormWizard (at [stepId].jsx:13)
    in ChildReferral (at _app.jsx:28)
    in main (at Layout/index.jsx:14)
    in div (at Layout/index.jsx:12)
    in Layout (at _app.jsx:26)
    in MyApp
    in ErrorBoundary (created by ReactDevOverlay)
    in ReactDevOverlay (created by Container)
    in Container (created by AppContainer)
    in AppContainer
    in Root
```

Where you can see the stack error, as we are passing all the props directly from the `PhoneInput` to the `TextInput` one check is enough.
Let me know if you have any more question :)

sirLisko (approved):
Good one 👏

PR #58, opened 2020-11-25
add conditional redering

LBHABEBIC:
**What**  
Add conditional rendering for child FAST form so that it matches google form.

**Anything else?**
Some sections need to be reviewed by UX colleagues as the 'Consent' section for example it's unclear what should happen if the person says no to the first question. 

image


sirLisko (approved):
Code looks good 👍 

PR #61, opened 2020-11-26
Feature/add multi step

LBHABEBIC:
Added multi step and removed  components which are no longer necessary. 

image


sirLisko (approved):
Looks good 👍

PR #67, opened 2020-12-03
Hotfix: Login Redirect Fix

Miles-Alford:
**What**  
When a user log's in to the site the first login will return Forbidden and will not show the application. If a user refreshes the page they can then enter the application. This PR changes the redirect URL on a successful authentication to go to the correct URL

**Why**  
On a successful login the application would redirect to a forbidden AWS resource (API Gateway) on the first login thus stopping users from accessing the application until they refresh the page. This PR fixes that issue.
**Anything else?**
Within your env file a new variable needs to be added called REDIRECT_URL, value should be the local environment url


sirLisko (approved):
👍

PR #69, opened 2020-12-08
form search

naomigreen:
**What**  

- Form search page to locate forms
- Expand and collapse person details 
- Redirect between application and google forms

Screenshot 2020-12-08 at 19 23 48
Screenshot 2020-12-08 at 19 23 05
Screenshot 2020-12-08 at 19 23 25

**Why**  
Feature to locate forms

**Anything else?**


components/PersonView/PersonDetails.jsx
```diff
@@ -10,68 +11,99 @@ const PersonDetails = ({
...
+        {expandView && !expandDetails ? (
+          <span className="govuk-expand-toggle" onClick={handleToggle}>
+            Expand view
+          </span>
+        ) : null}
+        {expandView && expandDetails ? (
+          <span className="govuk-expand-toggle" onClick={handleToggle}>
+            Collapse view
+          </span>
+        ) : null}
```
sirLisko, line 38: We can simplify this:

```suggestion
        {expandView && (
          <span
            className="govuk-expand-toggle"
            onClick={() => setExpandDetails(!expandDetails)}
          >
            {expandDetails ? 'Expand' : 'Collapse'} view
          </span>
        )}
```

pages/record/[id]/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+        Add a new record for
+      </h1>
+      <PersonView personId={query.id} expandView={true} nameSize="m" />
+      <p className="govuk-label govuk-!-margin-top-7 govuk-!-margin-bottom-5">
+        Use forms to create a new record for a person
+      </p>
+      <Select
+        name="formList"
+        options={FORM_LIST}
+        register={register}
```
Comment (sirLisko) [inline pages/record/[id]/index.jsx:27]: Why not using a simple `onChange` instead of using `useForm` with a `watch`?

`onChange={(value) => setValue(value)}`

stylesheets/all.scss
```diff
@@ -43,3 +43,34 @@ $screen-sm-min: 576px;
 .govuk-warning {
   color: #da3711;
 }
+
+.govuk-link-button {
+  min-width: 210px;
+  padding: 10px 0;
+}
+
+.govuk-record-header {
```
sirLisko, line 52: We shouldn't use `.govuk-` prefix, otherwise we can get confused about what it's from the govuk file system and what's not.

For styling can we use css modules or css-in-js, instead of global css? I am fine with both

pages/record/[id]/index.jsx
```diff
@@ -0,0 +1,45 @@
+import { NextSeo } from 'next-seo';
```
Comment (sirLisko) [inline pages/record/[id]/index.jsx:1]: Shouldn't this page be `pages/people/[id]/record.jsx`?

components/LinkButton/LinkButton.jsx
```diff
@@ -1,31 +1,24 @@
...
     <button
       href="#"
       role="button"
       draggable="false"
-      className="govuk-button govuk-button--start govuk-!-margin-top-3"
+      className={`govuk-button govuk-link-button govuk-!-margin-bottom-3`}
       data-module="govuk-button"
-      onClick={() => Router.push(`${route}`)}
+      onClick={() =>
+        isGoogleForm ? handleLink(route) : Router.push(`${route}`)
```
sirLisko, line 15: We are mixing business logic with presentational logic.
Can we have a prop like `isExternal` driving it we want to open the link in an external page?

sirLisko (commented):
Added a few comments, we can simplify even more the "toggle" logic, I have a few ideas, we can discuss it, it's going to be easier :)

data/formList.js
```diff
@@ -0,0 +1,24 @@
+export default [
+  { text: 'Adult Referral', value: '/form/adult-referral/client-details' },
```
sirLisko, line 2: Not sure this one should be here 🤔 
At the moment it's the "create new user form"

components/LinkButton/LinkButton.jsx
```diff
@@ -1,31 +1,22 @@
 import PropTypes from 'prop-types';
 import Router from 'next/router';
 
-const LinkButton = ({ label, route }) => (
-  <div>
+const LinkButton = ({ label, route }) => {
+  const handleLink = (url) => window.open(url, '_blank');
+  const isExternal = route && route.includes('google');
```
sirLisko, line 6: I would check if it's a relative or absolute URL rather than checking if contains "google".

```suggestion
  const isExternal = route && route.includes('https://');
```

sirLisko (commented):
Looks good, only one small thing

sirLisko (approved):

PR #75, opened 2020-12-11
Feature/edit  asc case notes

LBHABEBIC:
Changed ASC case note form so that it uses form wizard. Getting a 404 not sure why.


components/Form/TextArea/TextArea.jsx
```diff
@@ -16,7 +16,7 @@ const TextArea = ({
     })}
   >
     {label && (
-      <label className={`govuk-label govuk-label--${labelSize}`} for={name}>
+      <label className={`govuk-label govuk-label--${labelSize}`} htmlFor={name}>
```
sirLisko, line 19: Good catch 👍

data/forms/case-notes-recording.jsx
```diff
@@ -0,0 +1,46 @@
...
+          component: 'TextInput',
+          name: 'caseNoteTitle',
+          width: '30',
+          label: 'Case Note Title',
+        },
+        {
+          component: 'TextArea',
+          name: 'caseNoteDescription',
+          width: 80,
+          labelSize: 'm',
```
sirLisko, line 40: we can have this (`labelSize: 'm'`) as a default prop in `TextArea` component, like we do for `TextInput`

components/Form/TextArea/TextArea.jsx
```diff
@@ -37,6 +37,7 @@ const TextArea = ({
       id={name}
       name={name}
       rows="5"
+      width="30"
```
sirLisko, line 40: The `width` is not working, we can use `govuk-input govuk-input--width-${width}`.

sirLisko (commented):
Two small things, but it looks good :)

sirLisko (approved):
Looks good 👍

PR #77, opened 2020-12-14
 N and Y values address fix

naomigreen:
**What**  
Filtered out the N and Y values in the address table 
Screenshot 2020-12-14 at 13 18 01

**Why**  

**Anything else?**

### DIFF
--- a/components/PersonView/PersonDetails.jsx
+++ b/components/PersonView/PersonDetails.jsx
@@ -14,6 +14,9 @@ const PersonDetails = ({
   expandView = false,
 }) => {
   const [expandDetails, setExpandDetails] = useState(false);
+  const address =
+    addressList &&
+    addressList.find((address) => address.contactAddressFlag === 'Y');
   return (
     <>
       <div>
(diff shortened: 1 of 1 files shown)


components/PersonView/PersonDetails.jsx
```diff
@@ -63,9 +63,9 @@ const PersonDetails = ({
                 <dt className="govuk-summary-list__key">Address</dt>
                 <dd className="govuk-summary-list__value">
                   {addressList &&
-                    Object.values(addressList[0]).map((addressLine) => (
-                      <p>{addressLine}</p>
-                    ))}
+                    Object.values(addressList[0])
```
sirLisko, line 66: we should filter the `addressList` and get the one with value `Y` and then just show `addressLine1`.

sirLisko (changes requested):

components/PersonView/PersonDetails.jsx
```diff
@@ -62,10 +62,12 @@ const PersonDetails = ({
               <div className="govuk-summary-list__row">
                 <dt className="govuk-summary-list__key">Address</dt>
                 <dd className="govuk-summary-list__value">
-                  {addressList &&
-                    Object.values(addressList[0]).map((addressLine) => (
-                      <p>{addressLine}</p>
-                    ))}
+                  <p>
+                    {addressList &&
+                      addressList.filter(
```
sirLisko, line 67: [Picky mode ON]
In this case we can use `find` instead of filter, as there will be only one and the loop is interrupting when it found it 😅 
[Picky mode OFF]

sirLisko (approved):
Looks good 👍

PR #83, opened 2020-12-16
in progress fix

naomigreen:
**What**  
In progress fix and search label fix 

**Why**  

**Anything else?**

### DIFF
--- a/components/Layout/Header/Header.jsx
+++ b/components/Layout/Header/Header.jsx
@@ -4,6 +4,8 @@ import Link from 'next/link';
 import { useRouter } from 'next/router';
 
 import { useContext } from 'react';
+import { getDataIncludes } from 'utils/saveData';
+
 import UserContext from 'components/UserContext/UserContext';
 import Logo from './Logo.jsx';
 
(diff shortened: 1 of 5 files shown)


components/Search/Search.jsx
```diff
@@ -118,7 +118,8 @@ const Search = ({ query, type }) => {
             <>
               <div className="lbh-table-header">
                 <h2 className="govuk-fieldset__legend--m govuk-custom-text-color">
-                  {type.toUpperCase()} SEARCH RESULT
+                  {type === 'cases' ? 'RECORDS' : type.toUpperCase()} SEARCH
```
sirLisko, line 121: Can we rename the `type` instead?

naomigreen: We're using type for the people label so we need to keep it 
Screenshot 2020-12-16 at 10 28 03

sirLisko: yes, what I meant is that we are defining `type="cases"` in https://github.com/LBHackney-IT/lbh-social-care/blob/create_new_user/pages/cases/index.jsx#L9. We should make that `records`.

pages/cases/index.jsx
```diff
@@ -6,7 +6,7 @@ const SearchCasesPage = ({ query }) => {
   return (
     <div>
       <NextSeo title="Find unlinked case notes" noindex />
-      <Search query={query} type="cases" />
+      <Search query={query} type="Records" />
```
sirLisko, line 9: Can we have `records` lowercase so we are consistent.
You also need to change the checks `type === "cases"` in the `Search` component.

components/Layout/Header/Header.jsx
```diff
@@ -4,8 +4,9 @@ import Link from 'next/link';
 import { useRouter } from 'next/router';
 
 import { useContext } from 'react';
-import UserContext from 'components/UserContext/UserContext';
 import { getDataIncludes } from 'utils/saveData';
+
+import UserContext from 'components/UserContext/UserContext';
```
sirLisko, line 9: Can we roll back these changes that are confusing? Same for the `_app` component

sirLisko (changes requested):

sirLisko (approved):
👍

PR #90, opened 2020-12-17
Add cfs case notes form

LBHABEBIC:
image

image


pages/people/[id]/index.jsx
```diff
@@ -20,7 +21,7 @@ const CasesPage = ({ query }) => {
           </p>
         </div>
         {/* commented out as the feature is not ready to be in prod */}
-        {/* <LinkButton label="Add a new record" route={`${query.id}/record`} /> */}
+        <LinkButton label="Add a new record" route={`${query.id}/record`} />
```
Comment (sirLisko) [inline pages/people/[id]/index.jsx:24]: we should still hide this

pages/form/csc-case-note recording/[...stepId].jsx
```diff
@@ -0,0 +1,43 @@
...
+import BackButton from 'components/Layout/BackButton/BackButton';
+import FormWizard from 'components/FormWizard/FormWizard';
+import { addCase } from 'utils/api/cases';
+import form from 'data/forms/csc-case-notes-recording';
+import PersonView from 'components/PersonView/PersonView';
+
+const CscCaseNotesRecording = ({ query }) => {
+  const onFormSubmit = async (formData) => {
+    const ref = await addCase(formData.mosaic_id, {
+      caseFormData: JSON.stringify(formData),
```
Comment (sirLisko) [inline pages/form/csc-case-note recording/[...stepId].jsx:12]: we need to send along the person and user information as we did for the other form

sirLisko (commented):
I think you mentioned but I forgot 😅 
We are waiting for the select values to be provided?
We also need to check with the BE/DB what input names we want to use.

pages/form/csc-case-note recording/[...stepId].jsx
Comment (LBHABEBIC) [reply pages/form/csc-case-note recording/[...stepId].jsx:12]: > I think you mentioned but I forgot 😅
> We are waiting for the select values to be provided?
> We also need to check with the BE/DB what input names we want to use.

Yes...still working on this . the dropdowns are massive ...about 10 conditional components based on what they choose in the first select. Joy also told me late this afternoon that she is changing things so who knows.

sirLisko (approved):
Looks good, we need to check with BE about the values that they are expecting, but it's good enough to be merged :)

PR #91, opened 2020-12-17
In progress design update

naomigreen:
**What**  

- Design changes to form in progress page. Displaying two tables, one containing user details and the other without
- Removed the title from the form layout as the designs have changed but the form name is still required
- Had to centre the spinner 
Screenshot 2020-12-21 at 13 49 56

**Why**  

**Anything else?**


components/SaveFormData/SaveFormData.jsx
```diff
@@ -1,25 +1,66 @@
...
-    </ul>
+    <>
+      <p className="govuk-fieldset__legend--s gov-weight-lighter">
+        {`Displaying ${formQty} unfinished ${formQty > 1 ? 'forms' : 'form'}`}{' '}
+      </p>
+      <table className="govuk-table">
+        <thead className="govuk-table__head">
+          <tr className="govuk-table__row">
+            {tableHeader.map(({ id, text }) => (
+              <th key={id} className="govuk-table__header">
```
sirLisko, line 39: Not a big deal but here we can use directly the `text` as `key`, so the `tableHeader` can become a list of strings.

stylesheets/all.scss
```diff
@@ -62,3 +62,16 @@ $screen-sm-min: 576px;
   justify-content: space-between;
   align-items: flex-end;
 }
+
+.cancelIcon {
```
sirLisko, line 66: Can we avoid using global style for component-specific styling?

Css modules or css-in-js are better.

sirLisko (commented):
Tested and looks good 👍 
Just one small comment about styling.

LBHABEBIC (approved):

data/forms/case-notes-recording.jsx
```diff
@@ -1,7 +1,8 @@
 import CASE_NOTE_TYPES from 'data/caseNoteTypes';
 export default {
-  title: '',
+  title: 'Case Notes Recording',
   path: '/form/case-notes-recording/',
+  includeDetails: true,
```
sirLisko, line 5: we can have this in the `FormCasesWrapper`, as it has been created for the purpose to wrap all the cases form.

components/SaveFormData/Tables.jsx
```diff
@@ -0,0 +1,79 @@
...
+        {tableHeader.map((text) => (
+          <th key={text} className="govuk-table__header">
+            {text}
+          </th>
+        ))}
+      </tr>
+    </thead>
+    <tbody className="govuk-table__body">
+      {data.map((cell) => (
+        <tr key={'key'} className="govuk-table__row">
```
sirLisko, line 59: the `key` hardcoded is firing warnings in the console

components/SaveFormData/SaveFormData.jsx
```diff
@@ -1,25 +1,58 @@
...
+    'Last saved',
+    'Actions',
+    '',
+  ];
+  const standardHeader = ['Form type', 'Last saved', 'Actions', ''];
+  const sortData = Object.values(savedForms);
+  const formQty = sortData.length;
+  const detailData = sortData.filter((item) => item.includesDetails);
+  const standardData = sortData.filter((item) => !item.includesDetails);
+  console.log(sortData);
```
sirLisko, line 36: you forgot `console.log`

utils/saveData.js
```diff
@@ -1,9 +1,23 @@
...
+export const saveData = (
+  formPath,
+  data,
+  title,
+  step,
+  includesDetails,
+  personDetails
+) => {
+  const timeStamp = new Date(Date.now()).toLocaleString().split(',')[0];
+  console.log(step);
```
sirLisko, line 12: you forgot a `console.log`

utils/saveData.js
```diff
@@ -1,9 +1,23 @@
 export const windowGlobal = typeof window !== 'undefined' && window;
 
-export const saveData = (formPath, data, step) => {
+export const saveData = (
+  formPath,
+  data,
+  title,
+  step,
+  includesDetails,
```
sirLisko, line 8: Minor thing but I think we need to deconstruct all these arguments, we can get rid of a lot of duplication 🤔

stylesheets/all.scss
```diff
@@ -62,3 +62,16 @@ $screen-sm-min: 576px;
...
+
+.cancel-icon {
+  width: 20px;
+  height: 20px;
+  top: 5px;
+  position: relative;
+  cursor: pointer;
+}
+
+.loading-spinner {
```
sirLisko, line 74: Can we avoid to have a class in `all.scss` if it's used in only one place?

components/FormWizard/FormWizard.jsx
```diff
@@ -109,7 +106,12 @@ const FormWizard = ({
             saveData(
               formPath,
               updatedData,
-              `${stepId.join('/')}?${getQueryString(queryString)}`
+              title,
+              includesDetails
```
sirLisko, line 110: We should keep the query string even if the form doesn't have `includesDetails`.

sirLisko (commented):
Looks good, few minor comments.

Probably we can avoid code duplication even further in `saveFormData/Tables` but I guess it's not a big deal for the moment, we can revise it if we need to add a new table :)

components/SaveFormData/SaveFormData.jsx
naomigreen: removed but forgot to save -_-

components/FormWizard/FormWizard.jsx
naomigreen: We're getting an error if we do

stylesheets/all.scss
```diff
@@ -62,3 +62,16 @@ $screen-sm-min: 576px;
...
+
+.cancel-icon {
+  width: 20px;
+  height: 20px;
+  top: 5px;
+  position: relative;
+  cursor: pointer;
+}
+
+#loading-spinner {
```
sirLisko, line 74: The problem is not the class, it's the global file and at global level 😅 
We should have a discussion about styling and pick if we prefer going with css modules or css-in-js and stick with that. So we can keep this file for generic rules and override the govuk one.

naomigreen: Since you've added your desired change, I've removed this https://github.com/LBHackney-IT/lbh-social-care/blob/573093a5f1a138866a7d99b60ca286078a7d548c/components/Spinner/Spinner.jsx#L8

PR #94, opened 2020-12-18
Improvements on DateInput

sirLisko:
**What**  
- prevent annoying warning message of uncontrolled controller (see [this discussion](https://github.com/react-hook-form/react-hook-form/discussions/3653) for more info)
- fixed dateInput ref (on error it should focus on the date input)
- force the date year to 4 digits (preventing `12-21-13` to be flag as valid date)

### DIFF
--- a/components/Form/AddressLookup/AddressLookup.jsx
+++ b/components/Form/AddressLookup/AddressLookup.jsx
@@ -160,6 +160,7 @@ const AddressLookup = ({
             ...rules?.validate,
           },
         }}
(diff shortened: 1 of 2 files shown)


LBHABEBIC (approved):
Looks fine. I will have to investigate what forward ref does .

sirLisko:
@LBHABEBIC basically you can't attach `ref` to a functional component, because of that needs to be wrapped by `forwardRef`, happy to go more in the details if it's not clear :)

PR #99, opened 2020-12-21
expose all available phone numbers

LBHABEBIC:
image

### DIFF
--- a/components/PersonView/PersonDetails.jsx
+++ b/components/PersonView/PersonDetails.jsx
@@ -80,9 +80,13 @@ const PersonDetails = ({
                 <div className="govuk-summary-list__row">
                   <dt className="govuk-summary-list__key">Phone Number</dt>
                   <dd className="govuk-summary-list__value">
-                    {Object.values(phoneNumber[0]).map((number) => (
-                      <p key={number}>{number}</p>
-                    ))}
+                    <ul>
+                      {phoneNumber.map(({ phoneNumber, phoneType }) => (
+                        <li>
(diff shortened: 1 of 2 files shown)


components/PersonView/PersonDetails.jsx
```diff
@@ -80,9 +80,13 @@ const PersonDetails = ({
                 <div className="govuk-summary-list__row">
                   <dt className="govuk-summary-list__key">Phone Number</dt>
                   <dd className="govuk-summary-list__value">
-                    {Object.values(phoneNumber[0]).map((number) => (
-                      <p key={number}>{number}</p>
-                    ))}
+                    <ul>
+                      {phoneNumber.map(({ phoneNumber, phoneType }) => (
+                        <li className="govuk-list" key={phoneNumber}>
```
sirLisko, line 85: `govuk-list` should be on the `ul`, see reference at https://design-system.service.gov.uk/styles/typography/

naomigreen (approved):

naomigreen:
Looks great! Just Luca's comment to update

components/PersonView/PersonDetails.jsx
LBHABEBIC: done now. doesn't seem to make any difference.

sirLisko (approved):
Looks good 👍

components/PersonView/PersonDetails.jsx
sirLisko: yes, it's more because it's cleaner to set it to the parent rather than in all the children :)

PR #100, opened 2020-12-22
Feature/add allocated worker

LBHABEBIC:
Add allocated workers component and tests. Only currently visible with 44000000
image


sirLisko (approved):
Code looks good 👍 
Tested and it works as expected (with mocked data 😅 ) 🚀

PR #115, opened 2021-01-06
Feature/deallocate worker

Miles-Alford:
**What**  
First part to allow workers to be deallocated from a person, this PR adds the modal to deallocate a worker with a reason
**Why**  
To allow workers to be deallocated from a person.

**Anything else?**
This is the first part to allowing a worker to be deallocated, the second part is wiring up the API to the modal to fully complete the feature
image


components/AllocatedWorkers/AllocatedWorkersTable.jsx
```diff
@@ -1,4 +1,8 @@
+import Modal from 'components/Modal/Modal';
 import PropTypes from 'prop-types';
+import Button from '../Form/Button/Button';
+import { useState } from 'react';
```
sirLisko, line 4: It's not changing much, but by convention, the external dependencies (i.e. `react`, `prop-types`) should be at the top.

components/AllocatedWorkers/AllocatedWorkersTable.jsx
```diff
@@ -7,58 +11,82 @@ const AllocatedWorkersEntry = ({
...
-);
+  openModal,
+}) => {
+  return (
+    <>
+      <div className="lbh-table-header">
+        <h3 className="govuk-fieldset__legend--m govuk-custom-text-color govuk-!-margin-top-0">
+          Allocated Worker {index + 1}
+        </h3>
+        {<Button isSecondary label="Deallocate Worker" onClick={openModal} />}
```
sirLisko, line 22: for the meantime, we should only show this for `admin` users

components/AllocatedWorkers/AllocatedWorkersTable.jsx
```diff
@@ -68,6 +96,8 @@ AllocatedWorkersTable.propTypes = {
       allocatedTeam: PropTypes.string,
       startDate: PropTypes.number,
       endDate: PropTypes.number,
+      id: PropTypes.number,
+      mosaicId: PropTypes.number,
```
sirLisko, line 100: are we using these 2 props?

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
```diff
@@ -0,0 +1,97 @@
+import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
+import { Button, TextArea } from 'components/Form';
+import Spinner from 'components/Spinner/Spinner';
+import { useState } from 'react';
+import { useForm } from 'react-hook-form';
+import Link from 'next/link';
```
sirLisko, line 6: Same as above, 3rd party libs on top (I remember that there was an eslint plugin to do that automatically, I should check this)

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
```diff
@@ -0,0 +1,97 @@
...
+    setLoading(true);
+    try {
+      const test = JSON.stringify(Object.assign(patchData, data));
+      await deleteAllocatedWorkers(personId, test);
+    } catch {
+      <ErrorMessage label="Something went wrong, please retry"></ErrorMessage>;
+      console.log('Something went wrong');
+    }
+    setLoading(false);
+    setComplete(true);
```
sirLisko, line 34: The `setComplete` should be inside the `try` otherwise you are showing the "complete message" also if the call fails

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+  return (
+    <form onSubmit={handleSubmit(onSubmit)}>
+      <h1 className="govuk-heading-l">Reason for worker deallocation</h1>
```
sirLisko, line 38: Not sure if this would have a different page scope, but just to be sure I would change it to a `h2` 🤔 
We should investigate this

pages/people/[id]/allocated-worker/[workerId]/delete.jsx
```diff
@@ -0,0 +1,23 @@
+import Button from 'components/Form/Button/Button';
+import Textarea from '../../../../../components/Form/TextArea/TextArea';
```
Comment (sirLisko) [inline pages/people/[id]/allocated-worker/[workerId]/delete.jsx:2]: we don't need this page anymore

sirLisko (commented):
Looks really good, just a couple of minor things and then it's good to go :)

sirLisko (approved):
Code looks good 👍 

Good one 👏

PR #119, opened 2021-01-11
decoupled allocatedWorkers from personView

sirLisko:
**What**  
`AllocatedWorkers` are now part of the `PersonView`, this is wrong as we are showing them also in the `records` page (with the `PersonDetails` toggle)
This PR fixes this problem plus:
- fixed the problem of the double spinner
- if the person API fails the cases and allocations stay hidden
- the cases/allocation are not appearing before the person details
- fixed route for `/people/:id/records` to make more consistent


LBHABEBIC (approved):

LBHABEBIC (commented):
i can't see where you removed the spinner. is this all of the changes?

sirLisko:
@LBHABEBIC the spinner is still there, and it's shown if the person API is faster.
As soon as the person's data is back from APIs the 3 components are displayed if the `Cases` are not there yet the spinner is shown.

PR #127, opened 2021-01-15
Refactored authentication

sirLisko:
**What**  
- all the pages have now access to the auth, without the need to be server-rendered even in clien-side navigation (no need to have `getServerSideProps` empty to access user)
- `/api/check-auth` is returning the auth object, so all the authentication logic is decoupled and lead by the BE. It's returning `UNAUTHORIZED` if the user is not logged, `FORBIDDEN` if logged but not authorised, and `OK` with the user details if everything it's correct. It will also make it easier to understand the user permissions
- `useAuth()` is returning directly the data stored in the `UserContext` so `user`, `isLoading`, etc.
- simplified the `isAuthorised` function
- redirects are now centralised into the `AuthProvider` and not spread across different `getServerSideProps/getInitialProps`
- removed the necessity to save the user state in `_app`
- removed the `getInitialProps` from `_app`, this is also solving 👇 
- we don't have anymore the warning during build time
  > Warning: You have opted-out of Automatic Static Optimization due to `getInitialProps` in `pages/_app`. This does not opt-out pages with `getStaticProps`
Read more: [https://err.sh/next.js/opt-out-auto-static-optimization](https://err.sh/next.js/opt-out-auto-static-optimization)
- all the pages that don't need server-side code will get the benefits of [automatic-static-optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization)

**Caveats**
When the pages are statically generated the first render will set the `Router.query` to `{}`, after the hydration the query will be populated with the correct values. To solved this problem we are using `Router.isReady` (https://github.com/LBHackney-IT/lbh-social-care/pull/127/files#diff-eb4e5aa2ef02c9e38944a6f5123cfecd65aa8d83d645926f4ae615084ef23cbfR32) to wait until the query is populated before showing the page components. `Router.isReady` has been introduced by Next https://github.com/vercel/next.js/releases/tag/v10.0.5, because of that we need #125.

**How to test it?**


LBHABEBIC (approved):
Makes sense and all looks good.

naomigreen (approved):

naomigreen (approved):

PR #129, opened 2021-01-15
aligned AddAllocationWorker to the BE

sirLisko:
**What**  
- aligned post body to BE
- added POST params validation, using [`yup`](https://github.com/jquense/yup)
- refresh allocations on add allocation success
- changed admin (or both adult/child) permission to `B`
- improved `POST` status codes and error messaging


components/AllocatedWorkers/AddAllocatedWorker/AddAllocatedWorker.jsx
```diff
@@ -52,19 +56,23 @@ const AddAllocatedWorker = ({ personId, currentlyAllocated }) => {
     setIsModalOpen(false);
     setWorkers();
   });
-  const addWorker = useCallback(async (formData) => {
+  const addWorker = useCallback(async ({ worker }) => {
```
Miles-Alford, line 59: Rather basic react question, never really used `useCallback`, what does it exactly do?, from the docs it seems to help with optimisation and avoiding unnecessary renders

utils/server/allocatedWorkers.js
```diff
@@ -34,13 +35,16 @@ export const deleteResidentAllocatedWorker = async (id, body) => {
...
-    }
-  );
+const addAllocatedWorkerSchema = yup.object().shape({
+  mosaicId: yup.number().required().integer(),
+  allocatedWorkerId: yup.number().required().integer(),
+  allocatedBy: yup.string().email().required(),
+});
+
+export const addAllocatedWorker = async (mosaicId, params) => {
+  const body = await addAllocatedWorkerSchema.validate({ mosaicId, ...params });
```
Miles-Alford, line 45: Never seen `yup` before, just confirming this line validates if the request body matches the defined schema above correct?

sirLisko: Yes, in this case it's solving two problems, validation and casting of the values.
The BE is expecting `ids` as number, but we are getting back as `string`, `yup` is checking that the string contains a number and if it's so it is transforming it in an integer, making the BE happy.
As we are only using it in the BE side we shouldn't worry about the size of the lib, so I thought it was a nice to have and a good feedback for the devs.

components/AllocatedWorkers/AddAllocatedWorker/AddAllocatedWorker.jsx
sirLisko: Every time something change in a component (i.e. state, useEffects) the whole component is rerun.
You use `useCallback` if you want to "store" the definition of the function, so you don't need to reexecute its definition on each rerun.
As you said is an extra optimisation, it really depends how many times you think that component might render. In this case, as the user can click on several sections it could happen quite a few times, so `useCallback` is a nice to have

Miles-Alford (approved):
All makes sense, thanks for answering my question!

PR #130, opened 2021-01-15
google forms

naomigreen:
**What**  
To test please change the cookie to children but note that the full name is "Admin Children" in the user. So it will populate that username in the forms. To see your name change getUserType(user) === 'Children' to getUserType(user) === 'Admin'
Form
Adult
ChildForms

**Why**  

**Anything else?**

### DIFF
--- a/components/AddForm/AddForm.jsx
+++ b/components/AddForm/AddForm.jsx
@@ -1,0 +1,86 @@
+import { useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+
+import LinkButton from 'components/LinkButton/LinkButton';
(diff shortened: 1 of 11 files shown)


components/AddForm/AddForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+    getPerson();
+  }, []);
+  return (
+    <>
+      {error && <ErrorMessage />}
+      {person && (
+        <>
+          <div>
+            <Select
+              style={{ display: ' inline-block' }}
```
sirLisko, line 45: we can use the grid system for this https://design-system.service.gov.uk/styles/layout/#grid-system

sirLisko (approved):
Tested and it works as expected 👍

PR #135, opened 2021-01-19
reintroduced getInitialProps for auth in the app

sirLisko:
**What**  
- Reintroduced `getInitialProps` in `_app` as static served pages have problems with `NEXT_PUBLIC_*` env variables
- Used the cookie directly to populate the `user`, no need to wait for the API `/check-auth`
- Left the `/check-auth` for debugging


naomigreen (approved):
Frontend tested and working

PR #139, opened 2021-01-20
Wire deallocation functionality to backend

Miles-Alford:
**What**  
This PR amends the `deleteAllocationWorker` to actually deallocate a worker instead of return mock data, some changes to the deallocation modal has also been made, mainly to show the deallocation user what they wrote for the deallocation reason after they confirm the deallocation.

**Why**  
This is to allow workers to be deallocated from a person

**Anything else?**

- The confirmation modal links are to be made functional at a later date, kept them in for now but if they wished to be hidden then can do that instead


utils/server/allocatedWorkers.js
```diff
@@ -20,19 +20,17 @@ export const getResidentAllocatedWorkers = async (mosaic_id, params) => {
...
-    setTimeout(() => resolve(console.log('data is ' + id + body)), 3000)
-  );
-  return { test: 'test' };
+const deleteAllocatedWorkerSchema = yup.object().shape({
+  id: yup.number().required().integer(),
+  deallocationReason: yup.string().required(),
+});
+
+export const deleteAllocatedWorker = async (params) => {
+  const body = await deleteAllocatedWorkerSchema.validate({ ...params });
```
sirLisko, line 29:
```suggestion
  const body = await deleteAllocatedWorkerSchema.validate(params);
```

Spreading a recreating an object is an overhead here.

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
```diff
@@ -5,29 +5,33 @@ import PropTypes from 'prop-types';
...
 import Spinner from 'components/Spinner/Spinner';
-import { deleteAllocatedWorkers } from 'utils/api/allocatedWorkers';
+import { deleteAllocatedWorker } from 'utils/api/allocatedWorkers';
 
 const DeallocatedWorkersDetails = ({
   personId,
+  id,
   allocatedWorker,
   firstName,
   lastName,
```
sirLisko, line 15: `firstName` and `lastName` are not longer returned by the `API`, we can pass it down from the parent component. Or, even rid of the person bit 😅

sirLisko: A good idea is to fix the component `propTypes` that at the moment are incomplete, and set `isRequired` to the meaningful ones.

sirLisko (commented):
Looks good, just a couple of small notes

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
Miles-Alford: :thumbsup: decided on removing the person section since it doesn't really bring any value as the user would know what case their deallocating. made all the `propTypes` required beside `workerType` as I don't think that is very meaningful compared to knowing their name or what team their in

utils/server/allocatedWorkers.js
Miles-Alford: 👍

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
```diff
@@ -40,15 +43,15 @@ const DeallocatedWorkersDetails = ({
     <form onSubmit={handleSubmit(onSubmit)}>
       <h2 className="govuk-heading-l">Reason for worker deallocation</h2>
```
sirLisko, line 44: This one could be a `h1` and we can remove the `className`

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
```diff
@@ -21,14 +21,17 @@ const DeallocatedWorkersDetails = ({
   const [complete, setComplete] = useState(false);
   const [error, setError] = useState(false);
   const [deallocationReason, setDeallocationReason] = useState('');
+  const [worker] = useState(allocatedWorker);
+  const [type] = useState(workerType);
+  const [workerTeam] = useState(allocatedWorkerTeam);
```
sirLisko, line 26: we don't need to store these anymore as we are closing the modal.

sirLisko (changes requested):
Last small thing 😅

components/AllocatedWorkers/DeallocateWorkerDetails/DeallocatedWorkerDetails.jsx
Miles-Alford: Should we not keep them for the future when we will use the confirmation screen?

sirLisko (approved):
Code looks good 👍 
Amazing work 👏

PR #140, opened 2021-01-20
link colour update

naomigreen:
**What**  
Requested style change to links.


stylesheets/all.scss
```diff
@@ -81,3 +81,12 @@ $screen-sm-min: 576px;
 .ReactModal__Body--open {
   overflow: hidden;
 }
+
+.link-override {
```
sirLisko, line 85: We should align all the links to use the same colour, so, in this case, we should override the `govuk-link`. Otherwise, this one would be the only different one.

sirLisko (approved):

PR #143, opened 2021-01-22
add clear search to Search residents and search cases

LBHABEBIC:
Added clear search to search forms. Now working on date inputs as added onClick and used reset from useForm.


components/Search/forms/SearchCasesForm.jsx
```diff
@@ -120,6 +122,26 @@ const SearchCasesForm = ({ onFormSubmit, defaultValues }) => {
       </div>
 
       <Button label="Search" type="submit" disabled={!isDirty} />
+      <input
+        className="govuk-body"
+        type="reset"
```
sirLisko, line 127: Let's make it a `span` (with `role="button"`) so we are sure it's not conflicting with the `form` itself.
Also, it will prevent the necessity of override a bunch of styles 👎 )

components/Search/forms/SearchCasesForm.jsx
```diff
@@ -120,6 +122,26 @@ const SearchCasesForm = ({ onFormSubmit, defaultValues }) => {
...
+          });
+        }}
+        style={{
+          background: 'transparent',
+          color: '#00664f ',
+          textDecoration: 'underline',
+          border: 'none',
+          cursor: 'pointer',
+          marginLeft: '20px',
+          marginTop: '10px',
```
sirLisko, line 140: We don't need most of these rules, we can simplify it as:
```suggestion
              margin-left: 1rem;
              line-height: 2.5rem;
```

components/Search/forms/SearchCasesForm.jsx
```diff
@@ -120,6 +122,26 @@ const SearchCasesForm = ({ onFormSubmit, defaultValues }) => {
       </div>
 
       <Button label="Search" type="submit" disabled={!isDirty} />
+      <span
+        className="govuk-body"
```
sirLisko, line 126: Let's make it `govuk-link` so it more consistent with all the other links.
```suggestion
        className="govuk-link"
```

components/Search/forms/SearchCasesForm.jsx
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
sirLisko, line 1: Can we fix the need to add these at the top of each file? 😅

components/Search/forms/SearchResidentsForm.jsx
```diff
@@ -2,7 +2,6 @@ import PropTypes from 'prop-types';
 import { useForm } from 'react-hook-form';
 import isPast from 'date-fns/isPast';
 import isPostcodeValid from 'uk-postcode-validator';
-
```
sirLisko, line 5: Can we keep an empty line between node modules and local imports? 🙏

components/Search/forms/SearchCasesForm.jsx
```diff
@@ -120,6 +122,26 @@ const SearchCasesForm = ({ onFormSubmit, defaultValues }) => {
       </div>
 
       <Button label="Search" type="submit" disabled={!isDirty} />
+      <span
+        className="govuk-body"
+        onClick={() => {
+          reset({
+            start_date: '',
```
sirLisko, line 129: you need to set it to `null` otherwise in the next search it will complain that is required

sirLisko (approved):
Code looks good 👍 
Well done 👏

PR #144, opened 2021-01-22
made the DateInput component uncrontrolled

sirLisko:
**What**  
- made the `DateInput` component (used by the `ControlledDateInput`) uncontrolled, this is making the component resettable and it's avoiding a bunch of conflicts related to the control of the inputs
- moved out `stringDateToObject` and `objectDateToString` and added to the `utils/date`
- improved `parseDate` util
- check that the _day_ and _month_ are two digits, as 1 digit is not returning any result

**How to test it**
Please double-check that the component is working as expected


naomigreen (approved):
Now able to clear date input with form reset

PR #146, opened 2021-01-23
Added ObjectInput

sirLisko:
**What**  
- Added ObjectInput form component, it's also supporting inline formatting (with `isInline` prop)
- Extended Summary to show object components inline, using `summaryInline`
- Added _object step_ in the test form, for demo purpose of the feature

**Example:**
The component is going to be declared as follow:

```
{
  component: 'Object',
  name: 'phone_number',
  label: 'Phone Number',
  components: [
    {
      component: 'PhoneInput',
      name: 'phoneNumber',
      label: 'Phone number',
      rules: { required: true },
    },
    {
      component: 'TextInput',
      name: 'phoneType',
      label: 'Phone type',
    },
  ],
},
```
and it's returning:
```
{
  phoneType: "main contact",
  phoneNumber: "077..."
}
```

Screenshot 2021-01-22 at 18 09 32
Screenshot 2021-01-22 at 18 19 11
Screenshot 2021-01-22 at 18 21 10

**Extras**
- Moved `Button` component outside the `Form` as it's used in several different places, unrelated to the Forms
- Moved `DynamicStep` and `DynamicInput` inside the `FormWizard` folder

**How to test it**
- go to http://dev.hackney.gov.uk:3000/form/test/first-step
- select **show object step**
- check that is working properly


components/Form/ObjectInput/ObjectInput.jsx
```diff
@@ -0,0 +1,51 @@
...
+import DynamicInput from 'components/FormWizard/DynamicInput';
+
+import styles from './ObjectInput.module.scss';
+
+const ObjectInput = ({
+  label,
+  hint,
+  name,
+  labelSize = 'm',
+  isInline,
```
naomigreen, line 13: Could the default value be set to false?

naomigreen (changes requested):
Tested and working.

components/Form/ObjectInput/ObjectInput.jsx
sirLisko: `undefined` is falsy, do we really need to specify a `false`? We are not doing it in any other place in the codebase

naomigreen: It would be nice to not have to disable linting for this.
https://github.com/LBHackney-IT/lbh-social-care/blob/5ac4ddd3bbadac1f122d555f3fd49b3ba94d6f63/components/Form/Radios/Radios.jsx#L18

sirLisko: Even if you have it the `eslint-disable-next-line no-unused-vars` is still needed because `  summaryInline` is used in another component and not in this one.

sirLisko: I double-checked and the `isRadiosInLine` is the only `bool` prop with default value, I am still thinking that is going to be superfluous specify it to `false`.

naomigreen (approved):

PR #150, opened 2021-01-26
use nationality.js in Select component

LBHABEBIC:
removed all mentions of the Nationalities.jsx component. British appears in the nationalities.js list already.


data/forms/create-new-person.jsx
```diff
@@ -1,3 +1,6 @@
+// @ts-nocheck
```
sirLisko, line 1: I think you have some weird settings in your editor 🤔

sirLisko (commented):
We can get rid of `Nationality` component (check all the occurrences).

Also, `british` disappeared, as it was hardcoded in the component, we need to add it back.

sirLisko (approved):
Code looks good 👍 

PR #151, opened 2021-01-26
add ethnicities and sub-ethnicities to add new person

LBHABEBIC:
image

### DIFF
--- a/data/forms/create-new-person.jsx
+++ b/data/forms/create-new-person.jsx
@@ -71,16 +71,16 @@ export default {
         },
         {
           component: 'Select',
-          name: 'ethnicities',
+          name: 'ethnicity',
           label: 'Ethnicity',
           options: Object.keys(ETHNICITIES),
         },
         {
           component: 'Select',
-          name: 'Sub-ethnicity',
+          name: 'macro_ethnicity',
           label: 'Sub-ethnicity',
-          options: ({ ethnicities }) => ETHNICITIES[ethnicities],
-          conditionalRender: ({ ethnicities }) => ethnicities,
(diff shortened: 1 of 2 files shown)


data/forms/create-new-person.jsx
```diff
@@ -66,6 +69,20 @@ export default {
...
         },
+        {
+          component: 'Select',
+          name: 'ethnicities',
+          label: 'Ethnicity',
+          options: Object.keys(ETHNICITIES),
+        },
+        {
+          component: 'Select',
+          name: 'Sub-ethnicity',
```
sirLisko, line 80: This should be `ethnicity` because we are only interested in the `id` returned by the the sub-ethnicity select.
We can call the other one `macro_ethnicity`, or something like that.

data/forms/create-new-person.jsx
```diff
@@ -66,6 +69,20 @@ export default {
...
         },
+        {
+          component: 'Select',
+          name: 'ethnicity',
+          label: 'Ethnicity',
+          options: Object.keys(ETHNICITIES),
+        },
+        {
+          component: 'Select',
+          name: 'macro_ethnicity',
```
sirLisko, line 80: They should be the opposite, this one is going to indicate the final ethnicity id.

Miles-Alford: 🤦  understood it after I pushed up the changes, committed new changes now

sirLisko (approved):
Code looks good 👍 

data/forms/create-new-person.jsx
sirLisko: It's not really intuitive, it's actually a trick in order to not to have to change the data once submitted :)

PR #157, opened 2021-01-27
Fix props for allocation table

Miles-Alford:
**What**  
Removes the unneeded prop of `personName` from the `AllocatedWorkersTable` component and adds the missing prop types of `id` and `personId` to the `records` attribute and the `updateWorkers` function prop to the list of required prop types
**Why**  
To better reflect what props are actually needed for the component
**Anything else?**
Forgive the confusing branch name, this PR is supposed to just rectify the missing props from the component


sirLisko (approved):
👍

PR #169, opened 2021-02-03
Refactored `AddForm` to use the getResident results from the parent

sirLisko:
**What**  

- removed `getResident` from `AddForm` and passed it down from `PersonView`
- `PersonView` supports children as `node` and `func` (with render prop)

**Details**

At the moment the `AddForm` component is waiting that the `getPerson` in `PersonView` is finished to rerun the same call. We can show the `AddForm` component as soon as the first call is completed.
Screenshot 2021-02-03 at 18 22 07

**How to test it?**

- check that `/people/:id/records` is working properly


naomigreen (approved):
Working

PR #170, opened 2021-02-04
fixed auth in api

sirLisko:
**What**  
At the moment is possible to access the APIs if only a valid hackneyToken is present

**How to test it?**
- Check the APIs as `unauthenticate` user
- Check the APIs as `authenticate` user but not part of the correct group


naomigreen (approved):
Tested, getting 500 for not a user and 403 for not authorised

PR #171, opened 2021-02-04
don't allow phone input to accept letters

LBHABEBIC:
Fixed validation on inputs by using register as a function with the rules. The definition of register with rules has been moved from the Dynamic Input to the single components.  

https://trello.com/c/cLxzEZlh/687-phone-input-bug

image


sirLisko (approved):
👍

PR #172, opened 2021-02-04
confirmation page for new person

naomigreen:
Confirmation page for create new person
Joy asked us to remove the confirmation message and Felix requested the ID be on the page
Screenshot 2021-02-09 at 19 54 47

### DIFF
--- a/components/FormWizard/FormWizard.jsx
+++ b/components/FormWizard/FormWizard.jsx
@@ -20,6 +20,8 @@ const FormWizard = ({
   title,
   personDetails,
   includesDetails,
+  customConfirmation,
+  customSummary,
 }) => {
   Router.events.on('routeChangeComplete', () => {
     window.scrollTo(0, 0);
@@ -34,7 +36,11 @@ const FormWizard = ({
     ...(continueForm ? getData(formPath)?.data : {}),
   });
   const [queryString] = useState(otherQS);
-  const steps = createSteps(formSteps);
+  const steps = createSteps(formSteps, {
+    confirmation: customConfirmation,
+    summary: customSummary,
+  });
+
   const stepPath = `${formPath}[step]`;
   const step = steps.find(
     ({ id }) => id === (Array.isArray(stepId) ? stepId[0] : stepId)
--- a/components/Steps/Summary.jsx
+++ b/components/Steps/Summary.jsx
@@ -21,7 +21,13 @@ const SummaryStep = ({ formData, formSteps, formPath, onFormSubmit }) => {
       );
       deleteData(formPath);
       Router.replace(
-        `${formPath}confirmation${data?.ref ? `?ref=${data.ref}` : ''}`
+        `${formPath}confirmation${
+          data?.ref
+            ? `?ref=${data.ref}`
+            : data?.personId
+            ? `?personId=${data.personId}`
+            : ''
+        }`
       );
     } catch {
       setHasError(true);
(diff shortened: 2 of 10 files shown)


data/forms/create-new-person.jsx
```diff
@@ -195,6 +195,7 @@ export default {
           name: 'emailAddress',
           width: '20',
           label: 'Email address',
+          rules: { required: true },
```
sirLisko, line 198: Checked this with Tuomo and email address is not required

components/Steps/Confirmation.jsx
```diff
@@ -10,8 +10,37 @@ const ConfirmationStep = ({
   successMessage,
 }) => {
   const router = useRouter();
-  const { ref } = router.query;
+  const { ref, personId } = router.query;
   if (!formSteps) return null;
+  if (personId)
+    return (
```
sirLisko, line 16: `Confirmation` component should stay generic, with this we are adding business logic with a component that should stay as generic as possible.

We can do it in two ways:
- pass a confirmation component to the `formWizard` 
- define a confirmation step (as component) in the `data/create-new-person`

Also using `personId` as a check to show this might have problems with the `case` forms, as we are using the `personId` in the URL to define to which person we want to add the case.

sirLisko (changes requested):

components/FormWizard/FormWizard.jsx
```diff
@@ -47,19 +49,22 @@ const FormWizard = ({
...
     <div className="govuk-width-container">
       <NextSeo title={`${step.title} - ${title}`} noindex={true} />
-      {currentStepIndex !== 0 && step.id !== 'confirmation' && (
-        <a className="govuk-back-link" href="#" onClick={() => Router.back()}>
-          Back
-        </a>
-      )}
+      {currentStepIndex !== 0 &&
+        step.id !== 'confirmation' &&
+        step.id !== 'person-confirmation' && (
```
sirLisko, line 54: We don't need this anymore

components/Steps/Summary.jsx
```diff
@@ -20,9 +20,17 @@ const SummaryStep = ({ formData, formSteps, formPath, onFormSubmit }) => {
         filterDataOnCondition(formSteps, sanitiseObject(formData))
       );
       deleteData(formPath);
-      Router.replace(
-        `${formPath}confirmation${data?.ref ? `?ref=${data.ref}` : ''}`
-      );
+      if (formPath.includes('create-new-person')) {
```
sirLisko, line 23: We don't need this

utils/steps.jsx
```diff
@@ -1,10 +1,13 @@
 import Summary from 'components/Steps/Summary';
 import Confirmation from 'components/Steps/Confirmation';
-
-export const createSteps = (formSteps) => [
+export const createSteps = (formSteps, customStep) => [
   ...formSteps,
   { id: 'summary', title: 'Summary', component: Summary },
```
sirLisko, line 5: can we have the same for Summary, so we are consistent?

utils/steps.spec.jsx
```diff
@@ -18,17 +19,33 @@ describe('FormWizard', () => {
     { id: 'second-step', title: 'Second Step', components: ['foo', 'bar'] },
     { id: 'third-step', title: 'Third Step', components: ['foobar'] },
   ];
+  const customStep = { confirmation: PersonConfirmation };
+  const emptyStep = { confirmation: null };
```
sirLisko, line 23: If we specify `customStep?.confirmation` in the `createSteps` we don't need to pass the `emptyStep`.

sirLisko (changes requested):

sirLisko (approved):

PR #174, opened 2021-02-05
add provisioned concurrency for lambda latency issue

LBHABEBIC:
add provisioned concurrency for lambda latency issue


LBHMKeyworth (approved):
Looks good

sirLisko (approved):
Looks good, even if I think that the problem is the BE APIs 🤔

PR #176, opened 2021-02-08
unified buttons

sirLisko:
**What**

- merged LinkButton in Button
- removed unused page hub

Initially, the `LinkButton` was created for https://design-system.service.gov.uk/components/button/start/index.html, now that they look the same it doesn't make sense keep two different components.

**How to test it**


LBHABEBIC (approved):
makes sense

PR #178, opened 2021-02-08
refactored api gets using swr

sirLisko:
**What**  
- Improved API caching across the app, using [swr](https://swr.vercel.app/)
- Simplified get API, removed `try/catch` and `async`
- Fixed navigation in add allocation / removed allocation

**Todo**
Move postcode API, this PR is already massive, I didn't want to touch the `AddressLookup` component

**Note**
~~Bases on #177~~
E2E are fixed by #179

**How to test it**
Everything should work as before, but much faster.
A smoke test to the whole application would be ideal.


components/AllocatedWorkers/AddAllocatedWorker/AddAllocatedWorker.jsx
```diff
@@ -15,80 +15,62 @@ import {
...
-    setWorkersLoading();
-  });
-  useEffect(() => {
-    formValues.team && getAllocatedWorkers(formValues.team);
-  }, [formValues.team]);
-  useEffect(() => {
-    setPostError();
-  }, [formValues.team, formValues.worker]);
-  const addWorker = useCallback(async ({ worker }) => {
+  const { data: { teams } = {}, error: errorTeams } = getTeams();
```
Miles-Alford, line 25: Never seen this syntax before. From the look of it, its getting all the teams, is ``data`` an object of teams?

sirLisko: It means define `const teams` as `data.teams`.

`= {}` means that if data is `undefined` or `null` is not breaking, as it's falling back to `{}`. Otherwise, you will get `impossible to get teams of undefined`.

components/Cases/Cases.jsx
```diff
@@ -8,34 +7,36 @@ import Spinner from 'components/Spinner/Spinner';
...
+      </div>
+    );
+  }
   return (
     <>
-      {loading ? (
+      <div className="lbh-table-header">
         <div>
-          <Spinner />
+          <h3 className="govuk-fieldset__legend--m govuk-custom-text-color govuk-!-margin-top-0">
```
naomigreen, line 29: You're duplicating code here.

naomigreen (changes requested):

components/Cases/Cases.jsx
sirLisko: fixed 👍

naomigreen (approved):
Working as expected

PR #180, opened 2021-02-09
Add error message to restricted person cases

Miles-Alford:
**What**  
The error summary component has been added to the cases component to display a forbidden message instead of their cases if the person is restricted.

**Why**  
So that workers know why they are unable to see cases for certain persons

**Anything else?**
Unsure if this is the best solution for this, not a fan of calling the ``getResident`` endpoint in the cases component, also could probably make the CSS better
image


components/Cases/__snapshots__/Cases.spec.jsx.snap
```diff
@@ -1,5 +1,50 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Cases component should render a error message when a person is restricted 1`] = `
```
sirLisko, line 3: Always keep an eye on what the snapshots rendered, in this case it's a spinner 😅

components/Cases/Cases.jsx
```diff
@@ -3,17 +3,29 @@ import PropTypes from 'prop-types';
...
 import Spinner from 'components/Spinner/Spinner';
 import { getCasesByResident } from 'utils/api/cases';
+import { getResident } from 'utils/api/residents';
 
 const Cases = ({ id }) => {
   const [error, setError] = useState();
   const [loading, setLoading] = useState(false);
+  const [personRestricted, setPersonRestricted] = useState();
   const [results, setResults] = useState();
+  const getPerson = async () => {
```
sirLisko, line 17: We can pass down `person` from `PersonView` as a render prop, as we did in the `record` page.

sirLisko (commented):
Looks good, added few notes so we don't forget about what we just discussed :)

components/Cases/Cases.jsx
```diff
@@ -2,11 +2,12 @@ import PropTypes from 'prop-types';
 
 import CasesTable from 'components/Cases/CasesTable';
 import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
+import ErrorSummary from 'components/ErrorSummary/ErrorSummary';
 import Button from 'components/Button/Button';
 import Spinner from 'components/Spinner/Spinner';
 import { getCasesByResident } from 'utils/api/cases';
 
-const Cases = ({ id }) => {
+const Cases = ({ id, person }) => {
```
sirLisko, line 10: We shouldn't forget to update the `PropTypes` :)

components/ErrorSummary/ErrorSummary.module.css
```diff
@@ -0,0 +1,6 @@
+.restricted_summary_box {
```
sirLisko, line 1: I am being picky here, but we should use css modules or override the style in all, not both 😅

sirLisko (approved):
Code looks good 👍 

PR #181, opened 2021-02-09
aligned allocations api to BE

sirLisko:
- aligned allocations to the new API
- replace the delete allocation url (so the browser back button doesn't show an already deleted allocation)XX


naomigreen (approved):
Back button returns to person view as expected after allocation is deleted. Allocation APIs working

PR #189, opened 2021-02-15
don't call the APIs if restricted

sirLisko:
**What**  
If the person is restricted, along with showing the error, it shouldn't call the APIs.


LBHABEBIC (approved):
Tested and cas see that cases api is not called.

PR #190, opened 2021-02-15
fixed formCaseWrapper

sirLisko:
- cached personId and added missing component
- extended `FormWrapper` for hide the back button
- prevent error on console for unsupported prop


naomigreen (approved):
Great!!

PR #191, opened 2021-02-15
improved create new person

sirLisko:
- change path and don't need to specify step if only one
  Moved from `/form/create-new-person/client-details` to `/people/add`
- fixed protypes
- refactored _save form in progress_, centralised the localstorage in one key `social-care-forms`, as now the forms don't have necessarily `/form` prefix

this is preparatory for the people update, that ideally will be /people/:id/update 

**How to test it**
Everything should work as before but with the new url.


naomigreen (changes requested):
Unable to save create new person

naomigreen (approved):
Looks good

PR #194, opened 2021-02-16
fall back to timestamp if no date if event

sirLisko:
**What**
- improved sanitisation of the date
  supported syntax:
```
22/09/1941
22/01/2021 10:46:42
2000-12-12
2021-01-19T16:28:54.1295896Z
```
- fallback to `caseFormTimestamp` if `dateOfEvent` is not present

Screenshot 2021-02-16 at 12 29 19

It's not ideal using regex for extract dates but until the BE is not fixing this it's good enough.


naomigreen (approved):
Looks good

PR #195, opened 2021-02-16
properly show the allocation/deallocation page, based on the record

sirLisko:
**What**  
- added possibility to view the allocation records
- fixed the `getCase` if the allocation/deallocation is not in the first 20 records returned (this is going to be fixed when the BE will release a direct get)

Screenshot 2021-02-16 at 12 50 31

**How to test it**
Check http://dev.hackney.gov.uk:3000/people/33343831 as it has allocations and deallocations.


LBHABEBIC (approved):
Tested the branch and looks good.

PR #202, opened 2021-02-18
display one form if only one exists

LBHABEBIC:
Show no placeholder if there is only one form.


sirLisko (approved):
Looks good 👍 
The E2E should be fixed now 🤞

PR #203, opened 2021-02-18
filter form search by user

LBHABEBIC:
Created a list of combined forms.

Updated tests.of forms should be displayed tot he worker.

Created functionality to use permissionFlag to decide which list


components/Search/forms/SearchCasesForm.jsx
```diff
@@ -5,7 +5,11 @@ import isPast from 'date-fns/isPast';
 import { convertFormat } from 'utils/date';
 import { TextInput, Checkbox, DateInput, Select } from 'components/Form';
 import Button from 'components/Button/Button';
-import FORM_NAMES from 'data/formNames';
+import CHILD_FORMS from 'data/googleForms/childForms';
+import ADULT_FORMS from 'data/googleForms/adultForms';
```
sirLisko, line 9: the problem here is that `googleForms` will have the url as value for how the select is working it's going to be passed to the search instead of the name

LBHABEBIC: do we need to make seperate lists of form names that don't have the url as the value.

sirLisko: we can extract this logic in a utility function, something like `getFormNames` and then just pick what we need. We can also generate `ADULT_AND_CHILDREN` so we are sure that doesn't get out of date

sirLisko (approved):
We should check with the BE/prod how we want to align the form ids, but the functionality is there and looks good 👍

PR #208, opened 2021-02-22
converted server side code to typescript

sirLisko:
**What**
- typescript basic configuration
  `allowJs` is set to true, so we can continue using JS for the rest of the application, it would be nice migrate it over time
- converted all the code used server side `/pages/api`, `/utils/server` and all the utils used by those sections
- added basic generic `types`, this will be extended over time, ideally starting from a consistent swagger

**Why** 
As the application is growing it will make our life easier in the long term.
We will be more consistent with other internal applications. 
Currently, We don't have any test for the `/pages/api` this will add a extra layer of confidence.

**How to test it**


naomigreen (approved):
Tested

PR #211, opened 2021-02-23
autocomplete

naomigreen:
Autocomplete component 
Screenshot 2021-02-23 at 10 04 59

To test:
go to http://dev.hackney.gov.uk:3000/form/test/first-step
testing with           
options: [
            { text: 'foo', value: 22 },
            { text: 'bar', value: 41 },
          ],
submit the form and check data value is correct

### DIFF
--- a/components/Form/Autocomplete/Autocomplete.jsx
+++ b/components/Form/Autocomplete/Autocomplete.jsx
@@ -1,0 +1,199 @@
+import { useState, useCallback, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
+import DeleteIcon from 'components/Icons/TimesCircle';
+
+import { Controller } from 'react-hook-form';
+import style from './Autocomplete.module.scss';
+
+const AutocompleteComponents = ({
+  name,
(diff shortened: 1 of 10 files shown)


components/Form/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,199 @@
...
+          {error && <ErrorMessage label={error.message} />}
+          <div className={style.inputContainer}>
+            <input
+              className={style.input}
+              id={name}
+              placeholder={placeholder || 'Search or select a value'}
+              data-testid={name}
+              name={name}
+              value={displayView}
+              defaultValue={selected}
```
sirLisko, line 101: You shouldn't set `value` and `defaultValue` at the same time, react is not happy about it https://reactjs.org/docs/forms.html#controlled-components

sirLisko (changes requested):
I found a few issues:
- the component is not repopulating, check `/people/234/allocations/add?teamId=2` 
- it's not really accessible, there are a bunch of `aria` field missing and I can't navigate it with the keyboard (that I guess is something common in autocomplete)

My suggestion is to use a 3rd party component to make our life easier.

sirLisko:
Closing in favour of #215

PR #213, opened 2021-02-23
`addPerson` returns `ref` instead of `personId`

sirLisko:
**What**  
- Removed `personId` from `components/Steps/Summary`, used `ref` instead, so it's not bound to any page logic
- removed `ref` from `addCase` as doesn't make sense showing the id of the mongodb record to the user

**How to test it**
Create person should work as before
Create new case should work as before but the ref shouldn't be visible anymore

Screenshot 2021-02-23 at 17 27 06


utils/api/residents.js
```diff
@@ -12,5 +12,5 @@ export const useResident = (personId) => useSWR(`/api/residents/${personId}`);
 
 export const addResident = async (formData) => {
   const { data } = await axios.post(`/api/residents`, formData);
-  return data;
+  return { ref: data?.personId, data };
 };
```
naomigreen, line 16: We're not getting the personId back as a ref, resulting in undefined on the create person confirmation page

naomigreen (changes requested):

utils/api/residents.js
sirLisko: ehm... did you try it?

sirLisko: this is actually returning the `personId` as `ref` 👇 
Screenshot 2021-02-26 at 15 01 27

naomigreen (approved):
Working now

sirLisko:
It was working even before O_o

PR #215, opened 2021-02-25
autocomplete

naomigreen:
Found a solution to our object options.
You should see and filter the text but get the value on select
Screenshot 2021-02-26 at 17 46 25
Screenshot 2021-02-26 at 17 45 35
Screenshot 2021-02-25 at 02 13 21
Screenshot 2021-02-25 at 02 13 04

- Anything else in this PR that isn't covered by the what/why? Downshift, 

Test Search for records by person (filter by form type)
Test add record page


sirLisko:
There are the same problems as the other PR:
- if you go back to the form the value is not repopulated
- arrow navigation is not working
- I guess we want the search to be case sensitive

components/AddForm/AddForm.jsx
```diff
@@ -1,70 +1,48 @@
...
-          label="Choose a form category"
-          placeHolder="Choose one"
-          width={30}
-          onChange={(value) => setCategoryValue(value)}
+      <div className="govuk-form-group">
+        <Autocomplete
+          name="formList"
+          options={forms}
+          label="Choose a form"
+          control={control}
```
sirLisko, line 26: shouldn't we pass the control only if we are using `useForm`?

components/AddForm/__snapshots__/AddForm.spec.jsx.snap
```diff
@@ -3,41 +3,127 @@
...
+        <ul
+          aria-labelledby="downshift-1-label"
+          class="suggestionList"
+          id="downshift-1-menu"
+          role="listbox"
         >
-          FAST
-        </option>
-      </select>
+          <li
```
sirLisko, line 48: we should mock the `data/googleForm` import used by this snapshot, otherwise if we are going to change that list this test is going to break

components/Form/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+AutocompleteComponents.propTypes = {
+  name: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  labelSize: PropTypes.string,
+  hint: PropTypes.string,
+  onChange: PropTypes.func,
+  placeholder: PropTypes.string,
+  options: PropTypes.array,
+  register: PropTypes.func,
```
sirLisko, line 129: `register` is not used

sirLisko (commented):
Added few notes.

I noticed that the input is not adapting the width, now it's not a problem but when we will add it to a two column layout it would be.
Also, the line under the input is something required? It looks really weird

Screenshot 2021-03-01 at 13 16 57

sirLisko:
Also the placeholder text looks a bit different than the default one
Screenshot 2021-03-01 at 13 24 10

components/AddForm/AddForm.jsx
naomigreen: We get a runtime error "controller is missing control" without it

sirLisko: It's because you should wrap `AutocompleteComponents` with `Controller` only if it's meaningful. In this case we are not using `useForm` so it doesn't make sense to be wrapper in `Controller`.

sirLisko: see comment above

components/Form/Autocomplete/Autocomplete.jsx
```diff
@@ -152,26 +151,4 @@ AutocompleteComponents.propTypes = {
   }),
 };
 
-const Autocomplete = ({ name, control, ...otherProps }) => (
```
sirLisko: We still need the "controlled" version of the component otherwise it's not going to work in the forms.

sirLisko (approved):
Tested for allocations and it's working.

Let's merge this, as it's blocking the release of the feature, but there are 2 problems that need to be solved before we can use this component in the forms:
- repopulation
- show `text` instead of `value` in summary
Screenshot 2021-03-02 at 12 38 49

PR #223, opened 2021-03-01
only allow to access add and removed url if part of the correct user group

sirLisko:
**What**  
A user that is not part of the allocators group won't be able to access `/person/:id/allocations/add` and `/person/:id/allocations/:allocationId/remove`

**How to test it**
- try to navigate to the addresses listed above as user without `hasAllocationsPermissions` and see that you are going to be redirected to the person page
- try to do the same but with a user with `hasAllocationsPermissions` and it should work as before


naomigreen:
Looks great!

Since users often share links, could we have a popup informing the user they don't have permission to access that page?

sirLisko:
Potentially yes, but I don't think is a priority right now.

PR #226, opened 2021-03-02
removed next/seo, used next/head instead

sirLisko:
**What**  
- Removed third-party module `next/seo` in favour of the native `next/head` (https://nextjs.org/docs/api-reference/next/head)
- `next-seo` was roughtly 7kb gzip according to bundlephobia https://bundlephobia.com/result?p=next-seo@4.20.0
- `noidex` is already been covered by https://github.com/LBHackney-IT/lbh-social-care/blob/main/public/robots.txt

**How to test it**
Page titles should work as before

### DIFF
--- a/components/FormWizard/FormWizard.jsx
+++ b/components/FormWizard/FormWizard.jsx
@@ -1,9 +1,9 @@
 import { useState } from 'react';
 import PropTypes from 'prop-types';
 import Router, { useRouter } from 'next/router';
-import { NextSeo } from 'next-seo';
 import { useBeforeunload } from 'react-beforeunload';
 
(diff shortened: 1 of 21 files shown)


LBHABEBIC (approved):
Code looks fine but i dont really understand the purpose of this seo component.

sirLisko:
in order to set a page title you need to do:

```
  <Head>
    <title>I am the title</title>
  </Head>
```

`next/head` is supporting multiple fields, i.e. the meta tags, if we want to add them we just need to extend `Seo` component instead of changing all the occurrences of `next/head`.

I hope it clarifies a bit, otherwise we can char further :)

PR #228, opened 2021-03-02
design changes

naomigreen:
Gill and Joy's change requests 

Placeholder change and down arrow
Screenshot 2021-03-02 at 14 53 36


components/Icons/DownArrow.jsx
```diff
@@ -0,0 +1,48 @@
...
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
+    <g></g>
```
sirLisko, line 40: Do we need all these empty groups?

components/Form/Autocomplete/Autocomplete.jsx
```diff
@@ -84,6 +85,11 @@ export const Autocomplete = ({
                 error ? { borderColor: '#d4351c' } : { borderColor: '#000' }
               }
             />
+            {!isOpen && !inputValue && (
+              <span onClick={toggleMenu}>
+                <DownArrow onClick={toggleMenu} />
```
sirLisko, line 90: Can we have the onclick only in the span, and have it `role="button"`?

sirLisko (approved):
👍

PR #229, opened 2021-03-02
updated form view 

naomigreen:
Updated form view

Screenshot 2021-03-03 at 17 59 54

To test /44000000/records/6039265d832298cc73518c42
or create a case note and view it.
Please note old case notes are not visible as they don't have a form_name_overall which is require in caseLink to redirect you to the view form page

### DIFF
--- a/components/Cases/CaseLink.jsx
+++ b/components/Cases/CaseLink.jsx
@@ -1,11 +1,16 @@
 import PropTypes from 'prop-types';
 import Link from 'next/link';
+import * as form from 'data/forms';
 
 const getLink = (recordId, { form_name_overall, ...caseFormData }) => {
+  const fileName = form[form_name_overall] ? form_name_overall : null;
+
   switch (form_name_overall) {
     case 'API_Allocation':
     case 'API_Deallocation':
       return `/people/${caseFormData.mosaic_id}/allocations/${caseFormData.allocation_id}?recordId=${recordId}`;
+    case fileName:
+      return `/people/${caseFormData.mosaic_id}/records/${recordId}`;
     default:
       return null;
   }
--- a/components/Cases/CaseRecap/CaseRecap.jsx
+++ b/components/Cases/CaseRecap/CaseRecap.jsx
@@ -1,0 +1,41 @@
+import PropTypes from 'prop-types';
+
+import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
+import Spinner from 'components/Spinner/Spinner';
+import Summary from 'components/Summary/Summary';
+import { useCaseByResident } from 'utils/api/cases';
+import * as form from 'data/forms';
+
+const CaseRecap = ({ personId, recordId }) => {
+  const { data: record, error: recordError } = useCaseByResident(
+    personId,
+    recordId
+  );
+
+  const recordData =
+    record && record.caseFormData && record.caseFormData.form_name_overall;
+  const fileData = form[recordData];
+
+  if (recordError || (recordData && !fileData)) {
+    return <ErrorMessage />;
+  }
+  if (!record || !fileData) {
+    return <Spinner />;
+  }
+  return (
+    <>
+      <Summary
+        formData={record.caseFormData}
+        formSteps={fileData}
+        formPath={`/people/${personId}/records/${recordId}/`}
+      />
+    </>
+  );
+};
+
+CaseRecap.propTypes = {
+  personId: PropTypes.string.isRequired,
+  recordId: PropTypes.string.isRequired,
+};
+
+export default CaseRecap;
--- a/pages/people/[id]/records/[recordId]/index.tsx
+++ b/pages/people/[id]/records/[recordId]/index.tsx
@@ -1,0 +1,26 @@
+import { useRouter } from 'next/router';
+
+import Seo from 'components/Layout/Seo/Seo';
+import BackButton from 'components/Layout/BackButton/BackButton';
+import CaseRecap from 'components/Cases/CaseRecap/CaseRecap';
+import PersonView from 'components/PersonView/PersonView';
+
+const CaseView = (): React.ReactElement => {
+  const {
+    query: { id, recordId },
+  } = useRouter();
+  return (
+    <>
+      <Seo title={`#${recordId} record`} />
+      <BackButton />
+      <h1 className="govuk-fieldset__legend--l gov-weight-lighter">
+        Case note for
+      </h1>
+      <PersonView personId={id as string} expandView>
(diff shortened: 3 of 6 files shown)


data/dictionary.js
```diff
@@ -0,0 +1,8 @@
+export default {
+  form_name: 'Form',
+  case_note_title: 'Title',
+  date_of_event: 'Date of event',
+  timestamp: 'Date created',
+  worker_email: 'Created by',
+  case_note_description: 'Description',
```
sirLisko, line 7: We already have these defined in the `data/forms`, in addition to be a duplication I can see that would go easily out of sync.

components/Cases/CaseLink.jsx
```diff
@@ -6,6 +6,8 @@ const getLink = (recordId, { form_name_overall, ...caseFormData }) => {
     case 'API_Allocation':
     case 'API_Deallocation':
       return `/people/${caseFormData.mosaic_id}/allocations/${caseFormData.allocation_id}?recordId=${recordId}`;
+    case 'ASC_case_note':
+      return `/people/${caseFormData.mosaic_id}/case-notes-recording/${recordId}`;
```
sirLisko, line 10: I think this should be more generic, like `/people/${caseFormData.mosaic_id}/records/${recordId}`, we should have custom case links only for records not created using our forms

components/Cases/CaseRecap/CaseRecap.jsx
```diff
@@ -0,0 +1,76 @@
...
+    return <ErrorMessage />;
+  }
+  if (!record) {
+    return <Spinner />;
+  }
+
+  const sortedData = [];
+
+  const isDate = (value) => {
+    if (!isNaN(new Date(value))) {
```
sirLisko, line 24: This is only working for `date` what about all the other components? 
We should reuse the `Summary` component, otherwise we are duplicating the logic.

components/Cases/CaseRecap/CaseRecap.jsx
```diff
@@ -0,0 +1,46 @@
...
+  );
+
+  const recordData =
+    record && record.caseFormData && record.caseFormData.form_name_overall;
+  const fileData = form[recordData];
+
+  if (recordError) {
+    return <ErrorMessage />;
+  }
+  if (!record || !fileData) {
```
sirLisko, line 23: I think if `fileData` doesn't exist (but `recordData` does) we should show an error, otherwise the spinner will just hang

pages/people/[id]/records/[recordId]/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+    query: { id, recordId },
+  } = useRouter();
+  return (
+    <>
+      <Seo title={`#${id} Case note`} />
+      <BackButton />
+      <h1 className="govuk-fieldset__legend--l gov-weight-lighter">
+        Case note for
+      </h1>
+      <PersonView personId={id as string} expandView={true as boolean}>
```
Comment (sirLisko) [inline pages/people/[id]/records/[recordId]/index.tsx:19]: We don't need to cast true as boolean, we can just have `expandView`

pages/people/[id]/records/[recordId]/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import CaseRecap from 'components/Cases/CaseRecap/CaseRecap';
+import PersonView from 'components/PersonView/PersonView';
+
+const CaseView = (): React.ReactElement => {
+  const {
+    query: { id, recordId },
+  } = useRouter();
+  return (
+    <>
+      <Seo title={`#${id} Case note`} />
```
Comment (sirLisko) [inline pages/people/[id]/records/[recordId]/index.tsx:14]: Here should be `#recordId record`. I know we have some mess around the codebase about titles, I will create a ticket

components/Summary/SummaryList.jsx
```diff
@@ -1,7 +1,19 @@
 import PropTypes from 'prop-types';
 
-const SummaryList = ({ list }) => (
+const SummaryList = ({ list, creation }) => (
   <dl className="govuk-summary-list">
+    {creation && (
```
sirLisko, line 5: This is creating the "created by" in every single section.

Maybe we need input from design on how to solve this problem

components/Cases/CaseLink.jsx
```diff
@@ -6,6 +6,8 @@ const getLink = (recordId, { form_name_overall, ...caseFormData }) => {
     case 'API_Allocation':
     case 'API_Deallocation':
       return `/people/${caseFormData.mosaic_id}/allocations/${caseFormData.allocation_id}?recordId=${recordId}`;
+    case 'ASC_case_note':
+      return `/people/${caseFormData.mosaic_id}/records/${recordId}`;
```
sirLisko, line 10: I think we should refactor a bit this logic, we need to check that the formName is included in `data/form` as in `CaseRecap`.

sirLisko (commented):
Looks good, I added a few comments, let me know if you have any question or you want to discuss those points further :)

components/Cases/CaseRecap/CaseRecap.jsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <>
+      <h3 className="govuk-heading-m govuk-custom-text-color govuk-!-margin-bottom-0">
+        FORM CREATION
+      </h3>
+      <dl className="govuk-summary-list">
+        <div className="govuk-summary-list__row">
+          <dt className="govuk-summary-list__key">Date created</dt>
+          <dd className="govuk-summary-list__value">
+            {formatDate(record.caseFormTimestamp)}
```
sirLisko, line 35: This should be `dateOfEvent` and if it's not present `caseFormTimestamp`

components/Cases/CaseRecap/CaseRecap.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+  if (recordError || (recordData && !fileData)) {
+    return <ErrorMessage />;
+  }
+  if (!record || !fileData) {
+    return <Spinner />;
+  }
+  return (
+    <>
+      <h3 className="govuk-heading-m govuk-custom-text-color govuk-!-margin-bottom-0">
```
sirLisko, line 28: I guess we want to use the same styling used for the subheaders, like:

Screenshot 2021-03-08 at 10 58 20

components/Cases/CaseRecap/CaseRecap.jsx
sirLisko: and I think we want the 2 formats consistent Screenshot 2021-03-08 at 11 00 13

sirLisko (approved):
👍

PR #234, opened 2021-03-03
initialise phoneNumber type as main if not present

sirLisko:
**What**  
If the user inserts a number but not the type the form is valid but the BE breaks.

This is a temporary quick fix as production is broken.

**How to test it**
- create a new person and as a phone number specify the number but not the type.


Miles-Alford (approved):
👍

PR #235, opened 2021-03-03
Form conversation 3

Miles-Alford:
**What**  
Add conversation 3 form and its page to the application

**Why**  
First iteration of the conversation 3 form, this allows us to gather feedback on the form from users


sirLisko (approved):
Looks good, I changed a couple of small things:
- removed duplicated names (this will prevent inputs overriding each other)
- removed duplicated keys for react element
- removed options `yes / no`, by default if we don't specify the options it will fallback to those values (not a big deal 😅)

I also removed the required from the `TextArea`s in section `care-act-outcomes-and-eligibility-5`, as I think they shouldn't be required, feel free to rever my commit if you think they should :)

I check the form and looks already better than the original google forms :)
I didn't check specifically all the questions but I think it's good enough for the users to test it out.

Massive work, well done 👏

sirLisko:
Merging this, hope you don't mind 😅

PR #240, opened 2021-03-05
moved server side code into lib folder

sirLisko:
Moved all the code related to the BE APIs into `lib/` folder.

**Why**  
The `lib/` folder is more a standard way to isolate the server code in next.js.
It's also less confusing than having two similar folders in utils.


Miles-Alford (approved):
👍

PR #245, opened 2021-03-11
added warning note form

sirLisko:
**What**  
Added warning note form for Adult and Child

Ad the Child form is the same as the Adult but without _WARNING_DISCLOSURE_, I created a block per each section and then merged them in the Step definition, instead of duplicating the whole form definition.

**How to test it**  
Go to `/people/:peopleId/warning-notes/add`

**What is missing**
- Improving spacing in the page
- What to do about the line rendered on conditional question
- Check on dates (start date / end date should be validated together, this is going to be fun)
- Changes on copy on Summary / Confirmation, I would keep as generic as possible at the moment, we can see if we can push them back or find a generic labeling, but this PR focused only on the form itself

### DIFF
--- a/data/forms/warning-note.tsx
+++ b/data/forms/warning-note.tsx
@@ -1,0 +1,280 @@
+import { FormStep, FormComponentStep } from 'components/Form/types';
+import ExpandDetails from 'components/ExpandDetails/ExpandDetails';
+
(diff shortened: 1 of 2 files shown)


LBHABEBIC (approved):
tried and tested the form submission. looks really good!

PR #254, opened 2021-03-15
update confirmation

naomigreen:
**What**  
Joy's new design include buttons to the person view and search page. 
Will confirm with Joy if we can keep the person detail and not have name in the green section (would require additional refactoring)
Screenshot 2021-03-15 at 07 54 38

### DIFF
--- a/components/Steps/DetailConfirmation.jsx
+++ b/components/Steps/DetailConfirmation.jsx
@@ -1,0 +1,36 @@
+import PropTypes from 'prop-types';
+import { useRouter } from 'next/router';
+import Button from 'components/Button/Button';
+
+const DetailConfirmation = ({ formSteps, successMessage }) => {
+  const router = useRouter();
+  const { id } = router.query;
+  if (!formSteps) return null;
+  return (
+    <>
+      <h1 className="govuk-fieldset__legend--l gov-weight-lighter govuk-!-margin-bottom-7">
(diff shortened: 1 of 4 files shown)


components/Steps/Confirmation.jsx
```diff
@@ -1,45 +1,33 @@
...
-      <Summary
-        formData={formData}
-        formPath={formPath}
-        formSteps={formSteps}
-        isSummaryCollapsable={isSummaryCollapsable}
-      />
-    </div>
+      <div style={{ display: 'flex', justifyContent: 'space-between' }}>
+        <Button label="Return to search" isSecondary wideButton route={`/`} />
+        <Button label="View person" wideButton route={`/people/${id}`} />
```
sirLisko, line 21: This is wrong and we fixed it already on https://github.com/LBHackney-IT/lbh-social-care/pull/213. This is assuming that all the forms are related to residents.

sirLisko (changes requested):

components/Steps/Confirmation.jsx
naomigreen: Screenshot 2021-03-09 at 12 15 52
Joy's new confirmation designs. Should we create a new confirmation page?

sirLisko (approved):
Looks good 👍

PR #258, opened 2021-03-16
removed unused govuk boilerplate

sirLisko:
**What**  
- Removed govuk.js as it has not been used
- Removed unused fonts and images

**How to test it**
Everything should work as before


Miles-Alford (approved):
👍

PR #259, opened 2021-03-16
Warning Note Review Page: Add warning note recap

Miles-Alford:
**What**  
This is the first half of the warning note review page, it adds the WarningNoteRecap table and mocks the API calls to display data on the component, The page for the review has also been created and tests have been added to the component and mock API calls

**Why**  
Case workers should be able to renew or end a warning note depending on the circumstances allowing colleagues to know any changes in situation with persons the warning note is attached to

**Anything else?**
This is only one half of the work needed to be done on this page, what's next is adding the form to the page and a review page after the form has been completed
image


sirLisko (approved):
Looks good 👍

PR #263, opened 2021-03-17
Historic view

naomigreen:
Created a full page component for historic data
Screenshot 2021-03-17 at 19 56 44
 To test use branch historic-view-test, I've updated the data to not contain any personal information, you can replace this with the html files provided.


components/Icons/UpArrow.tsx
```diff
@@ -0,0 +1,36 @@
...
+  default: '#fff',
+  green: '#01674e',
+  black: '#000',
+};
+
+interface Props {
+  color?: keyof typeof colors;
+}
+
+const UpArrow = ({ color = 'default' }: Props): React.ReactElement => (
```
sirLisko, line 13: Can't we reuse the `DownArrow`?

components/PageView/PageView.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const scrollToTop = () => {
+    window.scrollTo({
+      top: 0,
+      behavior: 'smooth',
+    });
+  };
+
+  useEffect(() => {
+    window.addEventListener('scroll', toggleVisibility);
```
sirLisko, line 39: I would add a throttle here

sirLisko (commented):
Looks good, just a couple of small things

stylesheets/all.scss
```diff
@@ -89,10 +89,9 @@ $screen-sm-min: 576px;
 .down-arrow-icon {
   width: 15px;
   height: 15px;
-  top: 3px;
+  top: 7px;
   position: relative;
   cursor: pointer;
-  right: 25px;
 }
```
naomigreen, line 95: this is only being used in autocomplete which has been updated in my pr to this.

sirLisko: this is breaking the autocomplete ui, we should keep it in mind if we want to deploy this before the autocomplete

sirLisko (approved):
👍

PR #265, opened 2021-03-18
Review warning note form

Miles-Alford:
**What**  
Add the review form to the warning note review/end page, add custom confirmation component to display the correct messages once a warning note has been reviewed or ended

**Why**  
This form allows users to review a warning note and update its details or if need be end the warning note

**Anything else?**
The backend implementation is still not implemented as such this PR is only for getting the form ready once the backend  is ready.
image


sirLisko (approved):
Looks good 👍 

Well done 👏

PR #268, opened 2021-03-22
improved ui of the table

sirLisko:
**What**  
- improved UI of the table
- added title and case note prefix on the records

**before:**
Screenshot 2021-03-22 at 09 15 22

**after:**
Screenshot 2021-03-22 at 09 15 02
Screenshot 2021-03-22 at 09 31 00


naomigreen (approved):
Looks good!

PR #269, opened 2021-03-22
add view case link to record search

anabebs:
View link is now displaying on search by records. 

http://dev.hackney.gov.uk:3000/cases?first_name=john


sirLisko (approved):
Code looks good 👍 

PR #272, opened 2021-03-22
add case link conditions and new endpoint

naomigreen:
Added conditions and end point for historic data

Please note some case notes are blank but still have a title
Test with 
http://dev.hackney.gov.uk:3000/people/33159329


components/PageView/PageView.tsx
```diff
@@ -4,21 +4,16 @@ import Arrow from '../Icons/DownArrow';
...
 }
-const PageView = ({
-  created,
-  title,
-  type,
-  date,
-  note,
-  team,
-}: Props): React.ReactElement => {
+const PageView = ({ formData }: Props): React.ReactElement => {
```
sirLisko, line 16: we should create a type `HistoryCaseFromData` as `https://github.com/LBHackney-IT/lbh-social-care/blob/main/types.ts#L44` and used it here.

components/PageView/PageView.spec.tsx
```diff
@@ -4,12 +4,14 @@ import PageView from './PageView';
...
 describe(`PageView`, () => {
   const props = {
-    created: 'test@test.com',
-    title: 'Foo',
-    type: 'Bar',
-    date: '05/06/2020',
-    team: 'Foo 1223',
-    note: 'Foo Bar',
+    formData: {
+      created: 'test@test.com',
```
sirLisko, line 8: Did you agree with the BEers to call it `created`, because for the other cases we have `created_by`. We should stick with the same structure.

Other possible alignments could be:

`data` => `timestamp` or `date_of_event`

`title` => `case_note_title`

`type` => `form_name`

naomigreen: These were only created for the demo as I didn't have access to the rest of the data and assumed it would be different from our current case note data

components/Cases/CaseRecap/CaseNote.tsx
```diff
@@ -0,0 +1,44 @@
+import PropTypes from 'prop-types';
```
sirLisko, line 1: we don't need proptypes in ts

components/Cases/CaseRecap/CaseRecap.tsx
```diff
@@ -1,35 +1,21 @@
...
-  const fileData =
-    recordData && (form as Record<string, FormStep[]>)[recordData];
-  if (recordError || (recordData && !fileData)) {
-    return <ErrorMessage />;
+const CaseRecap = ({
+  personId,
+  recordId,
+  is_historical,
+}: Props): React.ReactElement => {
+  if (is_historical) {
```
sirLisko, line 15: we can do this in the `page` component, so we don't have a component just for an `if`

components/Cases/CaseRecap/HistoricNote.tsx
```diff
@@ -0,0 +1,36 @@
+import PropTypes from 'prop-types';
```
sirLisko, line 1: Proptypes are superfluous here as well

components/PageView/PageView.tsx
```diff
@@ -2,23 +2,15 @@ import { useEffect, useState } from 'react';
...
-  type,
-  date,
-  note,
-  team,
-}: Props): React.ReactElement => {
+  officerEmail,
+  formName,
+  dateOfEvent,
+  content,
+}: HistoricCaseData): React.ReactElement => {
```
sirLisko, line 13: 👍

sirLisko (commented):
Looks good, just a couple of small notes

sirLisko (approved):
Code looks good 👍 

PR #302, opened 2021-04-16
Manage worker

anabebs:
Added functionality for searching workers in the system using their email. This page should display the workers details through the WorkerRecap Component and reuse the allocations functionality to display their allocations.
Screen Shot 2021-04-16 at 15 37 27

### DIFF
--- a/components/AllocatedCases/AllocatedCases.tsx
+++ b/components/AllocatedCases/AllocatedCases.tsx
@@ -17,7 +17,7 @@ const AllocatedCases = ({ id }: Props): React.ReactElement => {
   }
   return (
     <>
-      {allocations.workers.length === 0 ? (
+      {allocations.workers?.length === 0 ? (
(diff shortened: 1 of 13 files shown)


components/AllocatedCases/AllocatedCases.tsx
```diff
@@ -17,7 +17,7 @@ const AllocatedCases = ({ id }: Props): React.ReactElement => {
   }
   return (
     <>
-      {allocations.workers.length === 0 ? (
+      {allocations.workers?.length === 0 ? (
```
sirLisko, line 20: I think we can simplify this conditional render (even if the problem is not related to this PR)

sirLisko (approved):
Looks good, we need to align a bit the UI but functionality wise works great.
We can improve it once we have all the different steps in place.
Also because it's hidden to the users, so good to go :)

PR #304, opened 2021-04-20
Fix filtering for CFS forms

Miles-Alford:
**What**  

This PR adds two new data lists, one for all the names of the adult forms and another for child forms. It also changes the form filter to use these new lists and removes unneeded code within the `getFormsByUserPermission` function.

**Why**  

This fixes an issue where child forms would not appear if filtering under the search by record filter.


naomigreen (approved):
Looks great!

utils/forms.ts
```diff
@@ -1,34 +1,23 @@
...
   return permission === 'C'
-    ? normaliseForms(CHILD_FORMS).sort(sortForms)
+    ? CHILD_FORM_NAMES.sort(sortForms)
     : permission === 'A'
-    ? normaliseForms(ADULT_FORMS).sort(sortForms)
-    : [...normaliseForms(ADULT_FORMS), ...normaliseForms(CHILD_FORMS)].sort(
-        sortForms
-      );
+    ? ADULT_FORM_NAMES.sort(sortForms)
+    : adminArray.filter((u, i) => i === adminArray.indexOf(u));
```
sirLisko, line 22: What is doing this line? I am a bit confused 🤔 
`adminArray.filter((u, i) => i === adminArray.indexOf(u))`

utils/forms.ts
```diff
@@ -1,34 +1,23 @@
-import CHILD_FORMS from 'data/googleForms/childForms';
-import ADULT_FORMS from 'data/googleForms/adultForms';
+import ADULT_FORM_NAMES from 'data/formFilterNames/adultFormNames';
+import CHILD_FORM_NAMES from 'data/formFilterNames/childFormNames';
 
 import { getPermissionFlag, UserPermissions } from 'utils/user';
 
-type Form = typeof ADULT_FORMS[number];
+const sortForms = (formA: string, formB: string) =>
```
sirLisko, line 6: As we are dealing with an array of string this sort is superfluous, we can just use `.sort()`

utils/forms.ts
Miles-Alford: Both lists have names which are the same such as case audit, if your an admin, you get duplicates within the filter list, this line just removes the duplication if you are an admin

PR #305, opened 2021-04-21
add case note recording

sirLisko:
**What**
- moved case-notes-recording
- used internal case notes recording

**How to test it**
- go to `http://dev.hackney.gov.uk:3000/people/7/records`
- select **case notes recording**
- you should see the internal form
- fill the form
- it should work properly


anabebs (approved):
tested and works as expected!

PR #307, opened 2021-04-26
e2e test update

naomigreen:
**What**  
End to end test update


cypress/integration/add_record_adult_admin.ts
```diff
@@ -0,0 +1,37 @@
...
+    cy.setCookie('hackneyToken', Cypress.env('TEST_KEY_ADULT_GROUP'));
+    cy.getCookie('hackneyToken').should(
+      'have.property',
+      'value',
+      Cypress.env('TEST_KEY_ADULT_GROUP')
+    );
+    cy.visit(Cypress.env('HOST') + '/search');
+  });
+
+  it('Confirm person type is adult', () => {
```
sirLisko, line 13: This is already covered by https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/main/cypress/integration/search_by_name_adult_admin.ts, not sure of the value it will bring 🤔

cypress/integration/add_record_childred_admin.ts
```diff
@@ -0,0 +1,41 @@
...
+    cy.setCookie('hackneyToken', Cypress.env('TEST_KEY_CHILDREN_GROUP'));
+    cy.getCookie('hackneyToken').should(
+      'have.property',
+      'value',
+      Cypress.env('TEST_KEY_CHILDREN_GROUP')
+    );
+    cy.visit(Cypress.env('HOST') + '/search');
+  });
+
+  it('Confirm person type is adult', () => {
```
sirLisko, line 13: Same as above, probably it would be more useful accessing directly the person page and check that the `case-notes-recording` is working properly.

sirLisko (approved):

PR #314, opened 2021-04-28
Extend phone numbers to support text

sirLisko:
**What**  
This is needed to support "re-population" of the input on _edit person_.

At the moment the data is not in a good shape and often there is text inside the number field.

Screenshot 2021-04-28 at 10 12 04


anabebs (approved):
works fine

PR #315, opened 2021-04-29
aligned warning notes BE names

sirLisko:
**What**  
- aligned warning notes `POST` object to BE
- fixed API `/residents/:id/warningnotes` (now the BE is returning `200` even if there aren't notes)
- added warningnote to the forms (only for admins)

**How to test it**
- go into a person page
- click add record
- search for **warning note**
- add a warning note
- check that it appears correctly in the person page


Miles-Alford (approved):
👍

PR #320, opened 2021-04-29
Wire up PATCH warning note endpoint to front end

Miles-Alford:
**What**  
This PR adds the PATCH endpoint call to the front end with all the required data.

**Why**  
To allow users to add a review to a warning note and either end it or renew it.

**Anything else?**
Would like some feedback on some of my decisions here specially on the tests for the API call and how I get the `endedDate` property.

The table with information won't render as of yet as the GET endpoint for an individual note doesn't seem to be ready, furthermore in a future PR the note must not be allowed to be re-reviewed if it is ended.

### DIFF
--- a/lib/warningNotes.spec.ts
+++ b/lib/warningNotes.spec.ts
@@ -79,7 +79,8 @@ describe('warningNotesAPI', () => {
       mockedAxios.patch.mockResolvedValue({ data: {} });
       await warningNotesAPI.updateWarningNote({
         status: 'Closed',
-        closedBy: 'foo@bar.com',
+        endedBy: 'foo@bar.com',
+        endedDate: '2021-04-29',
       });
       expect(mockedAxios.patch).toHaveBeenCalled();
       expect(mockedAxios.patch.mock.calls[0][0]).toEqual(
@@ -87,7 +88,8 @@ describe('warningNotesAPI', () => {
       );
       expect(mockedAxios.patch.mock.calls[0][1]).toEqual({
         status: 'Closed',
-        closedBy: 'foo@bar.com',
+        endedBy: 'foo@bar.com',
+        endedDate: '2021-04-29',
       });
       expect(mockedAxios.patch.mock.calls[0][2]?.headers).toEqual({
         'Content-Type': 'application/json',
--- a/lib/warningNotes.spec.ts
+++ b/lib/warningNotes.spec.ts
@@ -53,4 +53,55 @@ describe('warningNotesAPI', () => {
       expect(data).toEqual(['foo']);
     });
   });
+
+  describe('updateWarningNote', () => {
+    it('should work properly if the status is open', async () => {
+      mockedAxios.patch.mockResolvedValue({ data: {} });
+      await warningNotesAPI.updateWarningNote({
+        status: 'Open',
+        reviwedBy: 'foo@bar.com',
+      });
+      expect(mockedAxios.patch).toHaveBeenCalled();
+      expect(mockedAxios.patch.mock.calls[0][0]).toEqual(
+        `${ENDPOINT_API}/warningnotes`
+      );
+      expect(mockedAxios.patch.mock.calls[0][1]).toEqual({
+        status: 'Open',
+        reviwedBy: 'foo@bar.com',
+      });
+      expect(mockedAxios.patch.mock.calls[0][2]?.headers).toEqual({
+        'Content-Type': 'application/json',
+        'x-api-key': AWS_KEY,
+      });
+    });
+
+    it('should work properly if the status is closed', async () => {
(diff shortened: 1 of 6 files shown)


pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx
```diff
@@ -18,21 +19,17 @@ const ReviewWarningNote = (): React.ReactElement => {
...
-        firstName: person.firstName,
-        lastName: person.lastName,
-        contextFlag: person.contextFlag,
-        dateOfBirth: person.dateOfBirth,
-        workerEmail: user.email,
-        formNameOverall:
-          person.contextFlag === 'A' ? 'ASC_case_note' : 'CFS_case_note',
-        formName: form_name,
-        caseFormData: JSON.stringify(formData),
+    (person: Resident) => async ({ ...formData }: Record<string, unknown>) => {
```
Comment (chrishutchinson) [inline pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx:22]: I have a feeling the spread inside a destructure isn't necessary here now you've removed the `form_name` argument. I think the below is equivalent (worth testing though!):

```suggestion
    (person: Resident) => async (formData: Record<string, unknown>) => {
```

pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx
```diff
@@ -18,21 +19,17 @@ const ReviewWarningNote = (): React.ReactElement => {
...
+      updateWarningNote(warningNoteId, {
+        warningNoteId,
+        reviewedBy: user.email,
+        endedBy: formData.reviewDecision === 'No' ? user.email : undefined,
+        endedDate:
+          formData.reviewDecision === 'No'
+            ? new Date().toISOString().substr(0, 10)
```
Comment (chrishutchinson) [inline pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx:29]: Does it make more sense to pass a full ISO-8601 date here (without `.substr()`) or is this just the way the back end works?

pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx
```diff
@@ -18,21 +19,17 @@ const ReviewWarningNote = (): React.ReactElement => {
...
-        formNameOverall:
-          person.contextFlag === 'A' ? 'ASC_case_note' : 'CFS_case_note',
-        formName: form_name,
-        caseFormData: JSON.stringify(formData),
+    (person: Resident) => async ({ ...formData }: Record<string, unknown>) => {
+      updateWarningNote(warningNoteId, {
+        warningNoteId,
+        reviewedBy: user.email,
+        endedBy: formData.reviewDecision === 'No' ? user.email : undefined,
+        endedDate:
```
Comment (chrishutchinson) [inline pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx:27]: I think you asked for feedback on this in your PR notes. I might have said `endDate` is a little clearer, but I think it's a small thing. It probably makes most sense to remain consistent with naming we have elsewhere in the codebase and backend, or the terminology used by the people using the tool.

lib/warningNotes.spec.ts
```diff
@@ -53,4 +53,57 @@ describe('warningNotesAPI', () => {
       expect(data).toEqual(['foo']);
     });
   });
+
+  describe('updateWarningNote', () => {
+    it('should work properly if the status is open', async () => {
```
chrishutchinson, line 58: I think `should work properly...` is a little vague for a test title. What does 'work properly' mean? Instead I might expect something like: "it should respond to a PATCH request with a 200 OK and {some data} if the status is open".

Perhaps not one to resolve here and now, but something to think about. If you came back to this test in 6 months, would this name help you understand what it was testing?

chrishutchinson (commented):
Nice PR! I've left a handful of small comments for you

lib/warningNotes.spec.ts
Miles-Alford: Yeah I agree on this, many of our tests already follow this naming convention so I kept it the same for consistency but I agree they should be more descriptive

pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx
Comment (sirLisko) [reply pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx]: Yep, for the BE should be the same

pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx
Comment (sirLisko) [reply pages/people/[id]/warning-notes/[warningNoteId]/[[...stepId]].tsx:27]: I agree, but this feature is like 2 months that has been blocked by the BE and it's the 4th time that we agreed on naming and then they changed them. So we decided to align ourselves to their naming convention, I know it's crazy but this feature has been a real nightmare.

sirLisko (approved):

PR #322, opened 2021-04-30
fixed cases sanitisation regex

sirLisko:
**What**  
The current regex was breaking in case JSON was containing `(${text}), `.
Now the RegEx has been fixed.

Ideally the BE should filter out this property but it's a quick fix in the meantime.


Miles-Alford (approved):
👍

PR #324, opened 2021-05-05
fixed restricted on resident

sirLisko:
**What**  
- Fixed `restricted` on resident object, aligned it to the BE API
- Fixed re-population of the "restricted" input on resident edit

**How to test it**
- Check an edit resident page
- The "restricted" input is re-populated as expected

### DIFF
--- a/components/Cases/Cases.spec.tsx
+++ b/components/Cases/Cases.spec.tsx
@@ -5,14 +5,14 @@ import { UserContext } from 'components/UserContext/UserContext';
 import { userFactory } from 'factories/users';
 
 import { mockedNote, mockedAllocationNote } from 'factories/cases';
-import { mockedResident } from 'factories/residents';
+import { residentFactory } from 'factories/residents';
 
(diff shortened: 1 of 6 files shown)


anabebs (approved):
Looks good although i must admit it's a bit beyond me. Would John be a good person to look at this?

sirLisko:
@anabebs more details:
The BE was returning `restricted: 'Y' or 'N'`, the FE/BE layer was transforming this in a `boolean`, that we were using on the _record history_.
Because of this, the re-population of the _person update_ wasn't working, as it was expecting `Y` on `N` but it was getting a boolean.

In an ideal world, I would stick with booleans but apparently, Tuomo and Naomi agreed to have `Y` and `N` instead 🤷 


@JohnFarrellDev do you mind having a look at this :)

sirLisko:
This is a blocker for the UAT of the edit person, merging this, happy to revise it if needed :)

PR #341, opened 2021-05-10
add flexible answers component

jhackett1:
this adds a flexible answers component to enable the app to display records from the [care recording app](https://github.com/LBHackney-IT/social-care-recording)

here's a working demo of the component: https://codepen.io/jhackett1/pen/KKWwoJy?editors=0010

because this component, by its nature, decodes a block of json which could have wildly different nested structures, the typescript interfaces look a bit unusual - i've been as specific with them as is safe.

it adds the component to `components/Cases/CaseRecap/CaseNote`, where it is conditionally rendered if the record has the `form_name` of "Case recording".

this is far from ideal and should be replaced with a more robust means of telling what "kind" of recording it is.


components/Cases/CaseRecap/CaseNote.tsx
```diff
@@ -13,6 +14,11 @@ interface Props {
 const CaseNote = ({ personId, recordId }: Props): React.ReactElement => {
   const { data: record, error: recordError } = useCase(recordId);
 
+  if (record?.caseFormData?.case_note_description) {
```
JohnFarrellDev: Do we need to test this?

What is case_note_description is an empty string, malformed JSON, JSON in a format we expect.

components/FlexibleAnswers/FlexibleAnswers.tsx
```diff
@@ -0,0 +1,81 @@
...
+  </ul>
+);
+
+const RepeaterGroupAnswers = ({
+  data,
+}: {
+  data: (string | RepeaterGroupAnswer)[];
+}): React.ReactElement => (
+  <ul className="govuk-list lbh-list">
+    {data.map((item, i) => (
```
JohnFarrellDev: If this is a repeater group can new items be added, can items be removed?

If so we need to be careful with using the index as a key.

However do we have anything appropriate here that was could use as a unique identifier for each element.

tsconfig.json
```diff
@@ -17,6 +17,6 @@
     "jsx": "preserve",
     "baseUrl": "."
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "components/FlexibleAnswers/blah"],
```
JohnFarrellDev, line 20: What is this change for?

components/FlexibleAnswers/FlexibleAnswers.tsx
```diff
@@ -0,0 +1,81 @@
...
+      // questions and answers
+      [key: string]: string | (string | RepeaterGroupAnswer)[];
+    };
+  };
+}
+
+const FlexibleAnswers = ({ data }: Props): React.ReactElement => {
+  return (
+    <div>
+      {Object.keys(data).map((stepName) => (
```
JohnFarrellDev, line 54: Object.entries here could be nice as we are having to access the value below.

components/FlexibleAnswers/FlexibleAnswers.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <section key={stepName}>
+          <h3 className="lbh-heading-h3 govuk-!-margin-top-6 govuk-!-margin-bottom-4">
+            {stepName}
+          </h3>
+          <dl className="govuk-summary-list lbh-summary-list">
+            {Object.keys(data[stepName]).map((questionName) => (
```
JohnFarrellDev, line 60: I think the map inside of the map makes this confusing too read.

It also means the typeof check can be moved out of the JSX and then we could avoid the cast I believe.

JohnFarrellDev (commented):
Small changes, looks good

tsconfig.json
jhackett1: what the heck is that. will revert

components/Cases/CaseRecap/CaseNote.tsx
jhackett1: not happy with this whole bit of the code. we need a more foolproof way to know when to render flexible answers i think

components/FlexibleAnswers/FlexibleAnswers.tsx
```diff
@@ -0,0 +1,86 @@
...
+          : data[questionName]}
+      </li>
+    ))}
+  </ul>
+);
+
+const RepeaterGroupAnswers = ({
+  data,
+}: {
+  data: (string | RepeaterGroupAnswer)[];
```
JohnFarrellDev, line 25: We're using `data` a lot as a variable name, could we use something more descriptive?

jhackett1: answers?

components/Cases/CaseRecap/CaseNote.tsx
```diff
@@ -23,6 +25,19 @@ const CaseNote = ({ personId, recordId }: Props): React.ReactElement => {
   if (!record || !fileData) {
     return <Spinner />;
   }
+
+  // this needs to be fixed with a better way of distinguishing
+  if (
+    record?.caseFormData?.case_note_description &&
+    record.caseFormData.form_name === 'Case recording'
```
jhackett1, line 32: how might we better distinguish whether to render flexible answers?

anabebs (approved):
all good

PR #343, opened 2021-05-11
activate the first version of conversation 3

jhackett1:
this pr activates our first iteration of the conversation 3 form, making it available and ready for use by practitioners.

it:

- launches our conversation 3 form rather than the google form when adding a new record
- updates `/pages/person/[id]/records/conversation-3` form to submit answers as a new record
- modifies the `CaseNote` component to support displaying answers from the conversation 3 form
- makes some lightweight usability improvements to the conversation 3 form, like adding missing error messages

we know that there are issues with the ux of this form and the wording of questions. we're intending to iterate on this quickly, but it's important to launch this as-is to solve a pressing need for practitioners.


anabebs (approved):
Tested and works as expected.

PR #346, opened 2021-05-13
Add view link to a created warning note in the records history

Renny4Real:
**What**  
The creation of a warning note is visible in record history, but the ‘view’ link is missing.
This adds a condition to the case link component to render a `view` link to the warning note
within the records history.

It also adds a warning note id attribute to the case form data type so that it can be used to create the relevant path.

### Before:
Screenshot 2021-05-13 at 12 30 48 pm

### After:
Screenshot 2021-05-13 at 12 31 01 pm

### DIFF
--- a/components/Cases/CaseLink.tsx
+++ b/components/Cases/CaseLink.tsx
@@ -5,6 +5,7 @@ import {
   AllocationCaseFormData,
   CaseFormData,
   DeallocationCaseFormData,
+  WarningNoteCaseFormData,
 } from 'types';
 
 const getLink = (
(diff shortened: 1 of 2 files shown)


components/Cases/CaseLink.tsx
```diff
@@ -29,8 +29,8 @@ const getLink = (
       return `/people/${caseFormData.mosaic_id}/records/${recordId}?is_historical=${caseFormData.is_historical}`;
     case 'Historical_Visit':
       return `/people/${caseFormData.mosaic_id}/visits/${recordId}?is_historical=${caseFormData.is_historical}`;
-    case 'Warning_Note':
-      return `/people/${caseFormData.mosaic_id}/warning-notes/${recordId}`;
+    case 'API_WarningNote':
+      return `/people/${caseFormData.mosaic_id}/warning-notes/${caseFormData.warning_note_id}`;
```
chrishutchinson, line 33: If `warning_note_id` is typed as optional below (in `types.ts`) do we need to handle that case here, or is it more that our types don't accurately reflect the actual data we get (e.g. if `form_name_overall === "API_WarningNote"` then `warning_note_id` is always present)?

Renny4Real: I think it might be the latter, but my knowledge of typescript is still growing.
@anabebs @jhackett1

chrishutchinson: I've just had a quick skim of the `types` file and it looks like we already have a pattern for doing this in a more structured way, which is great news! @Renny4Real did you want to pair over that? I can go over the details a bit more on a call

chrishutchinson (approved):

PR #349, opened 2021-05-13
Fix/show more box

jhackett1:
this drastically simplifies the `<ShowMoreBox/>` component, retaining the basic functionality.

the old implementation relied on changing the height of the box by css, which isn't an accessible way of doing things, because screen reader-read content will be out of sync with what's visible on the screen.

the old implementation was also afflicted by [this bug](https://hackit-lbh.slack.com/archives/G01CYA7LM6V/p1620916703417300).

this:

- refactors the `<ShowMoreBox/>` component and its tests
- adds an arrow indicator to improve usability
- provides a generic `truncate()` method to shorten strings by word count, which will probably be useful in lots of places


chrishutchinson (approved):
Nice refactor 💯

chrishutchinson (changes requested):
Just un-approving while we add a test for new line characters (and review that there are no other case notes which might render HTML).

Details: https://hackit-lbh.slack.com/archives/G01CYA7LM6V/p1620988802440200?thread_ts=1620916703.417300&cid=G01CYA7LM6V

chrishutchinson (approved):

PR #350, opened 2021-05-14
feat(SCT-179): list all residents when searching, regardless of user type

chrishutchinson:
**What**  
* List all residents, adults and children, irrespective of the current user's group (ASC or CSF).
* Changes to our Cypress test structure, so they make a little more sense. Now each test file is for a single feature (e.g. `search_for_person` or `search_for_records`, and within each file we test the different user states). This gives us nicer test output, and a clearer way of testing different features for different user states. 
* As part of the above, introduces a new `cy.visitAs()` command which allows you to visit a page as a given user type, e.g. `cy.visitAs("/", AuthRoles.ChildrensGroup);` visits the home page as a user in the Children's group. This masks all the cookie-setting logic, so you can focus on the test at hand :D 

**Why**  
At the moment only residents of the type matching the user's context (A or C) are shown when searching. This is leading to duplicate records, and frustration amongst practitioners.

**Anything else?**


pages/api/residents/index.ts
```diff
@@ -19,10 +19,7 @@ const endpoint: NextApiHandler = async (
   switch (req.method) {
     case 'GET':
       try {
-        const data = await getResidents({
```
chrishutchinson, line 22: Unclear if we have tests for the `/api/...` endpoints, this is something I'll take a look at early next week.

cypress/integration/add_record.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('Adding records', () => {
+  describe('As a user in the Adults group', () => {
+    it('should go to the add case note form when "Case Note Recording" is selected', () => {
+      cy.visitAs(
+        `/people/${Cypress.env('ADULT_RECORD_PERSON_ID')}`,
+        AuthRoles.AdultsGroup
+      );
+
+      cy.contains('ASC');
+      cy.contains('Add a new record').click();
```
Renny4Real, line 12: Does this mean we can test for a component's presence (`contains`) and then interact with it (`click`)?
i.e. It's doing an `assertion` and then afterwards an `act` on the same line. 

Or is this a clever refactor and combining what could be multiple lines? 😁

cypress/integration/add_record.ts
```diff
@@ -0,0 +1,57 @@
...
+      cy.contains('Add a new record for').should('be.visible');
+      cy.contains('Expand view').click();
+      cy.contains('Collapse view').should('be.visible');
+      cy.get('[data-testid="formList"]').click();
+      cy.contains('CFS Case Note').should('be.visible');
+      cy.contains('CFS Visit').should('be.visible');
+      cy.get('[data-testid="formList"]').type('CFS Visit');
+      cy.contains('CFS Visit').should('be.visible');
+      cy.contains('CFS Case Note').should('not.exist');
```
Renny4Real, line 54: Where does `cy` get defined? Is it from a separate class?

cypress/integration/form.ts
```diff
@@ -1,70 +1,65 @@
...
 
-  it('Should work properly', () => {
-    cy.intercept(
-      {
-        method: 'POST',
-        url: '/test',
-      },
-      []
-    ).as('apiCheck');
+      cy.intercept(
```
Renny4Real, line 8: What does `intercept` do? Is it stubbing the POST request?

cypress/integration/form.ts
```diff
@@ -1,70 +1,65 @@
...
-        show_next_step: true,
-        show_multi_select_step: false,
-        show_object_step: false,
-        show_address_step: false,
-        show_autocomplete_step: false,
-        conditional_text: 'conditional name',
-        title_2: 'conditional step title',
-        'multi-step': [{ title_3: 'foo first' }, { title_3: 'foo second' }],
+      cy.contains('Submit').click();
+      cy.wait('@apiCheck').then((interception) => {
```
Renny4Real, line 47: It might be related to the previous question, but what does block do? 🙂

cypress/integration/search_for_records.ts
```diff
@@ -0,0 +1,25 @@
+import { AuthRoles } from '../support/commands';
+
+describe('Search for records by person', () => {
+  describe('As a user in the Admin Dev group', () => {
+    it('Search of Adult group record', () => {
+      cy.visitAs('/cases', AuthRoles.AdminDevGroup);
+
+      cy.contains('First name:').type(Cypress.env('ADULT_RECORD_FIRST_NAME'));
```
Renny4Real, line 8: How would Cypress know this would be an Adult first name? 
Is there some hidden test setup? 🤔 
I am assuming `contains` is an assertion?

cypress/support/commands.ts
```diff
@@ -0,0 +1,49 @@
...
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Cypress {
+    interface Chainable {
+      visitAs: typeof visitAs;
+    }
+  }
+}
+
+const visitAs = (
```
Renny4Real, line 32: Ah, so this a custom method we're adding to Cypress?

Renny4Real (commented):
Looks good. Got some questions 😀

.gitignore
```diff
@@ -34,8 +34,8 @@ assets/
 # cypress
 cypress/fixtures
 cypress/screenshots
-cypress/support
 cypress/videos
+cypress.env.json
```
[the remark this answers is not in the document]
chrishutchinson: This is being added – I've moved my env vars used for testing to a separate file (`cypress.env.json`) rather than part of `.env`. This is a convenience thing, just so I have more logical separation between env vars used for running the app, and env vars just used for running the tests

components/Cases/Cases.tsx
```diff
@@ -62,8 +62,35 @@ interface WrapperProps {
...
+  const isPersonRestricted = person.restricted === 'Y';
+
+  if (user.hasUnrestrictedPermissions) {
+    return true;
+  }
+
+  if (
+    user.hasAdultPermissions &&
+    person.contextFlag === 'A' &&
+    !isPersonRestricted
```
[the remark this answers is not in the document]
chrishutchinson: That's defined above, it just checks the `Y` or `N` value of `person.restricted`:

```ts
const isPersonRestricted = person.restricted === 'Y';
```

components/Search/results/ResidentsTable.tsx
```diff
@@ -28,7 +28,7 @@ const ResultTable = ({
 }: {
   records: LegacyResident[];
 }): React.ReactElement => (
-  <table className="govuk-table">
+  <table className="govuk-table" data-testid="residents-table">
```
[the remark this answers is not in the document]
chrishutchinson: Good question – I accidentally introduced a subtle bug in our Cypress test, because I wrote a test that searched for a user, then asserted their ID was visible. 

The test passed (because the ID was 7, and that was visible), but it was visible _elsewhere_ on the page, not inside the table.

So I introduced this `data-testid` attribute so my Cypress test could assert that the ID was found _inside the table_, not just anywhere on the page.

cypress/integration/add_record.ts
```diff
@@ -0,0 +1,57 @@
+import { AuthRoles } from '../support/commands';
+
+describe('Adding records', () => {
+  describe('As a user in the Adults group', () => {
+    it('should go to the add case note form when "Case Note Recording" is selected', () => {
+      cy.visitAs(
```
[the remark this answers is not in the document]
chrishutchinson: Yeah great question, I can add some docs to it.

In short it takes two or three arguments:

```ts
cy.visitAs(path: string, role: AuthRole, options?: Partial<Cypress.VisitOptions>);
```

Where:
* `path` – this is a string of the path (or full URL) you want to visit (e.g. `/cases`)
* `role` – this is a TypeScript enum, so you have to pass one of the explicit roles I have defined
* `options` (optional) – this is the second argument of the usual `cy.visit` function, so you can pass anything defined [here](https://docs.cypress.io/api/commands/visit#Arguments)

cypress/integration/add_record.ts
chrishutchinson: That's exactly right, Cypress methods are chainable, so it's getting the element that says "Add a new record" then clicking that element.

I can't take credit for this one though – I haven't actually changed these tests, I just moved the files around

cypress/integration/add_record.ts
chrishutchinson: It's a bit like the methods Jest provides (`it`, `describe`, `jest.fn`), it's magically defined because these files are run by the Cypress test runner

cypress/integration/form.ts
chrishutchinson: That's absolutely right, it is a bit like a mock at the HTTP level

cypress/integration/form.ts
chrishutchinson: So when we create the `intercept` we give it a name (`@apicheck`). Here we then use it, and assert that it had some specific properties on the request. 

Much like above, these tests were already in the codebase, I've just moved them around a bit 😄

cypress/integration/search_for_records.ts
chrishutchinson: So this is saying:
1. Get the element that contains `First name:` (this is likely a text input)
2. Type the value of the environment variable `CYPRESS_ ADULT_RECORD_FIRST_NAME` into that element

cypress/support/commands.ts
chrishutchinson: That's right, I've effectively extended the standard `cy.visit` with new functionality that makes more sense to us (and makes our tests a little cleaner).

.gitignore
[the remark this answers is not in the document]
Renny4Real: Sorry, I realised I highlighted the wrong diff 😅 (it looks like we're removing `cypress/support`)

cypress/integration/add_record.ts
Renny4Real: I see! 👍🏾

cypress/integration/search_for_records.ts
Renny4Real: Ah... I thought `.type` checked the `TYPE` of an object and not type a value into the element 😅

cypress/integration/add_record.ts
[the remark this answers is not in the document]
chrishutchinson: I've added a docblock comment so your editor will give you this detail when you try and use it – https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/350/commits/0af6f3eae95c96c61966f07b57ca192d57ad2843

Screenshot 2021-05-17 at 16 51 14

cypress/integration/search_for_records.ts
chrishutchinson: Ah yeah, Cypress fun! Docs here: https://docs.cypress.io/api/commands/type

It's a bit wild at times, I find myself checking the docs quite often hah!

.gitignore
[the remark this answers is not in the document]
chrishutchinson: Oh yeah, `cypress/support` is where the code for `cy.visitAs` lives, so needs to be in the repo 😄

cypress/integration/add_record.ts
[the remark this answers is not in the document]
Renny4Real: Nice! 🙌🏾

cypress/integration/search_for_records.ts
Renny4Real: 🙌🏾

Renny4Real (approved):
👍🏾

cypress/support/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Cypress {
+    interface Chainable {
+      visitAs: typeof visitAs;
+    }
+  }
+}
+
+/**
```
Renny4Real, line 32: This is so cool! Didn't know you could do this! 🎉

cypress/support/commands.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @param {AuthRoles} role The user role to act as for this request
+ * @param {VisitOptions} [options] Pass in an options object to change the default behavior of the underlying `cy.visit()` call
+ * @example
+ *    cy.visitAs('http://localhost:3000', AuthRoles.ChildrensGroup)
+ *    cy.visitAs('/somewhere' AuthRoles.AdultsGroup) // opens ${baseUrl}/somewhere
+ *    cy.visitAs('/' AuthRoles.AdminDevGroup, {
+ *      method: 'POST'
+ *    })
+ *
+ */
```
Renny4Real, line 45: This is so cool! Didn't know you could do this! 🎉

PR #286, opened 2021-03-30
UI upgrade

jhackett1:
- refactor the app to use hackney design system, and remove the direct dependency on `govuk-frontend`
- improvements to ux microcopy for flow and consistency (eg. consistent capitalisation and phrasing)
- accessibility improvements (eg. fixing elements that aren't currently tabbable, adding visually hidden labels to some buttons and links)

more minor things:
- align the expandable component with the new designs
- add a new `lbh-button--add` modifier class for buttons that add another of a repeatable component
- fix _some_ typescript errors and warnings
- add a branded 404 page

PR #351, opened 2021-05-17
refactor(SCT-182): add lbh design system, adapt header, radios and checkboxes to lbh

anabebs:
**What**  
- Added LBH design dependencies. 
- Adapted LBH Header to work with LBH styling
- Changed margins on checkbox and radio components 

**Why**  
To be in line with LBH styling, to unlock the other things we need to work on this sprint (bringing components across from the care recording tool, etc).

**Anything else?**
Header JS is covered in separate ticket.

This will almost certainly cause small ux/visual bugs which we will need to clear up in future prs. it was either this or make one monster pr ([which we already tried](https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/286)).


jhackett1 (approved):

chrishutchinson (approved):
Code looks good 👍 I haven't played around with the UI but I'm still getting familiar so my input there probably isn't especially helpful!

Could you add a link to the LBH Design System docs in the `README`?

PR #353, opened 2021-05-18
SCT-193 - Phone number validation blocking users from creating people

mapleleafjack:
Phone number input field validation error is now triggered when it contains white spaces. The backend should now accept all phone numbers coming from the frontend.

### DIFF
--- a/components/Form/PhoneInput/PhoneInput.tsx
+++ b/components/Form/PhoneInput/PhoneInput.tsx
@@ -13,8 +13,8 @@ const PhoneInput = ({ rules, ...props }: Props): React.ReactElement => (
       // this should be removed when the BE
       // fix the inconsistencies in phone numbers
       pattern: {
-        value: /^[\d\s+]+$/,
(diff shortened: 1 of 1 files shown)


components/Form/PhoneInput/PhoneInput.tsx
```diff
@@ -13,8 +13,8 @@ const PhoneInput = ({ rules, ...props }: Props): React.ReactElement => (
       // this should be removed when the BE
       // fix the inconsistencies in phone numbers
       pattern: {
-        value: /^[\d\s+]+$/,
-        message: 'Move text to phone type box',
+        value: /^[\d]+$/,
```
chrishutchinson, line 16: This now doesn't match `+447890123456`, which is possibly a valid input.

We should double check if we need to support country code.

Separately, I'd expect a test change to be in this PR alongside a code change.

chrishutchinson (changes requested):

PR #354, opened 2021-05-18
feat(SCT-206): read-only resident view

chrishutchinson:
**What**  
This makes the resident view read-only if the current user doesn't have appropriate permissions. There are three main areas, each with a distinct permission function:

* Edit resident – children only, user must match the type of the resident, and requires unrestricted access if the resident is restricted
* Allocate worker – user must match the type of the resident, must have allocation permissions (for adults)
* Case notes – user must match the type of the resident, and requires unrestricted access if the resident is restricted

**Note: there are new Cypress tests that allow us to check behaviour for a wider set of auth roles. This also means there are new env vars:**

* `CYPRESS_CHILDREN_RESTRICTED_RECORD_PERSON_ID` – record ID for a restricted child in staging
* `CYPRESS_ADULT_RESTRICTED_RECORD_PERSON_ID` – record ID for a restricted adult in staging
* `CYPRESS_TEST_KEY_ADULT_ALLOCATOR_GROUP` – JWT for a user in the adults and allocators group
* `CYPRESS_TEST_KEY_ADULT_UNRESTRICTED_GROUP` – JWT for a user in the adults and unrestricted group
* `CYPRESS_TEST_KEY_CHILDREN_UNRESTRICTED_GROUP` – JWT for a user in the childrens and unrestricted group

**Anything else?**

- Are any new environment variables or configuration values needed? YES!


components/AllocatedWorkers/AllocatedWorkers.spec.tsx
```diff
@@ -32,23 +33,30 @@ describe(`AddAllocatedWorker`, () => {
     }));
 
   const props = {
-    id: 123,
+    person: mockedResident,
   };
 
   it('should render only the table if not allocator', async () => {
```
Renny4Real, line 39: Could we rephrase this test? e.g. should only render allocated workers table if the user does not have allocation permissions?

components/AllocatedWorkers/AllocatedWorkers.tsx
```diff
@@ -7,14 +7,34 @@ import Button from 'components/Button/Button';
...
 }
 
-const AllocatedWorkers = ({ id }: Props): React.ReactElement => {
-  const { data: { allocations } = {}, error } = useAllocatedWorkers(id);
+const canUserAllocateWorkerToPerson = (user: User, person: Resident) => {
+  if (user.hasAdminPermissions || user.hasDevPermissions) {
+    return true;
+  }
+
+  if (user.hasChildrenPermissions && person.contextFlag === 'C') {
```
Renny4Real, line 21: Should we also check `(user.hasAllocationsPermissions || false)` here too?

pages/people/[id]/index.tsx
```diff
@@ -11,24 +11,29 @@ import Stack from 'components/Stack/Stack';
 import { useAuth } from 'components/UserContext/UserContext';
 import { User } from 'types';
 
+const canViewWarningNotes = (user: User) => {
```
Comment (Renny4Real) [inline pages/people/[id]/index.tsx:14]: Would this mean only users with `dev permissions` can view warning notes?

Renny4Real (approved):
Looks good. 👍🏾

components/AllocatedWorkers/AllocatedWorkers.tsx
chrishutchinson: The `hasAllocationPermissions` only applies to adults, for childrens anyone can allocate irrespective of this role. So nice and confusing, haha 😱

pages/people/[id]/index.tsx
Comment (chrishutchinson) [reply pages/people/[id]/index.tsx:14]: That's right – I actually just refactored a previous check that was already doing this inline here: https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/354/files#diff-b5ded0b0b90dbeccfcd0da5f2457c48db56fb45bcccbe174155aa7081e98ce43L29

components/AllocatedWorkers/AllocatedWorkers.spec.tsx
chrishutchinson: Great shout – I didn't change this test as part of this PR but it is an awful name. I will update it while I'm here 😄

components/AllocatedWorkers/AllocatedWorkers.tsx
chrishutchinson: I added a comment to clarify this @Renny4Real

components/AllocatedWorkers/AllocatedWorkers.spec.tsx
chrishutchinson: Done in https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/354/commits/bede9b7b84626171026a63095779d48f6f00561c

Renny4Real: 🙌🏾

components/AllocatedWorkers/AllocatedWorkers.tsx
Renny4Real: 🙌🏾

pages/people/[id]/index.tsx
Comment (Renny4Real) [reply pages/people/[id]/index.tsx:14]: I see 👍🏾

PR #355, opened 2021-05-18
SCT-196  Phone number validation blocking users from creating people

mapleleafjack:
Changed the RegExp so it doesn't let the user input a number with spaces + tests

### DIFF
--- a/components/Form/PhoneInput/PhoneInput.spec.tsx
+++ b/components/Form/PhoneInput/PhoneInput.spec.tsx
@@ -26,4 +26,22 @@ describe('PhoneInput', () => {
     });
     expect(newValue).toEqual('07849436768');
   });
+
+  it('Tests phone number with spaces', () => {
+    let newValue = '';
+    const myAction = jest.fn((e) => (newValue = e.target.value));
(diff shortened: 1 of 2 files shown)


components/Form/PhoneInput/PhoneInput.spec.tsx
```diff
@@ -26,4 +26,22 @@ describe('PhoneInput', () => {
     });
     expect(newValue).toEqual('07849436768');
   });
+
+  it('Tests phone number with spaces', () => {
+    let newValue = '';
+    const myAction = jest.fn((e) => (newValue = e.target.value));
+    const { getByLabelText, getByText } = render(
+      <PhoneInput name={'my-input'} label={'My Input'} onChange={myAction} />
```
chrishutchinson, line 34: I think you can remove `newValue` and `myAction`:

```ts
it('Tests phone number with spaces', () => {
  const { getByLabelText, getByText } = render(
    <PhoneInput name="my-input" label="My Input" />
  );
```

chrishutchinson (approved):
Looks good – just one comment

PR #356, opened 2021-05-18
compy with hackney sign in patterns

jhackett1:
our language and design patterns for signing in and out are now [consistent with the hackney design system](https://design-system.hackney.gov.uk/patterns/sign-in).

specifically, we use "sign in/out" language rather than "log in/out", and have improved the microcopy on the sign in page.

we've left the paths of these pages unmodified to avoid accidentally creating bugs.


anabebs (approved):
looks fine.

PR #357, opened 2021-05-18
fix(SCT-137): only show the next review date if available on a warning note

chrishutchinson:
**What**  

* Changes "Review date" to "Last review date"
* Hides "Last review date" if there is a "Next review date" available for the warning note
* Fixes some UI issues following the design system upgrade 

Screenshot 2021-05-19 at 14 23 58

**Why**  
[SCT-137](https://hackney.atlassian.net/browse/SCT-137)

**Anything else?**


components/WarningNote/WarningNotes.tsx
```diff
@@ -49,9 +49,9 @@ export const WarningBox = ({ notes, personId }: Props): React.ReactElement => {
                     </dd>
```
jhackett1, line 49: probably the role on this file needs to be changed from `alert` to `complementary`.

alerts are more for content that is dynamic and caused by user action: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role#accessibility_concerns

chrishutchinson: Done in https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/357/commits/5ac4d6fd7746ee54e6c3239c18aa0135555be897

Renny4Real (approved):
Looks good 👍🏾 
Although I wonder if `last review date`. 🤔 e.g. could `last review date` imply a deadline?

chrishutchinson:
> Looks good 👍🏾
> Although I wonder if `last review date` could be slightly confusing? 🤔 e.g. could `last review date` imply a deadline?

@Renny4Real this is a good point! checked with Ruksana and have updated that to be "Previous review date" now

PR #358, opened 2021-05-18
begin flexible form builder, with task list page

jhackett1:
this adds the foundations of the new flexible form builder that we will eventually use to run the vast majority of what are now google forms.

it adds:

- new routes under `/submissions` to avoid interfering with current live forms
- new components like `<TaskList/>`
- new api routes to power the forms
- new configuration data to describe the forms available for completion


stylesheets/TaskList.module.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.taskList {
+  list-style-type: none;
+  padding-left: 0;
+  @include govuk-responsive-margin(9, 'top');
+  // margin-top: 0;
+  margin-bottom: 0;
+  // @include govuk-media-query($from: tablet) {
+  //   min-width: 550px;
+  // }
```
JohnFarrellDev, line 11: Delete the commented out CSS?

pages/submissions/[id]/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <h1 className="lbh-heading-h1 govuk-!-margin-bottom-8">{form?.name}</h1>
+      <div className="govuk-grid-row">
+        <div className="govuk-grid-column-two-thirds">
+          <TaskListHeader
+            steps={form?.steps}
+            completedSteps={completedSteps}
+            onFinish={() => {
+              null;
+            }}
+          />
```
Comment (JohnFarrellDev) [inline pages/submissions/[id]/index.tsx:35]: We can remove the form null conditional operators.

form.? shouldn't be required as form is not optional.

pages/submissions/[id]/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <title>{form?.name} | Social care | Hackney Council</title>
+      </Head>
+      <h1 className="lbh-heading-h1 govuk-!-margin-bottom-8">{form?.name}</h1>
+      <div className="govuk-grid-row">
+        <div className="govuk-grid-column-two-thirds">
+          <TaskListHeader
+            steps={form?.steps}
+            completedSteps={completedSteps}
+            onFinish={() => {
+              null;
```
Comment (JohnFarrellDev) [inline pages/submissions/[id]/index.tsx:33]: Is this a temporary measure?

components/TaskList/TaskListHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+  steps,
+  completedSteps,
+  onFinish,
+}: Props): React.ReactElement => {
+  if (completedSteps?.length < steps.length)
+    return (
+      <>
+        <h2 className="lbh-heading-h4">Submission incomplete</h2>
+        <p className="lbh-body  govuk-!-margin-top-2">
+          You&apos;ve completed {completedSteps?.length || '0'} of{' '}
```
JohnFarrellDev, line 19: Is it not tidier to do it just like `You've completed ${completedSteps.length} of ${steps.length} sections.`

Is the `&apos` required? 
Also Completed steps can't be null so should always have a length.

components/TaskList/TaskList.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <ul className={s.items}>
+            {theme.steps.map((step) => (
+              <li className={s.item} key={step.id}>
+                <span className={s.taskName}>
+                  <Link href={`/submissions/${id}/steps/${step.id}`}>
+                    <a className="lbh-link">{step.name}</a>
+                  </Link>
+                </span>
+
+                {completedSteps.includes(step.id) ? (
```
JohnFarrellDev, line 37: Can there ever be a large number of steps?

This is going to be an O(N^2) operation looking up every step if it exists in completed steps. Maybe O(NxM) I guess.

If instead of a string[] for our completed steps we use a set<string> then we remove the inner loop.

Just a thought but if the steps/completed steps are never that large it won't matter.

lib/utils.ts
```diff
@@ -1,3 +1,26 @@
+import { Form, Step } from '../data/flexibleForms/forms.types';
+// import { Person } from "./socialCareApi.types"
```
JohnFarrellDev, line 2: can delete commented out line

pages/api/submissions/[id].ts
```diff
@@ -0,0 +1,42 @@
...
+    updatedAt: '2021-05-17T10:48:12.892Z',
+    submittedAt: null,
+    discardedAt: null,
+  };
+
+  const person = await getResident(submission.socialCareId, {
+    context_flag: 'A',
+  });
+
+  // 3. grab form
```
Comment (JohnFarrellDev) [inline pages/api/submissions/[id].ts:32]: is there a step 1 or 2 :p

jhackett1:
we're building this functionality at a new set of routes, so we can start gradually using it on an opt-in basis.

we can deploy this safely as it - it doesn't interfere with the old functionality and won't even be visible to users right now.

pages/submissions/[id]/index.tsx
Comment (jhackett1) [reply pages/submissions/[id]/index.tsx:33]: yes

components/TaskList/TaskList.tsx
jhackett1: i actually used a set for building the completedSteps array here: https://github.com/LBHackney-IT/social-care-recording/blob/main/lib/helpers.ts#L37. perhaps we can use that in more places

components/TaskList/TaskListHeader.tsx
jhackett1: &apos is needed by the linter - it complains about unescaped apostrophes.

i believe writing it like this was intended to handle completedSteps being null, as it initially could be in the other codebase. if it will always be an array we can change it.

JohnFarrellDev (approved):
:+1:

PR #360, opened 2021-05-18
Fix/css for hds

anabebs:
**What**  
Updates css on search pages to have correct margins and spacing. 
**Why**  
To move towards LBH design


components/Search/forms/SearchResidentsForm.tsx
```diff
@@ -40,7 +40,7 @@ const SearchResidentsForm = ({
       <div className="govuk-grid-row">
         <div className="govuk-grid-column-one-half">
           <TextInput
```
jhackett1, line 42: i think we need to tweak the classes on these a tiny bit more to get the spacing just right. copying the [example from the design system](https://design-system.hackney.gov.uk/components/input), i think the output should look like:

```
<div class="govuk-form-group lbh-form-group">
    <label for="first_name" class="govuk-label lbh-label">First name </label>
    <input id="first_name" data-testid="first_name" name="first_name" type="text" class="govuk-input lbh-input">
</div>
```

anabebs: Bit confused what you mean here. These are all defined in the TextInput.tsx component. Do you mean you'd like changes to that?

jhackett1: if that's the easiest way, i think so, yes

jhackett1:
also adds a new dev dependency on jest-cli because my (jaye) laptop no longer runs jest commands without it, not sure why.

components/Form/Autocomplete/Autocomplete.module.scss
```diff
@@ -16,27 +17,44 @@
...
+  // font-size: 1.1875rem;
+  // max-height: 300px;
+  // overflow-y: auto;
+  // border-bottom-left-radius: 4px;
+  // border-bottom-right-radius: 4px;
+  // list-style-type: none;
+  // margin: 0;
+  // padding: 0;
+  // z-index: 1;
+  // width: 414px;
```
JohnFarrellDev, line 45: Can we delete this commented out section

jhackett1: done

JohnFarrellDev (approved):
:+1:

PR #362, opened 2021-05-18
feat(SCT-180): enhanced 'restricted' UX

chrishutchinson:
**What**  
With the new, stricter permissions, more users are likely to see restricted records. This PR introduces two main changes:
* A new `[Restricted]` label in resident search results, alongside details of which service the resident is in (ACS or CFS)
Screenshot 2021-05-18 at 17 42 19

* Updated copy for the `RESTRICTED` message seen at the bottom of the person view page
Screenshot 2021-05-18 at 16 57 07

I have also removed some custom styling that was being applied to `govuk-tag`, in favour of the new `lbh-tag`

**Why**  
[SCT-180](https://hackney.atlassian.net/browse/SCT-180)

**Anything else?**


components/Cases/Cases.tsx
```diff
@@ -116,7 +87,7 @@ const CasesWrapper = ({ id, person }: WrapperProps): React.ReactElement => {
       ) : (
         <ErrorSummary
```
jhackett1, line 88: this component has `role='alert'` on it, which isn't correct for this kind of notice. it's intended for errors caused by user action, whereas this is more just informational/notice. [see mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role#accessibility_concerns)

chrishutchinson: @jhackett1 Done in https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/362/commits/dc88af1c87f10b87934be72cb625e7a25534c436 and https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/362/commits/08fe07377bd71da55e4ac7f48c96ce9b1b8a87a4

Renny4Real: Are `roles` solely for accessibility?

components/Search/results/ResidentsTable.spec.tsx
```diff
@@ -1,14 +1,47 @@
...
+          records={[
+            legacyResidentFactory.build({
+              ageContext: 'C',
+            }),
+          ]}
+        />
+      </UserContext.Provider>
+    );
+
+    getByText('RESTRICTED');
```
Renny4Real, line 45: Could I check my understanding? I'm not sure I understand how the assertion works in this test.

Are we defining a `getByText` component that sets up a user without child permissions and creates a list of residents under the children group? Why do we pass in `RESTRICTED` as text to `getByText`?

lib/permissions.spec.ts
```diff
@@ -0,0 +1,352 @@
...
+          }),
+          residentFactory.build({
+            restricted: 'N',
+            contextFlag: 'C',
+          })
+        )
+      ).toEqual(false);
+    });
+
+    it('should return false when the resident is an adult', () => {
```
Renny4Real, line 27: Would this mean that users would not be able to edit an adult person only if they are in adults? 
Or would this be regardless of permissions too? 🤔

Renny4Real (approved):
👍🏾 Looks good. Got some questions 😅

components/Search/results/ResidentsTable.spec.tsx
chrishutchinson: `getByText` is a method we get once we render our component, provided by React Testing Library – https://testing-library.com/docs/queries/bytext

What we're saying here is:
* Render the component
* Look for an element with the text "RESTRICTED" in it

`getByText` with throw an error if it doesn't find anything, which fails the test, so we don't explicitly need an `expect()` statement, but if we did it would look like:

```ts
expect(getByText("RESTRICTED").toBeTruthy();
```

lib/permissions.spec.ts
chrishutchinson: Adults currently can't be edited at all, no matter what your permission is, so that's what this test is confirming.

Aside: I gather ASC have actually just signed off adult editing, so we will probably be changing this shortly anyway, https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/365

components/Search/results/ResidentsTable.spec.tsx
Renny4Real: I see, thanks! 🙌🏾

PR #369, opened 2021-05-19
fix: permissions for viewing the allocate worker form

chrishutchinson:
**What**  
When viewing the add or remove allocated worker form, it wasn't doing a full permission check. This updates that logic to repurpose the logic we use for showing / hiding the "Allocated worker" button. Yay for consistency

**Anything else?**


lib/permissions.ts
```diff
@@ -60,3 +60,27 @@ export const canManageCases = (
...
+
+  if (user.hasChildrenPermissions && person.contextFlag === 'C') {
+    // Children's doesn't require `hasAllocationsPermissions`, as anyone can allocate in CFS
+    return true;
+  }
+
+  if (
+    user.hasAdultPermissions &&
+    person.contextFlag === 'A' &&
+    (user.hasAllocationsPermissions || false)
```
Renny4Real, line 80: Why do we do an OR operation here: `(user.hasAllocationsPermissions || false)`?

Wouldn't `user.hasAllocationsPermissions` be either `true` or `false`, and then when compared to `false`, the OR operation would return the value of `user.hasAllocationsPermissions`. Is this right? What does `|| false` help us to do or check for? 🤔

Renny4Real (approved):
🚀

lib/permissions.ts
chrishutchinson: Entertainingly the type def for `user.hasAllocationPermissions` is:

```ts
  hasAllocationsPermissions?: boolean;
```

meaning the potential values here are `true`, `false`, and `undefined`.

The `|| false` I've added effectively defaults it to false if it isn't set.

With that in mind, I actually think the type definition is wrong. I've created this ticket – https://hackney.atlassian.net/browse/SCT-233 – to refactor this type across the app.

chrishutchinson:
Closing this as it’s gotten a bit stale and doesn’t appear to have an associated ticket. I will review if we actually need it next sprint, and re-create a ticket if so.

PR #378, opened 2021-05-20
Feature/start submission/sct 219

jhackett1:
this makes it possible to begin a new submission using the new flexible form builder.

it includes:

- a new route `/pages/submissions/new.tsx` where the form is chosen. this will eventually replace the existing `/pages/people/[id]/records/index.tsx` page
- a few flexible form components
- new internal api endpoint which will eventually handle starting a fresh submission and providing a list of the possible forms to power the page

it is also a proof of concept that the older and new form methodologies can safely coexist in one codebase.


pages/api/submissions/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (req.method === 'POST') {
+    const { socialCareId, formId } = req.body;
+
+    // TODO: start a new submission here
+    // const newSubmission = await addCase({
+    //   socialCareId,
+    //   formId,
+    // });
+
+    // TODO: send back new case object
```
JohnFarrellDev, line 18: I really don't think we want commented sections of code.

https://kentcdodds.com/blog/please-dont-commit-commented-out-code

JohnFarrellDev:
I think you would really benefit (and the people reviewing/later using your code) from much more frequent smaller commits.

If I was to look through the commit messages I can't tell what this PR is about, they certainly don't tell a story.

Commit message of `add first form input` for a commit of 600+ LOC can't be accurately describing the changes introduced.

JohnFarrellDev:
Would it benefit us to use to create .stories.tsx for these pure components?

pages/submissions/new.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+      {forms && <StartForm onSubmit={handleSubmit} forms={forms} />}
+    </div>
+  );
+};
+
+export const getServerSideProps: GetServerSideProps = async ({ req }) => {
+  const protocol = getProtocol();
+
+  const res2 = await fetch(
```
JohnFarrellDev, line 46: Again using fetch which seems inconsistent within the codebase

jhackett1:
> I think you would really benefit (and the people reviewing/later using your code) from much more frequent smaller commits.
> 
> If I was to look through the commit messages I can't tell what this PR is about, they certainly don't tell a story.
> 
> Commit message of `add first form input` for a commit of 600+ LOC can't be accurately describing the changes introduced.

that's what i normally do, but here we're not writing these - we're copying and pasting whole organisms of code across until they work

pages/submissions/new.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const NewSubmissionPage = ({ forms }: Props): React.ReactElement => {
+  const router = useRouter();
+
+  const handleSubmit = async (
+    values: FormValues,
+    { setStatus }: FormikHelpers<FormValues>
+  ): Promise<void> => {
+    try {
+      const res = await fetch(`/api/submissions`, {
```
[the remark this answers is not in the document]
jhackett1: done

jhackett1:
> Would it benefit us to use to create .stories.tsx for these pure components?

yes, but struggle to run storybook locally because of https://hackney.atlassian.net/browse/SCT-91. it doesn't seem important at the moment.

pages/api/submissions/index.ts
jhackett1: so be it

pages/api/submissions/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    const { socialCareId, formId } = req.body;
+
+    // TODO: start a new submission here
+
+    // TODO: send back new case object
+    res.json({
+      id: 1,
+    });
+  } else {
+    res.json({
```
chrishutchinson, line 19: This would return this JSON if I sent a `DELETE` to this endpoint? We should probably be more explicit to only handle the HTTP methods we want, and 404 the rest.

JohnFarrellDev (approved):
:+1:

components/FlexibleForms/Banner.tsx
```diff
@@ -0,0 +1,14 @@
+interface Props {
+  title: string;
+  children: React.ReactChild | React.ReactChild[];
+  className?: string;
+}
+
+const Banner = ({ title, children, className }: Props): React.ReactElement => (
+  <section role="alert" className={`lbh-page-announcement ${className}`}>
```
chrishutchinson, line 8: Do we use the `classnames` `cx()` function for this type of interpolation?

```ts
cx("lbh-page-announcement", className)
```

components/FlexibleForms/SelectField.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    expect(screen.getByText('Foo option'));
+    expect(screen.getByText('Bar option'));
+  });
+
+  it('accepts an initial value/option', () => {
+    render(
+      <Formik
+        onSubmit={mockSubmit}
+        initialValues={{
+          foo: '1',
```
chrishutchinson, line 55: I think this should use `2` not `1`. 1 is the first in the list, so the select could just be showing the first value by default. Using `2` makes the test a little more robust

components/FlexibleForms/TextField.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface FieldProps {
+  touched: FormikTouched<FormikValues>;
+  errors: FormikErrors<FormikValues>;
+  name: string;
+  label: string;
+  type?: string;
+  hint?: string;
+  className?: string;
+  required?: boolean;
```
chrishutchinson, line 18: You have `required` in the type here but don't seem to use it?

components/FlexibleForms/TextField.tsx
```diff
@@ -0,0 +1,74 @@
...
+          {msg}
+        </p>
+      )}
+    </ErrorMessage>
+
+    <RawField
+      name={name}
+      id={name}
+      type={type}
+      className={`${
```
chrishutchinson, line 62: Same comment here about `classnames` and `cx()`

chrishutchinson (changes requested):

components/FlexibleForms/SelectField.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface FieldProps {
+  touched: FormikTouched<FormikValues>;
+  errors: FormikErrors<FormikValues>;
+  name: string;
+  label: string;
+  type?: string;
+  hint?: string;
+  className?: string;
+  required?: boolean;
```
chrishutchinson, line 18: Same here, this doesn't appear to be used?

components/FlexibleForms/SelectField.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </p>
+      )}
+    </ErrorMessage>
+
+    <RawField
+      as="select"
+      name={name}
+      id={name}
+      aria-describedby={hint ? `${name}-hint` : false}
+      className={`govuk-select lbh-select ${className}`}
```
chrishutchinson, line 63: Same comment here about `cx()`

components/StartForm/StartForm.test.tsx
```diff
@@ -0,0 +1,102 @@
+import StartForm from './StartForm';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { useRouter } from 'next/router';
+
+jest.mock('next/router');
+(useRouter as jest.Mock)
+  .mockReturnValueOnce({
+    query: {},
+  })
+  .mockReturnValueOnce({
```
chrishutchinson, line 10: I think these three mocks are for each of the tests? It might be better if each `mockReturnValue` call could be inside the actual test case, not mysteriously up here out of the way

pages/api/submissions/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import { addCase } from 'lib/cases';
+
+const handler = async (
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> => {
+  if (req.method === 'POST') {
+    const { socialCareId, formId } = req.body;
+
+    // TODO: start a new submission here
```
chrishutchinson, line 12: Can we put ticket numbers (or something) to indicate how these TODOs will be sorted?

Otherwise I'd suggest this shouldn't be in this PR yet if it's not ready

pages/submissions/new.tsx
```diff
@@ -0,0 +1,56 @@
...
+import axios from 'axios';
+
+interface Props {
+  forms: Form[];
+}
+
+const NewSubmissionPage = ({ forms }: Props): React.ReactElement => {
+  const router = useRouter();
+
+  const handleSubmit = async (
```
chrishutchinson, line 16: I think we want a `useCallback` here

chrishutchinson (changes requested):

components/FlexibleForms/TextField.tsx
jhackett1: see above

components/FlexibleForms/TextField.tsx
jhackett1: holdover from an older attempt at validation

pages/api/submissions/index.ts
jhackett1: this will be a big piece of work in its own right and will probably bloat this pr. the work doesn't have a number yet

components/FlexibleForms/Banner.tsx
chrishutchinson: Because `className` is optional (as per your `Props`) you can end up with `"undefined"` in your class list. Best to avoid that if we can, it can get messy. `classnames` prevents that by automatically filtering them out

pages/submissions/new.tsx
jhackett1: i understand callbacks and memos as a performance optimisation to prevent them from expensively re-running or causing needless re-renders.

i don't think either of those apply here - the function only runs when you hit the button (once). [this is close to my understanding](https://dmitripavlutin.com/dont-overuse-react-usecallback/)

pages/api/submissions/index.ts
jhackett1: this is actually closer to next.js's suggestions in their docs - the switch statements we have in other places i find pretty verbose and harder to read, but eh 🤷‍♀️

pages/submissions/new.tsx
chrishutchinson: Yeah that's a fair point, I thought I had a specific reason for this one but now can't think what that was. Perhaps it was depending on whether `props.forms` changes frequently (as it's passed down to a child, I don't know how likely to change that is. If you think it's fairly static then I think fair to keep it as-is for now.

pages/submissions/new.tsx
```diff
@@ -0,0 +1,56 @@
...
+    values: FormValues,
+    { setStatus }: FormikHelpers<FormValues>
+  ): Promise<void> => {
+    try {
+      const { data } = await axios.post(`/api/submissions`, {
```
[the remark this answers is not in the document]
jhackett1: that's handled in the form itself - we use the `isSubmitting` state

components/FlexibleForms/Banner.tsx
jhackett1: done

pages/api/submissions/index.ts
jhackett1: done

components/StartForm/StartForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+  socialCareId: string;
+  formId: string;
+}
+
+interface Props {
+  forms: FormType[];
+  onSubmit: (
+    values: FormValues,
+    { setStatus }: FormikHelpers<FormValues>
+  ) => void;
```
chrishutchinson, line 21: If I've followed this correctly I _think_ this return type should be `Promise<void> | void`. That is based on [this](https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/378/files#diff-f56e9f8a1e1be2f77a6fc2c06eca3f260f4c84981288791c2d430870df09c5a8R17) func which is passed into a `<StartForm />`.

pages/api/submissions/index.ts
```diff
@@ -0,0 +1,30 @@
...
+    case 'POST':
+      // TODO: start a new submission here
+      // TODO: send back new case object
+      res.json({
+        id: 1,
+      });
+      break;
+    case 'GET':
+      res.json({
+        forms,
```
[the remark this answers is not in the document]
chrishutchinson: Ah, I see what you have in your PR notes which answers this q:
>new internal api endpoint which will eventually handle starting a fresh submission and providing a list of the possible forms to power the page

I still think `/api/submissions` is a bad name (because it doesn't return submissions). Maybe we could think of something more descriptive for the actual data you get back?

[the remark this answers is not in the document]
jhackett1: no, it will return submissions eventually - and probably people to boot. these apis don't just return one domain/model. they'll return the right mix of things that the page needs to render everything quickly.

components/StartForm/StartForm.tsx
jhackett1: done

pages/api/submissions/index.ts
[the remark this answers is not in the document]
jhackett1: plus, mirroring the routing structure of the frontend is super helpful when debugging. you can just drop the `/api` prefix in and inspect exactly what data the page is getting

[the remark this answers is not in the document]
jhackett1: you're only seeing an incomplete, stunted version of these endpoints at the moment. they'll look like this when done: https://github.com/LBHackney-IT/social-care-recording/tree/main/pages/api/submissions

anabebs (approved):

PR #379, opened 2021-05-20
SCT-223  Unable to edit people if you don't have admin permissions

mapleleafjack:
Bug fix that makes sure that the "restricted" parameter is in the form data


chrishutchinson (approved):

pages/people/[id]/edit/[[...stepId]].tsx
```diff
@@ -39,6 +39,7 @@ const UpdatePerson = (): ReactElement => {
       contextFlag: formData.contextFlag || user.permissionFlag,
       nhsNumber: Number(formData.nhsNumber),
       createdBy: user.email,
+      restricted: formData.restricted || person?.restricted,
```
Comment (JohnFarrellDev) [inline pages/people/[id]/edit/[[...stepId]].tsx:42]: Can we have a situation when formData object doesn't have a restricted key and also person is undefined?

PR #380, opened 2021-05-20
improve "go back" button layout

jhackett1:
at the moment, "go back" buttons are rendered _inside_ the page's layout wrapper. this has the effect of moving them too far down from the header.

they _should_ be rendered higher - right below the header's markup in fact, in the "breadcrumbs go here" area: https://design-system.hackney.gov.uk/developing/page-template

this is tricky to do in react, because we want to be able to wrap all our pages in a common layout component, which would be the obvious place to put the main layout wrapper element.

this pull request solves this issue by:
- setting a `goBackButton` boolean value on the page component (true for a back button, false by default)
- passing that value as a prop from next's `_app.js` file
- having the top-level `<Layout/>` component render the button if the relevant prop is true

it's a pretty elegant solution i've seen used in other places and took from: https://reacttricks.com/nested-dynamic-layouts-in-next-apps/

it could be enhanced in future to render all sorts of content outside the main layout wrapper, like warnings and flash messages.

after merging this, you should notice that go back links sit closer to the header, as they should.


anabebs (commented):
Looks good. I don't know if the smooth scrolling was your doing but looks good anyway.

jhackett1:
> Looks good. I don't know if the smooth scrolling was your doing but looks good anyway.

no, smooth scrolling was there already

PR #381, opened 2021-05-20
feature(SCT-248): Todays date by default for case notes

jhackett1:
this makes it easier/faster to add a case note to a person by:

- using `<input type="date"/>` rather than the more complicated "three text inputs" style
- enhances the `<DatePicker/>` element with a prop that causes it to render with today's date as an intial value
- reconfigure the form configuration data to use that prop for the case note form

### DIFF
--- a/components/Form/DatePicker/DatePicker.tsx
+++ b/components/Form/DatePicker/DatePicker.tsx
@@ -1,9 +1,27 @@
 import { TextInput } from '..';
+import { TextInputNoType } from 'components/Form/types';
 
-import type { TextInputNoType as Props } from 'components/Form/types';
+import { format } from 'date-fns';
 
-const DateInput = (props: Props): React.ReactElement => (
-  <TextInput width={10} {...props} type="date" />
-);
+format(new Date(2014, 1, 11), 'MM/dd/yyyy');
+
+interface Props extends TextInputNoType {
+  defaultToday?: boolean;
+}
+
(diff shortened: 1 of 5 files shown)


anabebs:
@jhackett1  why does this have a help wanted tag?

components/Form/DatePicker/DatePicker.tsx
```diff
@@ -1,9 +1,18 @@
...
+}
 
-const DateInput = (props: Props): React.ReactElement => (
-  <TextInput width={10} {...props} type="date" />
+const DateInput = ({ defaultToday, ...props }: Props): React.ReactElement => (
+  <TextInput
+    width={10}
+    {...props}
+    type="date"
+    defaultValue={defaultToday ? format(new Date(), 'yyyy-MM-dd') : undefined}
```
JohnFarrellDev, line 14: This is more of a question, with the format yyyy-MM-dd I get the display would be 2021-05-27 for today, does the date information still contain the time and could that also be changed.

I know we're trying to target the ISO-8061 standard more often for consistent datetime so am keen to not trim some of the time data.

jhackett1: this is only a `date` picker. there's a seperate `time` element that we might have use for in future.

it's a text input with extra trimmings. the underlying text format is always yyyy-mm-dd, but the user's browser and operating system settings control exactly how it renders.

so in short, it holds no time information at all

JohnFarrellDev (approved):
:+1:

PR #382, opened 2021-05-21
Fixes/Styling issues

Miles-Alford:
**What**  
Fixes some styling issues which have come out from the addition of the design library, mainly the `Select` and the `WorkerRecap` components 

**Why**  
To fix some styling issues relating to the introduction of the design system.

**WorkerRecap Component**
image

**Select Component**
image

### DIFF
--- a/components/Form/Select/Select.module.scss
+++ b/components/Form/Select/Select.module.scss
@@ -1,0 +1,8 @@
+@import 'lbh-frontend/lbh/base';
+
+.inputContainer {
+  position: relative;
+  display: flex;
+  max-width: 48ex;
+  margin-top: 0;
+}
--- a/components/Form/Select/Select.tsx
+++ b/components/Form/Select/Select.tsx
@@ -39,7 +39,7 @@ const Select = ({
         }
       )}
     >
-      <label className={`lbh-label`} htmlFor={name}>
+      <label className={`govuk-label lbh-label`} htmlFor={name}>
         {label} {required && <span className="govuk-required">*</span>}
       </label>
       <div className={styles.inputContainer}>
--- a/components/Form/Select/Select.tsx
+++ b/components/Form/Select/Select.tsx
@@ -4,6 +4,8 @@ import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
 
 import type { Select as Props } from 'components/Form/types';
 
+import styles from './Select.module.scss';
+
 const Select = ({
   label,
   labelSize = 'm',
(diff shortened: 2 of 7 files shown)


components/WorkerView/__snapshots__/WorkerSearch.spec.tsx.snap
```diff
@@ -52,24 +52,26 @@ exports[`WorkerSearch should display search results 1`] = `
...
+    >
       <h2
-        class="gov-weight-lighter"
+        class="lbh-heading-h2 gov-weight-lighter title"
       >
         Worker's details found in system
       </h2>
       <p
-        class="govuk-body"
+        class="lbh-body govuk-body"
```
jhackett1, line 64: you don't need the extra `govuk-body` class for paras. in fact, you might not need any classes at all since i think we're using global styles

components/WorkerView/WorkerRecap.tsx
```diff
@@ -1,21 +1,23 @@
...
   firstName,
   lastName,
   role,
   teams,
 }: Worker): React.ReactElement => {
   return (
-    <div>
-      <h2 className="gov-weight-lighter">
+    <div className={style.detailsContainer}>
+      <h2 className={`lbh-heading-h2 gov-weight-lighter ${style.title}`}>
```
jhackett1, line 13: `gov-weight-lighter` is no longer needed here - we've consolidated all font weights down to just two, which are correctly applied by default.

components/WorkerView/WorkerRecap.tsx
```diff
@@ -1,21 +1,23 @@
...
         Worker&apos;s details found in system
       </h2>
-      <p className="govuk-body">
+      <p className="lbh-body govuk-body">
```
jhackett1, line 16: you don't need the extra govuk-body class for paras. in fact, you might not need any classes at all since i think we're using global styles

components/Form/Select/Select.module.scss
```diff
@@ -0,0 +1,8 @@
+@import 'lbh-frontend/lbh/base';
+
+.inputContainer {
+  position: relative;
+  display: flex;
+  max-width: 48ex;
```
jhackett1, line 6: i've not come across an `ex` unit before? should it be `px`?

components/Form/Select/Select.tsx
```diff
@@ -30,50 +32,55 @@ const Select = ({
...
-      })}
+      className={cx(
+        `govuk-form-group lbh-form-group govuk-grid-column-${govGrid}`,
+        {
+          'govuk-form-group--error': error,
+        }
+      )}
     >
-      <label className={`lbh-label govuk-label--${labelSize}`} htmlFor={name}>
+      <label className={`lbh-label`} htmlFor={name}>
```
jhackett1, line 42: to get the spacing perfect, this may need one more class on the label. refer to: https://design-system.hackney.gov.uk/components/select/

jhackett1 (approved):

components/Form/Select/Select.module.scss
Miles-Alford: `px` would be too small of a max width and would cause the select's to be tiny, haven't used `ex` myself but it has been used elsewhere such as in the autocomplete

PR #386, opened 2021-05-22
Feature/step forms sct 141

jhackett1:
this brings across all the relevant code from the care recording app necessary to fulfil https://hackney.atlassian.net/browse/SCT-219.

it is not yet demonstrable end-to-end because it requires new api endpoints, which will be delivered as https://hackney.atlassian.net/browse/SCT-140. so, our definition of complete for this pr is "all the ui for steps in a form renders correctly, without errors".

it includes:
- formik components for all the flexible fields a form may consist of (text boxes, selects, comboboxes, repeater groups and so on). in future we might refactor these to avoid duplication with some components already in the codebase, but we decided to forego this for now in the interest of speed: https://hackney.atlassian.net/browse/SCT-225
- a page for each step in a form: `/pages/submissions/[id]/steps/[stepId].tsx`
- new mock api endpoints under `/pages/api/submissions` which will eventually allow saving of in-progress submissions.


components/FlexibleForms/CheckboxField.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <ErrorMessage name={name}>
+        {(msg) => (
+          <p className="govuk-error-message lbh-error-message" role="alert">
+            <span className="govuk-visually-hidden">Error:</span>
+            {msg}
+          </p>
+        )}
+      </ErrorMessage>
+
+      <div className={`govuk-checkboxes lbh-checkboxes ${className}`}>
```
JohnFarrellDev, line 60: Look to use the cx library here

components/FlexibleForms/ComboboxField.tsx
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
```
JohnFarrellDev, line 1: This would be a lot easier to review if a storybook component was included.

It's fairly complex to review this code without seeing it in action somewhere.

components/FlexibleForms/ComboboxField.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+              hint="Hint text"
+              choices={choices}
+            />
+          </Form>
+        )}
+      </Formik>
+    );
+    expect(screen.getByText('Example error'));
+  });
+});
```
JohnFarrellDev, line 106: I think we could do with a test for the logic from line 116 to 138 in `ComboBoxField.tsx`, my understanding is clicking a button reveals an unordered list of items which is then filtered based on input text.

A test would clarify the intention of the code as well as ensuring it works as expected.

components/FlexibleForms/FlexibleFields.tsx
```diff
@@ -0,0 +1,134 @@
...
+  errors: FormikErrors<FormikValues>;
+}
+
+const FlexibleField = ({
+  values,
+  field,
+  touched,
+  errors,
+}: Props): React.ReactElement | null => {
+  if (field.condition && values[field.condition.id] !== field.condition.value)
```
JohnFarrellDev, line 24: Could I have some clarification on the purpose of this if statement

components/FlexibleForms/FlexibleFields.test.tsx
```diff
@@ -0,0 +1,86 @@
+import FlexibleFields from './FlexibleFields';
+import { Formik } from 'formik';
+import { render, screen } from '@testing-library/react';
+
+const mockSubmit = jest.fn();
+
+describe('TextField', () => {
```
JohnFarrellDev, line 7: We're only testing TextField renders correctly but there are quite a few different field options, do we need to test those render as expected.

components/FlexibleForms/RadioField.tsx
```diff
@@ -0,0 +1,98 @@
...
+        {(msg) => (
+          <p className="govuk-error-message lbh-error-message" role="alert">
+            <span className="govuk-visually-hidden">Error:</span>
+            {msg}
+          </p>
+        )}
+      </ErrorMessage>
+
+      <div
+        className={`govuk-radios lbh-radios govuk-!-margin-top-3 ${className}`}
```
JohnFarrellDev, line 63: Use cx here

components/FlexibleForms/FlexibleFields.tsx
jhackett1: it makes conditional fields, that only show up based on a previous answer, work. for example:

- Do you take any medications? (yes/no radio)
- List your prescriptions (textarea)

components/FlexibleForms/ComboboxField.test.tsx
jhackett1: good catch! added a new test case

components/FlexibleForms/FlexibleFields.test.tsx
jhackett1: each of those fields is individually tested - doing it again here seems like unnecessary duplication?

jhackett1:
noticed that the `npm test` script expects things to all have the `spec.tsx` filename structure

components/FlexibleForms/ComboboxField.tsx
jhackett1: i agree. are we anywhere with working out that storybook bug? i still can't run it on my computer. i can dig into it but can't estimate how long it will take

components/FlexibleForms/FlexibleFields.test.tsx
JohnFarrellDev: What I am looking for is that under certain conditions we choose to render the correct "Field" such as a TextField or a RadioButton etc.

jhackett1: it'll be one `it()` for every field, probably. a bit verbose. that okay?

components/FlexibleForms/RepeaterField.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    onClick={() => remove(i)}
+                  >
+                    Remove
+                  </button>
+                </div>
+              ))}
+
+              <button
+                type="button"
+                onClick={() => push('')}
```
JohnFarrellDev, line 95: Why do we call push with an empty string

components/FlexibleForms/RepeaterField.test.tsx
```diff
@@ -0,0 +1,123 @@
+import RepeaterField from './RepeaterField';
+import { Formik, Form } from 'formik';
+import { render, screen } from '@testing-library/react';
+
+const mockSubmit = jest.fn();
+
+describe('RepeaterField', () => {
```
JohnFarrellDev: I think we are missing a couple of tests regarding the functionality for adding/removing fields.

components/FlexibleForms/RepeaterGroupField.test.tsx
```diff
@@ -0,0 +1,144 @@
+import RepeaterGroupField from './RepeaterGroupField';
+import { Formik } from 'formik';
+import { render, screen } from '@testing-library/react';
+
+const mockSubmit = jest.fn();
+
+describe('RepeaterGroupField', () => {
```
JohnFarrellDev: We could do with a test/s for adding.removing fields.

components/FlexibleForms/FlexibleFields.test.tsx
jhackett1: done

components/FlexibleForms/RepeaterField.test.tsx
jhackett1: that feels like testing formik internals to me, no? there's no state or anything involved, it all comes from formik's hooks

lib/utils.ts
```diff
@@ -28,3 +35,40 @@ export const truncate = (str: string, noWords: number): string => {
     return str;
   }
 };
+
+const noPrefill = new Set(['file']);
+const singlePrefills = new Set(['select']);
+const iterablePrefills = new Set(['checkboxes', 'repeated']);
+
+/** Generate flexible initial values for a flexible schema */
+export const generateInitialValues = (
```
JohnFarrellDev, line 44: A test for this functionality would be very helpful

lib/validators.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      three: ['example', 'example 2'],
+      four: ['example', 'example 2'],
+      five: 'value',
+      six: 'value',
+    });
+
+    expect(result).toBeTruthy();
+  });
+
+  it('handles required fields', async () => {
```
JohnFarrellDev, line 50: can we see a test for require passing

package.json
```diff
@@ -19,6 +19,7 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
+    "@testing-library/user-event": "^13.1.9",
```
JohnFarrellDev, line 22: Should this be a dev dependency?

jhackett1: it should!

lib/validators.test.ts
jhackett1: i'm not sure what you mean. you mean a test case where it doesn't throw an error?

lib/utils.ts
jhackett1: moved the tests over and fixed a bug we introduced when we refactored the types

lib/validators.test.ts
JohnFarrellDev: Yes so that when a field is required and we give it a response no errors are thrown.

jhackett1: done

components/FlexibleForms/RepeaterGroupField.test.tsx
jhackett1: done

components/FlexibleForms/RepeaterField.test.tsx
jhackett1: changed my mind - can't hurt.

JohnFarrellDev (approved):
I'm going to approve.

In full transparency there is a lot here and without being able to see it in action it's hard to grasp if this definitely is meeting our requirements.

Looking at the code I feel happy with what I see, I'll just be a lot more confident when I see the output.

PR #387, opened 2021-05-24
SCT-28  Decide when to update frontend to use the service API in new …

mapleleafjack:
Frontend bits for maintenance mode.

This needs a NEXT_PUBLIC_MAINTENANCE_MODE=1 variable in the .env file

For now the "blocked" pages are listed manually.
I tried adding a wildcard   - source: '/((?!maintenance).*)', -  but this prevents the maintenance page into loading the various resources it needs as they would be redirect as well.

The change in the header makes it not appear if it's in maintenance mode.
Everything else is routing.

Wording will probably need to be changed.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -4,4 +4,56 @@ module.exports = {
   future: {
     webpack5: true,
   },
+  async redirects() {
+    let redirects = maintenanceMode();
+    return redirects.filter(Boolean);
+  },
 };
+
+function maintenanceMode() {
+  let maintenance_array = [];
+
+  if (
+    process.env.NEXT_PUBLIC_MAINTENANCE_MODE &&
+    process.env.NEXT_PUBLIC_MAINTENANCE_MODE === '1'
+  ) {
+    let pages = [
+      '/access-denied',
+      '/forms-in-progress',
+      '/login',
+      '/logout',
+      '/index',
+      '/search',
+      '/my-records',
+      '/workers',
+      '/workers/:slug*',
+      '/api/:slug*',
+      '/people',
+      '/people/:slug*',
+      '/submissions:slug*',
+    ];
+
+    pages.forEach((elm) => {
+      maintenance_array = [
+        ...maintenance_array,
+        {
+          source: elm,
+          basePath: false,
+          destination: '/maintenance',
+          permanent: false,
+        },
+      ];
+    });
+  } else {
+    maintenance_array = [
+      ...maintenance_array,
+      {
+        source: '/maintenance',
+        destination: '/',
+        basePath: false,
+        permanent: false,
+      },
+    ];
+  }
+  return maintenance_array;
+}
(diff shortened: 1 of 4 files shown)


pages/maintenance.tsx
```diff
@@ -0,0 +1,14 @@
+import Seo from 'components/Layout/Seo/Seo';
+
+const Maintenance = (): React.ReactElement => (
+  <>
+    <Seo title="Maintenance mode" />
+    <h1>Maintenance mode</h1>
```
jhackett1, line 6: should probably closely mirror content here: https://design-system.service.gov.uk/patterns/service-unavailable-pages/

chubberlisk: Agree with @jhackett1!

chubberlisk (commented):
Sorry Jack, I can't review the implementation and how it's been done but I do know that we've got to update the serverless.yml to have the new environment variable. It might be fine to just hardcode the value as it's temporary rather than adding in a new SSM value.

Could we also add in a Cypress test for this?

https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/0c9de86e544f0614d95a17a5345f01a49233633f/serverless.yml#L38-L54

mapleleafjack:
Hi @jhackett1

I've edited the header so it should follow the gov guidelines - for now there's no description (the wording "Sorry, the service is unavailable" is there) - let me know if you want me to add something in the description.

@chubberlisk 
I've added the variable in the serverless.yml file (hardcoded "0" value)

mapleleafjack:
I'm not sure if I can add a Cypress test for this case: the NextJS redirects needs the application to be reloaded to be effective.
I wrote a test but it will pass only if the maintenance mode is activated (while all other tests will probably fail)

In case, I'll be happy to be proven wrong :)

next.config.js
```diff
@@ -4,4 +4,56 @@ module.exports = {
...
+      '/logout',
+      '/index',
+      '/search',
+      '/my-records',
+      '/workers',
+      '/workers/:slug*',
+      '/api/:slug*',
+      '/people',
+      '/people/:slug*',
+      '/submissions:slug*',
```
chubberlisk, line 33: Is this missing a slash? 🕵️ 

```suggestion
      '/submissions/:slug*',
```

pages/maintenance.tsx
```diff
@@ -0,0 +1,11 @@
+import Seo from 'components/Layout/Seo/Seo';
+
+const Maintenance = (): React.ReactElement => (
+  <>
+    <Seo title="Maintenance mode" />
+    <h1>Sorry, the service is unavailable</h1>
+    <p className="govuk-body"></p>
```
chubberlisk, line 7: I think it's worth still have some content. Maybe:

```suggestion
    <p className="govuk-body">You will be able to use the service from 8pm on Tuesday 24 May 2021.</p>
```

chubberlisk (commented):
No worries about the Cypress test, I wasn't sure myself tbh!

chubberlisk (approved):

PR #391, opened 2021-05-25
feat(SCT-228): Create warning note details pages for reviewed & ended warning notes

Renny4Real:
**N.B:** See this [user story](https://hackney.atlassian.net/browse/SCS-510) for a more detailed user journey for reviewing warning note.

**What**  
This adds a page to view details of the state of a warning note from the records history section. i.e when the user clicks `view` next to a **Warning Note Reviewed** record:
Screenshot 2021-05-25 at 11 43 51 am

**Before:**
Clicking the view link would take the user to a page that only showed the details of the initial warning note and did not display any review/end information:
Screenshot 2021-05-25 at 11 45 51 am

**After:**
Now, clicking the view link would take the user to a page displaying the details of all reviews submitted for the warning note as well as the original warning note itself:

**Warning Note Reviewed View**
Screenshot 2021-06-02 at 6 03 16 pm

**Warning Note Ended View**
Screenshot 2021-06-02 at 6 03 31 pm

The reviews are displayed, starting with the most recent review submitted.

The fields `next review date` (if looking at reviews for an open warning note) or `end date` (if looking reviews for a closed warning note) are only displayed on the most recent review as these are properties that are updated on the original warning note every time a review is done and not saved on the reviews themselves.

**Why**  
When a warning note is reviewed or ended, a log of this action is made in the records' history section. The view link takes the user to a warning note details page to allow the user to see all the reviews done on a specific warning note or any notes submitted when the warning note was closed.


components/Cases/CaseLink.tsx
```diff
@@ -32,7 +32,7 @@ const getLink = (
     case 'API_WarningNote':
       return `/people/${caseFormData.mosaic_id}/warning-notes/${
         (caseFormData as WarningNoteCaseFormData).warning_note_id
-      }/view`;
+      }/view?form_name=${caseFormData.form_name}`;
```
chrishutchinson, line 35: It'd be good if you could take me through this – I'm not sure the reason for passing the form name as a query param here?

components/WarningNote/WarningNoteRecap/WarningNoteRecap.tsx
```diff
@@ -24,8 +27,34 @@ const WarningNoteRecap = ({
     return <Spinner />;
   }
 
+  const displayReviews =
+    Array.isArray(warningNote.warningNoteReviews) &&
+    warningNote.warningNoteReviews.length > 0 &&
+    formName != 'Warning Note Created';
```
chrishutchinson, line 33: I think this may need to be `!==`, but even better this feels like something that might work better as a separate function, something like:

```ts
const displayReviews = shouldDisplayReviews(warningNote, formName); // returns `true` or `false`
```

components/WarningNote/WarningNoteRecap/WarningNoteRecap.tsx
```diff
@@ -24,8 +27,34 @@ const WarningNoteRecap = ({
...
+
   return (
     <>
+      {displayReviews &&
+        warningNote.warningNoteReviews
+          .slice(0)
```
chrishutchinson, line 39: What's the `.slice(0)` here doing?

Renny4Real: I saw this being used in a solution for returning an array in the reverse order.
My understanding is that `.slice()` would make a shallow copy of the array starting and ending from any indices provided. I'm not sure if it's entirely needed here, but I wondered if maybe having a shallow copy of the reviews array to manipulate was better than acting on the actual array. I might also be misunderstanding the intention behind having shallow copies and the `.slice()` function.

Renny4Real: Based on our conversation, I will be refactoring this to be more readable and clearer on its intent. 🚀 😁

chrishutchinson (approved):
This is really nice, great work @Renny4Real!

PR #410, opened 2021-05-27
feat(SCT-179): show all records in search results

chrishutchinson:
Putting this back on staging for Felix to test


Renny4Real (approved):
👍🏾

PR #413, opened 2021-05-27
SCT-204 - Write front end views for displaying relationships

mapleleafjack:
[Ticket](https://hackney.atlassian.net/browse/SCT-204)

**What**  
Adding dev-only view parameter for relationships.
canViewWarningNotes function has been renamed to canViewArea so it applies also to Relationship (other than warning notes)
The function itself evaluates only the user.hasDevPermissions, so it applies in this case to both

**Why**  
To make the section visible to developers only in staging


JohnFarrellDev (approved):
:rocket:

PR #415, opened 2021-05-27
feature(SCT-83): google form import job

JohnFarrellDev:
**What**  

[Ticket](https://hackney.atlassian.net/browse/SCT-83)

**Why**  

The idea is that we can ingest a google form with these scripts, output a json format that we then ingest via a flexible form builder to create react companents that utilise formik for the component and yup for validation.

**Anything else?**


data/flexibleForms/index.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const formData: Form[] = [];
+
+for (let i = 0; i < allForms.length; i++) {
+  const form = allForms[i] as unknown as Form;
+
+  formData.push(form);
+}
+
+export default formData;
```
jhackett1, line 12: is this the file we should be using to grab all the forms now? i notice that further down one of the json files is being directly imported

pages/api/submissions/[id].ts
```diff
@@ -1,5 +1,5 @@
 import { NextApiRequest, NextApiResponse } from 'next';
-import forms from 'data/flexibleForms/forms';
+import forms from 'data/flexibleForms/forms.json';
```
Comment (jhackett1) [inline pages/api/submissions/[id].ts:2]: should this be the typescript file?

jhackett1:
i can see that the review3c.json data file exists, but it doesn't look like it's being imported/used anywhere

jobs/transform-google-form.js
```diff
@@ -0,0 +1,83 @@
...
+
+const slugify = (string) => rawSlugify(string, { lower: true });
+
+// trawl through the fields, find google's SECTION_HEADERS and convert them to our steps
+const pullOutSteps = (allFields) => {
+  let newArray = [];
+  let currentStep = 0;
+
+  allFields.forEach((field) => {
+    if (field.type === !field[0] && 'PAGE_BREAK') {
```
jhackett1, line 13: this is the line we changed, right? why does this work?

jhackett1 (approved):

PR #418, opened 2021-05-28
Add functions to call the api. Write tests.

anabebs:
**What**  
This PR enables the app to speak to the API when starting, updating and finishing submissions. It follows the pattern other files like lib/cases have outlined as standard. All 4 functions are tested. 
Related to this card https://hackney.atlassian.net/secure/RapidBoard.jspa?rapidView=59&modal=detail&selectedIssue=SCT-250

**Why**  
Once these functions are ready they will be called up by in the api/submissions routes.

### DIFF
--- a/components/WarningNote/WarningNotes.spec.tsx
+++ b/components/WarningNote/WarningNotes.spec.tsx
@@ -25,7 +25,7 @@ describe(`useWarningNotes`, () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
-  it('should not render the previous review date if there is a scheduled future review date', () => {
+  it('should not render the review date if there is a scheduled future review date', () => {
     jest.spyOn(warningNotes, 'useWarningNotes').mockImplementation(() => ({
       data: [
         warningNoteFactory.build({
@@ -40,6 +40,10 @@ describe(`useWarningNotes`, () => {
     const { getByText, queryByText } = render(<WarningNotes id={123} />);
 
     getByText('Next review date');
-    expect(queryByText('Previous review date')).toBeNull();
+    expect(
+      queryByText('Review date', {
+        exact: true,
+      })
+    ).toBeNull();
   });
 });
--- a/components/WarningNote/WarningNotes.tsx
+++ b/components/WarningNote/WarningNotes.tsx
@@ -51,7 +51,7 @@ export const WarningBox = ({ notes, personId }: Props): React.ReactElement => {
                 )}
                 {note.reviewDate && !note.nextReviewDate && (
                   <>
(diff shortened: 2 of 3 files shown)


lib/submissions.ts
```diff
@@ -0,0 +1,73 @@
...
+  return data;
+};
+
+export const finishSubmission = async (
+  submissionId: string
+): Promise<Submission> => {
+  const { data } = await axios.patch(
+    `${ENDPOINT_API}/submissions/${submissionId}`,
+    {
+      submittedAt: new Date(),
```
jhackett1, line 66: what is the correct format we should send the date in?

lib/submissions.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+export const patchSubmissionForStep = async (
+  submissionId: string,
+  stepId: string,
+  editedBy: string,
+  stepAnswers: StepAnswers
+): Promise<Submission> => {
+  const { data } = await axios.patch(
+    `${ENDPOINT_API}/submissions/${submissionId}/steps/${stepId}`,
```
jhackett1, line 48: this seems the most sensible way to provide the `stepId` - it feels confusing to put it in the body because it doesn't end up in the stored record.

alternatively we could use a query parameter: `/submissions/123?step=456`

the stepId itself could be any arbitrary value, and will differ for every form

lib/submissions.ts
```diff
@@ -0,0 +1,73 @@
...
+  formId: string,
+  socialCareId: number,
+  createdBy: string
+): Promise<Submission> => {
+  const { data } = await axios.post(
+    `${ENDPOINT_API}/submissions`,
+    {
+      formId,
+      socialCareId,
+      createdBy,
```
jhackett1, line 20: at the moment this is just called `workerEmail`, but we want to somehow distinguish between editors and creators

lib/submissions.ts
Miles-Alford: Correct me if I'm wrong but I think we usually send the date in the ISO8601 format

Miles-Alford:
```suggestion
      submittedAt: new Date().toISOString(),
```

lib/submissions.ts
JohnFarrellDev: I think at creation (at least for now) it is fine to just submit createdBy.
The backend stores who created the submissions and also createdBy is assigned to a list of workers assigned to the submission. 

By default the person who create the submission is also assigned to work it, however, there is nothing stopping our patch endpoint to let us assign multiple workers to a submission or remove a worker from a submission.

jhackett1 (approved):

PR #424, opened 2021-06-01
feat(SCT-382): add 'created by' metadata to the end of the warning note view table

chrishutchinson:
**What**  
Adds some extra context to the warning note view page by adding the email address for the user who created the note to the bottom of the summary table.

This is done by passing the new optioanl `additionalMetadata` prop to `<Summary />` which appends additional data to the end of the child `<SummaryList />` component.

**Why**  
[SCT-382](https://hackney.atlassian.net/browse/SCT-382)

**Anything else?**


chrishutchinson:
@Renny4Real is this still needed after your changes in the other tickets you were working on? I just wanted to confirm it wasn't picked up elsewhere?

Renny4Real:
> @Renny4Real is this still needed after your changes in the other tickets you were working on? I just wanted to confirm it wasn't picked up elsewhere?

@chrishutchinson, I think this is still needed. 🙂 
The work I was doing was to view reviewed/ended warning notes details and make sure the details we want were being shown under each `Warning Note Review Details` section.

I didn't touch the original warning note summary, which I think this PR does.

components/WarningNote/WarningNoteRecap/WarningNoteRecap.spec.tsx
```diff
@@ -2,25 +2,47 @@ import { render } from '@testing-library/react';
...
-    revalidate: jest.fn(),
-    isValidating: false,
-  }));
   const props = {
     person: mockedResident,
     warningNoteId: 1,
   };
 
+  beforeEach(() => {
+    jest.resetAllMocks();
```
Renny4Real, line 17: Is this kind of like an initial teardown before running the tests? 🤔

Renny4Real (approved):
Looks good 👍🏾

components/WarningNote/WarningNoteRecap/WarningNoteRecap.spec.tsx
chrishutchinson: Yeah you're absolutely right. This a force of habit on my behalf, muscle memory I've always just used `beforeEach` 😅 . 

It effectively says:
>before each test, reset the mocks back to their default state

You could also do it in an `afterEach` and it'd basically do the same thing

PR #428, opened 2021-06-02
clean up form layout a 'lil

jhackett1:
- make the sidebar sticky like it should be
- add a go back button to the step page as well as the task list


JohnFarrellDev (approved):
LGTM :+1:

PR #429, opened 2021-06-02
warn about unsaved changes, interim fix

jhackett1:
in the source codebase, warning the user about unsaved changes was linked with autosave functionality.

since we've gutted autosave for now, we're no longer warned about unsaved changes.

this fixes that by using formik's `dirty` feature. when we bring autosave back we will revisit this.


JohnFarrellDev (approved):
Makes sense :+1:

PR #430, opened 2021-06-02
Feature/real api routes for submissions

jhackett1:
final piece of work required to get the new flexible form builder working end-to-end.

it uses the functions written [here](https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/418) to:

- kick off a submission
- either fetch or finish the current submission for the task list page
- either fetch or update the current step's answers for a step page


JohnFarrellDev (approved):
Looks good

lib/submissions.ts
```diff
@@ -0,0 +1,73 @@
...
+  return data;
+};
+
+export const finishSubmission = async (
+  submissionId: string
+): Promise<Submission> => {
+  const { data } = await axios.patch(
+    `${ENDPOINT_API}/submissions/${submissionId}`,
+    {
+      submittedAt: new Date(),
```
JohnFarrellDev, line 66: When we finish a submission will the submittedAt ever be dynamic.

I'd actually be quite keen to remove this from the JSON body and allow the backend to get DateTime.Now.

JohnFarrellDev: Also should we provide the value for the worker who is doing the submission, do we want to store SubmittedBy in the backend?

jhackett1: it will always be the current time, yes

jhackett1: it's definitely important to distinguish between the creator and editors, but i'm not sure whether it matters much who hits the "finish" button. let's leave it as is for now and enhance it later if we need.

jhackett1:
leaving this un-merged for now, because once we merge it, it'll break the mock api responses that we need for social care colleagues to assure the form content

PR #431, opened 2021-06-02
Autosave

jhackett1:
this fixes https://hackney.atlassian.net/browse/SCT-613

unsaved changes and autosave are quite closely allied, so the fastest fix is simply to bring back the missing autosave context 

---

flexible forms now support autosave

it works by:

- wrapping each `StepForm` in an autosave context component, which keeps track of whether the form has unsaved changes
- adding an `<AutosaveTrigger/>` component inside the form, which uses formik's context hook to submit the form after a delay (set to 2 seconds here for prudence). this component is also in charge of warning the user if they try to navigate away while there are unsaved changes (superceding https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/429)
- an `<AutosaveIndicator/>` ui component rendered in the sidebar, to show the user what's going on

we've spoken about whether our apis are speedy enough to support autosave. here we've conservatively set the maximum number of api calls to one every ten seconds, which should be slow enough to prevent any performance impact while also providing some minimal usefulness. we should lower the delay to whatever our apis can handle (edit: they seem good to let's try 2 seconds)

## future enhancements

1. strictly speaking, changes should stay "unsaved" if form submission fails due to a server-side submission error


anabebs (approved):
seems to have fixed the issue.

PR #433, opened 2021-06-03
connect fe backend to api

anabebs:
This is a copy of #432  without merge conflicts.

### DIFF
--- a/lib/submissions.ts
+++ b/lib/submissions.ts
@@ -63,7 +63,7 @@ export const finishSubmission = async (
   const { data } = await axios.patch(
     `${ENDPOINT_API}/submissions/${submissionId}`,
     {
-      submittedAt: new Date(),
+      submittedAt: new Date().toISOString(),
     },
     {
       headers: headersWithKey,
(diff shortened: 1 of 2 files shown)


jhackett1 (approved):

lib/submissions.ts
```diff
@@ -0,0 +1,73 @@
...
+  return data;
+};
+
+export const finishSubmission = async (
+  submissionId: string
+): Promise<Submission> => {
+  const { data } = await axios.patch(
+    `${ENDPOINT_API}/submissions/${submissionId}`,
+    {
+      submittedAt: new Date(),
```
jhackett1, line 66: shouldn't this be `.toISOString()`?

anabebs: I'm not sure. A test was failing because of this last week. I think you said leave the file as is and alter the test but I couldn't figure out how to do that with MockDate. It now seems to be passing fine.

jhackett1: @JohnFarrellDev can say for sure. i believe we will need to submit an iso date string, so the test should represent that

JohnFarrellDev: I don't think the backend will care, backend just expects a json value that can be deserialized into a DateTime object.

However, we are trying to standardise around ISO-8601 (standard date format) so I'm all for calling `.toISOString()`

PR #434, opened 2021-06-03
SCT-385  Change permissions to allow for ASC access

mapleleafjack:
[Ticket link](https://hackney.atlassian.net/browse/SCT-385)

**What**  
Adding ASC permissions for viewing relationships.
DEV permissions are still evaluated for CFS sign off

### DIFF
--- a/lib/permissions.ts
+++ b/lib/permissions.ts
@@ -65,7 +65,7 @@ export const canManageCases = (
   return false;
 };
 
-export const devCanViewArea = (user: User) => {
+export const canUserViewWarningNotes = (user: User) => {
   if (user.hasDevPermissions) {
     return true;
   }
--- a/lib/permissions.ts
+++ b/lib/permissions.ts
@@ -76,7 +76,7 @@ export const canViewRelationships = (user: User, resident: Resident) => {
   if (user.hasDevPermissions) {
     return true;
   }
-  if (user.hasAdultPermissions && resident.contextFlag == 'A') {
+  if (user.hasAdultPermissions && resident.contextFlag === 'A') {
     return true;
   }
 
--- a/lib/permissions.ts
+++ b/lib/permissions.ts
@@ -65,7 +65,7 @@ export const canManageCases = (
   return false;
 };
 
-export const canViewArea = (user: User) => {
+export const devCanViewArea = (user: User) => {
   if (user.hasDevPermissions) {
     return true;
   }
--- a/lib/permissions.ts
+++ b/lib/permissions.ts
@@ -72,11 +72,14 @@ export const devCanViewArea = (user: User) => {
 
   return false;
 };
-export const canViewRelationships = (user: User, resident: Resident) => {
+export const canViewRelationships = (
+  user: User,
+  person: Pick<Resident, 'restricted' | 'contextFlag'>
+) => {
   if (user.hasDevPermissions) {
     return true;
   }
-  if (user.hasAdultPermissions && resident.contextFlag === 'A') {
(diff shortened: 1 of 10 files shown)


pages/people/[id]/index.tsx
```diff
@@ -34,7 +34,12 @@ const PersonPage = (): React.ReactElement => {
             {canViewArea(user) ? <WarningNotes id={personId} /> : <></>}
             <PersonDetails person={person} />
             <AllocatedWorkers person={person} />
-            {canViewArea(user) ? <Relationships id={personId} /> : <></>}
+            {user.hasDevPermissions ||
```
Comment (chrishutchinson) [inline pages/people/[id]/index.tsx:37]: I've been modelling these permission checks as functions in this file – https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/main/lib/permissions.ts

So in this case it'd be something like (⚠️ pseudo code!):

```ts
const canUserSeeRelationshipsForPerson = (user: User, person: Resident) => {
  if(user.hasDevPermissions) {
    return true;
  }

  if(user.hasAdultPermissions && person.contextFlag == 'A') {
    return true;
  }

  return false;
}
```

A few main things here:
1) This is a pure function, only operating on inputs, which is nice in terms of re-use
2) We can re-use these functions more easily
3) More explicit condition checks make it easier to read later to understand why access was or wasn't granted
4) You can unit test the permission check explicitly as a separate func.
5) We will eventually be moving some of these permissions (either to the backend, or likely to some more common front end place), as part of the permissioning work, so putting all the checks nicely in one place should make it a little easier to migrate 😄

pages/people/[id]/index.tsx
```diff
@@ -34,7 +34,12 @@ const PersonPage = (): React.ReactElement => {
...
+            (user.hasAdultPermissions && person.contextFlag == 'A') ? (
```
Comment (chrishutchinson) [inline pages/people/[id]/index.tsx:38]: Also nit pick but we should probably do `===` equality here, not `==`

pages/people/[id]/index.tsx
Comment (mapleleafjack) [reply pages/people/[id]/index.tsx:37]: Hi @chrishutchinson,

sounds good, will definitely use this structure for the permissions!
I'll move also the pre-existing ones in this file (the "canViewArea" function)

Comment (chrishutchinson) [reply pages/people/[id]/index.tsx:37]: Nice 👍 and yeah sounds good, `canViewArea` isn't well named either, so a good opportunity to tidy that up 😄

pages/people/[id]/index.tsx
Comment (mapleleafjack) [reply pages/people/[id]/index.tsx:38]: Done ;)

pages/people/[id]/index.tsx
Comment (mapleleafjack) [reply pages/people/[id]/index.tsx:37]: I've renamed `canViewArea` to `devCanViewArea` :)

mapleleafjack:
@chrishutchinson 
all permission checks (that I know of, will update them as well in case) should be following the correct structure now, together with tests :)

lib/permissions.ts
```diff
@@ -64,3 +64,24 @@ export const canManageCases = (
 
   return false;
 };
+
+export const devCanViewArea = (user: User) => {
```
chrishutchinson, line 68: I'd give this a more explicit name, something like: `canUserViewWarningNotes` (the format is basically "can user do this {thing}". In this permissions file `devCanViewArea` is fairly abstract 😅

mapleleafjack: @chrishutchinson 
it's now called canUserViewWarningNotes

chrishutchinson (approved):

PR #435, opened 2021-06-03
make it easier to get a resident by id

jhackett1:
see https://hackit-lbh.slack.com/archives/C021Q88C953/p1622725373088900


JohnFarrellDev (approved):
Makes sense, context flag is not required by backend for getting resident by resident ID.

PR #438, opened 2021-06-04
spike(SCT-398): feature flag exploration in Next.js

chrishutchinson:
**What**  
Please read the documentation I have put together for this spike here – https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/spike/feature-flags/adr/feature-toggles/spike.md

This is currently a draft PR for people to review the approach, rather than it being the final agreed API we would choose to use. This is mostly to agree a general strategy rather than nailing down the exact implementation, however I would appreciate input on the examples. Any other thoughts are also appreciated!

**Why**  
[SCT-398](https://hackney.atlassian.net/browse/SCT-398)

**Anything else?**


neilkidd (commented):
😍 The concept and approach. Do it now! 😆 
Some things for consideration:
1. For the actual flag enablement values, could we use simple bools? I think it simplifies cognitive load. EG: Instead of `active`, anything other than an explicit `true` always resolves to false.
2. Could we have a simple page or UI control that will show **all** feature flags to the user. For example, if I don't see a feature I'm expecting, I can just check the values via the UI. (This could also be a post implementation enhancement.)
3. Where would the underlying values be stored and managed? (SSM and IaC?)
4. The ADR need a status 😜 See: https://github.com/LBHackney-IT/social-care-architecture/blob/main/decisions/0002-use-iso-8601-format-for-dates.md

.eslintrc
```diff
@@ -10,7 +10,8 @@
     "@typescript-eslint",
     "react",
     "react-hooks",
-    "jest"
+    "jest",
+    "unicorn"
```
Renny4Real, line 14: What is unicorn?

.eslintrc
```diff
@@ -52,6 +61,7 @@
         "plugin:cypress/recommended"
       ],
       "rules": {
+        "react/prop-types": "off",
```
Renny4Real, line 64: What does this rule enforce?

adr/feature-toggles/spike.md
```diff
@@ -0,0 +1,148 @@
...
+> The reason we have to use a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) is that the feature flag object must be serialised so it can be passed between server and client, and we can't serialise a function. We _can_ however serialise a getter, because it consolidates down to it's return value, in our case, either `true` or `false`.
+
+Below is an example where we toggle a 'dashboard' feature based on two things (a Local Storage value or a environment variable). This feature is also set to 'expire' on June 30th 2021.
+
+```ts
+{
+  // FEATURE-FLAG-EXPIRES[2021-06-30]
+  name: 'dashboard',
+  get isActive() => {
+    if (getLocalStorageFlagValue('dashboard') === 'active') {
```
Renny4Real, line 60: Would a local storage flag value be something similar to values added to an `.env` file?

adr/feature-toggles/spike.md
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return false;
+  },
+}
+```
+
+### React (Hook)
+
+A hook – `useFeatureFlags()` – provides a function that allows an engineer to interrogate the status of a given feature flag at runtime.
```
Renny4Real, line 75: What is a hook?

adr/feature-toggles/spike.md
```diff
@@ -0,0 +1,148 @@
...
+1. Initially we can keep things simple, checking globally available data (such as environment variables or local / session storage)
+2. When the need arises, our functions can become more advanced, for example they might:
+   - Check a URL query parameter value
+   - Check a property of your session token (JWT)
+   - Make an API call to the database to assess whether or not you should see the feature
+   - Call an external / third-party service to determine if the feature should be shown
+
+Early on this means our toggles will broadly be changed based on code releases, or environment variable changes, but gives us room to grow within the proposed API to explore more advanced options as the need arises, without the expensive cost of refactoring our entire feature flag system.
+
+## What documentation would a new system require?
```
Renny4Real, line 140: Could it be a wiki page/pages? 🤔

pages/_app.tsx
```diff
@@ -29,6 +31,12 @@ const CustomApp = ({
   pageProps,
 }: ExtendedAppProps): JSX.Element | null => {
   const [user] = useState(pageProps.user);
+  const [featureFlags, setFeatureFlags] = useState(pageProps.featureFlags);
+
+  useEffect(() => {
```
Renny4Real, line 36: What does `useEffect()` allow us to do? e.g could we have run `setFeatureFlags(getFeatureFlags())` here and get the same result?

Renny4Real (commented):
Sounds interesting. 🎉 
It might help to have a visual demo of what needs to be done to add feature flags to a hypothetical feature as a guide.

chubberlisk (commented):
This is really, really cool! 💯 I particularly love the expiring feature flag comment and the way you've documented and explained this.

chrishutchinson:
> 😍 The concept and approach. Do it now! 😆
> Some things for consideration:
> 
> 1. For the actual flag enablement values, could we use simple bools? I think it simplifies cognitive load. EG: Instead of `active`, anything other than an explicit `true` always resolves to false.
> 2. Could we have a simple page or UI control that will show **all** feature flags to the user. For example, if I don't see a feature I'm expecting, I can just check the values via the UI. (This could also be a post implementation enhancement.)
> 3. Where would the underlying values be stored and managed? (SSM and IaC?)
> 4. The ADR need a status 😜 See: https://github.com/LBHackney-IT/social-care-architecture/blob/main/decisions/0002-use-iso-8601-format-for-dates.md

Thanks @neilkidd, answers below:

1. Yeah you’re right here – this was mostly just for the use of local storage, which I think might be more effective as a cookie (available on server + client), but either way I agree, the value should be a simple Boolean!
2. Yeah this is a really nice idea, the app could expose them for use in a frontend in the future for sure. +1 to this idea!
3. I think SSM should be sufficient to start with, although the implementation also allows us to trial other options on a per-flag basis. We could even consider calling SSM within the app at run time (with some sort of in-memory cache perhaps to reduce cost), meaning we don’t have to rebuild and redeploy to enable new features, but I suspect that can be something we build up to!
4. Ah nice! Thanks for that – I figured there’d be a pattern to follow. I’ll update this doc into a full ‘decision’ rather than just a proposal once I’ve updated it with the above.

.eslintrc
chrishutchinson: Unicorn is the ESLint extension which allows us to do the expiring comments. Link here: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md

.eslintrc
chrishutchinson: Ah this actually isn’t meant to be part of this PR, it won’t be there in the full implementation.

However, in answer to your question, at the moment we type our React components as follows:

```tsx
type Props = {
  someProp: string;
};

const Component = ({ someProp }: Props): React.Component => {
  return <div>Whatever this component does here {someProp}…</div>;
};
```

This is fine-ish, although technically isn’t 100% accurate. React exposes a type called `React.FC` (FC = Functional Component) which is a generic and, using the same example, can be used as follows:

```tsx
Const Component: React.FC<{
  someProp: string;
}> = {
  return <div>Whatever this component does here {someProp}…</div>;
};
```

These are effectively the same thing, but I prefer the terser `React.FC<>` type. It also automatically types the `children` prop for us, which the first example doesn’t.

This `”react/prop-types”` ESLint rule doesn’t allow us to do the second pattern when it’s enabled, so I turned it off while putting this prototype together, purely for my own convenience!

adr/feature-toggles/spike.md
chrishutchinson: Yeah the idea of using local storage is that if you want to quickly flip the value in your browser you could do. It was more of a demonstration of the capability than a suggestion we should actually use it. Depending on the feature flag you may find this useful, you may not. I do think we should build something like this using cookies though, so the value of the flag can be read on both the server and client, because it’s in a cookie!

adr/feature-toggles/spike.md
chrishutchinson: I probably won’t explain it as well as the official React docs, so I’d take a read of this: https://reactjs.org/docs/hooks-intro.html

In short, they’re anything you using in a component that starts with `use`, so things like:
* `useState`
* `useReducer`
* `useCallback`
* `useEffect`

adr/feature-toggles/spike.md
chrishutchinson: +1 I think that’s a good idea yeah!

pages/_app.tsx
chrishutchinson: It’s a hook (see docs link in a previous comment). Effectively it’s the React way of triggering behaviour based on some action. That could be rendering the component, a user interaction that triggers some state change, or any number of other things.

chrishutchinson:
This is going into first implementation this sprint (Managing Workloads 3). As part of that we'll close this branch out, and file it as a formal ADR alongside the proper implementation! Watch this space

chrishutchinson:
This has now been implemented over a series of releases (https://github.com/LBHackney-IT/lbh-social-care-frontend/releases/tag/0.3.0, https://github.com/LBHackney-IT/lbh-social-care-frontend/releases/tag/0.4.0, and https://github.com/LBHackney-IT/lbh-social-care-frontend/releases/tag/0.5.1).

Given that, I am now going to close this PR 👍

PR #439, opened 2021-06-07
Undo change to reviewNotes component in review form step

Renny4Real:
**What**  
We've made a change in the structure of the API response when we request warning note reviews, so this reverses the previous change made in the frontend.

**Why**  

This was initially changed because the `reviewNotes` did not match the property
in the API response when we make a GET request for warning note reviews. 
See changes made in this [PR](https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/391) for better context.

The API response was returning a `Notes` property for each warning note review object.

From discussing the current naming conventions for the warning note data model,
it was agreed that any details of a review should be saved as `review notes`
and `notes` should be reserved for the original warning note.

### DIFF
--- a/data/forms/warning-note-review.tsx
+++ b/data/forms/warning-note-review.tsx
@@ -21,7 +21,7 @@ const formSteps: FormStep[] = [
       },
       {
         component: 'TextArea',
-        name: 'notes',
+        name: 'reviewNotes',
         label: 'Details of review',
         hint: 'include details of disclosure to individual, any updates and why renewing or ending',
         rules: { required: true },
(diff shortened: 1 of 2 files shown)


chubberlisk (approved):
Thanks for jumping onto a call and explaining this to me! As discussed, a bit of eyebrow raiser that we didn't need to update any tests for this change but as we've got broken functionality in staging and a card (if I remember correctly) to look more widely at how tests are being done for this... 🚢

Renny4Real:
> Thanks for jumping onto a call and explaining this to me! As discussed, a bit of eyebrow raiser that we didn't need to update any tests for this change but as we've got broken functionality in staging and a card (if I remember correctly) to look more widely at how tests are being done for this... 🚢

@chubberlisk Created a [ticket](https://hackney.atlassian.net/browse/SCT-446) for this. 🎉

PR #442, opened 2021-06-08
SCT-386 Change permission to allow for CFS access

mapleleafjack:
**What**  
-Added a new check in the canViewRelationships function:  user.hasChildrenPermissions && person.contextFlag === 'C'
-Added tests

**Why**  
This lets users with children permission to view children relationship


chubberlisk (approved):
Approving now but we should hold off merging until comms go out I think.

chubberlisk:
GitHub bugged out so you get two approvals! 😂

PR #449, opened 2021-06-09
fix(SCT-509): safely handle residents with partial addresses

chrishutchinson:
**What**  
Some resident's have partial addresses (e.g. missing postcode). Our UI doesn't gracefully handle these situations, until now!

**Why**  
[SCT-509](https://hackney.atlassian.net/browse/SCT-509)

### DIFF
--- a/components/Search/results/ResidentsTable.tsx
+++ b/components/Search/results/ResidentsTable.tsx
@@ -35,9 +35,13 @@ const ResultEntry = (person: LegacyResident): React.ReactElement => {
         {dateOfBirth && new Date(dateOfBirth).toLocaleDateString('en-GB')}
       </td>
       <td className="govuk-table__cell">
-        {address?.address && truncate(address.address, 4)}
+        {address?.address && truncate(address.address || '', 4)}
       </td>
-      <td className="govuk-table__cell">{address?.postcode.toUpperCase()}</td>
+      <td className="govuk-table__cell">
+        <span className={styles.uppercase}>
+          {(address && address.postcode) || ''}
+        </span>
+      </td>
       <td className="govuk-table__cell govuk-table__cell--numeric">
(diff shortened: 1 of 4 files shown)


components/Search/results/ResidentsTable.tsx
```diff
@@ -35,9 +35,13 @@ const ResultEntry = (person: LegacyResident): React.ReactElement => {
         {dateOfBirth && new Date(dateOfBirth).toLocaleDateString('en-GB')}
       </td>
       <td className="govuk-table__cell">
-        {address?.address && truncate(address.address, 4)}
+        {address?.address && truncate(address.address || '', 4)}
```
Renny4Real, line 38: Is `truncate` similar to a substring method and we want it to cut it down to four letters?

Renny4Real (commented):
Looks good 🎉 

Would it be possible to have screenshots of before/after UI changes?

I might have been looking in the wrong place (list of results when you search for a person) but I couldn't see any visual differences when I swapped between main and this branch.

chrishutchinson:
> Looks good 🎉
> 
> Would it be possible to have screenshots of before/after UI changes?
> 
> I might have been looking in the wrong place (list of results when you search for a person) but I couldn't see any visual differences when I swapped between main and this branch.

@Renny4Real good question – no visual changes here, just that the app now doesn't throw an error when this data is rendered (which was causing unintended log outs).

components/Search/results/ResidentsTable.tsx
chrishutchinson: Looks like it cuts down to a certain number of words, in this case 4. I haven't actually changed this, just made it work with a null `address.address` value

Renny4Real (approved):
👍🏾

PR #453, opened 2021-06-10
Fix(SCT-507): Update ASC warning note review form to include a disclosure question

Renny4Real:
**What**  

The form for submitting an adult warning note review was not displaying a radio option component that would ask the user if they have discussed the warning note review with the individual.

## Adult Warning Note Review Form
_**Before**_
Screenshot 2021-06-10 at 4 47 51 pm

_**After**_
Screenshot 2021-06-10 at 4 48 06 pm

**Why**  
The changes made include separating the `formSteps` used for generating the warning note review form into two:
- reviewFormStepsAdult
- reviewFormStepsChild

Previously, we used the `conditionalRender` parameter to display the `disclosure question` component depending on whether the person was an adult or child. However, this `conditionalRender` required the person object to be passed into the relevant formsStep component, and it did not seem to be picking up the person object. 
(Attempted to pass the person object in as part of the `formSteps` but couldn't get it to work 😅 )

Since this component was exclusive to Adult warning note reviews decided to separate the aformStepsa as detailed above and use the relevant formStep where required.

**Other related changes**
- Convert the value of `disclosedWithIndividual` from boolean to string answer (i.e. Yes/No) to display on the warning note review details page
- Convert `disclosedWithIndividual` form response (i.e. Yes/No) to the respective boolean when constructing the `PATCH` request.

### DIFF
--- a/components/WarningNote/WarningNoteRecap/WarningNoteRecap.tsx
+++ b/components/WarningNote/WarningNoteRecap/WarningNoteRecap.tsx
@@ -5,7 +5,10 @@ import { useWarningNote } from 'utils/api/warningNotes';
 import { formStepsAdult, formStepsChild } from 'data/forms/warning-note';
 import { Resident } from 'types';
 import SummaryList from '../../Summary/SummaryList';
-import reviewFormSteps from 'data/forms/warning-note-review';
+import {
+  reviewFormStepsAdult,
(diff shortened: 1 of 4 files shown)


chrishutchinson (approved):
Looks great, nice work!

PR #455, opened 2021-06-11
Feature/footer

jhackett1:
add a footer to the app, with (for now) links to the public roadmap and the give feedback form.


anabebs (commented):
Why is the Give feedback link blue while the Roadmap one is white?  Colour contrast doesn't seem great against the black. Font size seems tiny too.

jhackett1:
@anabebs chrome-specific bug i've now fixed. should be good to go

anabebs (approved):

PR #457, opened 2021-06-11
Improve add person flow

jhackett1:
this builds on https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/444 by adding an extra "speedbump" to the add new person flow, which warns the user about potential duplicates before they hit the confirm button.

we hope this will result in fewer duplicate people being created.

it does this by:
- adding a new `DuplicateWarningPanel` component which is fed the data that the user typed about the new person they'd like to add
- the component is rendered on the "review details" step of the add new person flow
- the component searches for potential matches with the same surname and first name as the new one, if it finds none, it renders nothing. otherwise, the user sees a prompt with a table of possible duplicates

we want to improve on this further by making enhancements to the back-end resident search algorithm

it also prefills the first name and last name on the add person form from what the user originally typed into their search, to save a little bit of time

more info: https://hackit-lbh.slack.com/archives/C021P4E5C59/p1623344339143500

### DIFF
--- a/components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
+++ b/components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
@@ -28,8 +28,6 @@ const DuplicateWarningPanel = ({
       .then(({ data }) => setMatchingResidents(data.residents));
   }, [newResident]);
 
-  console.log(matchingResidents);
-
   if (matchingResidents?.length > 0)
     return (
       <div className={s.mutedPanel}>
--- a/components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
+++ b/components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
@@ -1,9 +1,8 @@
-import { useState, useEffect } from 'react';
 import cx from 'classnames';
 import s from './DuplicateWarningPanel.module.scss';
-import { LegacyResident } from 'types';
-import axios from 'axios';
 import ResidentsTable from 'components/Search/results/ResidentsTable';
+import { useResidents } from 'utils/api/residents';
+import { LegacyResident } from 'types';
 
 interface Props {
   newResident: Record<string, unknown>;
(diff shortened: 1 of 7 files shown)


anabebs (approved):
Looks good

components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const DuplicateWarningPanel = ({
+  newResident,
+}: Props): React.ReactElement | null => {
+  const [matchingResidents, setMatchingResidents] = useState<LegacyResident[]>(
+    []
+  );
+
+  useEffect(() => {
+    axios
```
chrishutchinson, line 20: I think this should use the `useResidents()` hook instead of re-implementing the axios / networking behaviour inside this component?

components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
```diff
@@ -0,0 +1,61 @@
...
+        headers: { 'x-api-key': process.env.AWS_KEY },
+        params: {
+          first_name: newResident.firstName,
+          last_name: newResident.lastName,
+        },
+      })
+      .then(({ data }) => setMatchingResidents(data.residents));
+  }, [newResident]);
+
+  console.log(matchingResidents);
```
chrishutchinson, line 31: Stray console.log

components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  if (matchingResidents?.length > 0)
```
chrishutchinson, line 33: A nit pick, which is mostly personal preference, but I do think it makes for more logical component flow. I tend to follow the reverse pattern for these early returns. So something like this:

```tsx
if(!matchingResidents || matchingResidents.length === 0) {
  return null;
}

return <div ... /> // And the rest of your UI here...
```

I find it easier to reason about, and to add extra error cases in the future, but as I said, mostly personal preference.

components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
```diff
@@ -0,0 +1,61 @@
...
+  useEffect(() => {
+    axios
+      .get(`/api/residents`, {
+        headers: { 'x-api-key': process.env.AWS_KEY },
+        params: {
+          first_name: newResident.firstName,
+          last_name: newResident.lastName,
+        },
+      })
+      .then(({ data }) => setMatchingResidents(data.residents));
```
chrishutchinson, line 28: What happens if this errors? We aren't handling the `.catch` right now (also a good argument for using `useResidents()`).

jhackett1: since it's not the main purpose of the page, failing silently is probably the correct (and current) behaviour. it would be more confusing to render an error message.

components/DuplicateWarningPanel/DuplicateWarningPanel.tsx
jhackett1: i avoided it because it looked excessive for what we needed (infinite loading, etc) but maybe it's fine

chrishutchinson: Sure – I think better to reuse or expand the shared capability given we have it, rather than reimplement

jhackett1: done

PR #458, opened 2021-06-11
fix(SCT-505): fix placement of the RESTRICTED label with new search UI

chrishutchinson:
**What**  
Since the new search UI went live with the dashboard, the `RESTRICTED` label has looked a little off! This removes address, and expands the header to cover the label.

### Before
Screenshot 2021-06-11 at 15 51 14

### After
Screenshot 2021-06-11 at 15 50 49

**Why**  
[SCT-505](https://hackney.atlassian.net/browse/SCT-505)


Renny4Real (approved):
🎉

PR #462, opened 2021-06-14
Fix/open residents in new tab when adding

jhackett1:
duplicates flagged for review when creating a new person will now open in a new tab, as a quick fix to prevent disrupting back button behaviour.

fixing the back button behaviour itself would require re-writing parts of the formwizard, which we're not keen to do because we're already about to replace it.

https://hackit-lbh.slack.com/archives/G01CYA7LM6V/p1623679740379300

### DIFF
--- a/components/Search/results/ResidentsTable.tsx
+++ b/components/Search/results/ResidentsTable.tsx
@@ -4,9 +4,14 @@ import { LegacyResident, User } from 'types';
 import { canManageCases } from '../../../lib/permissions';
 import { useAuth } from '../../UserContext/UserContext';
 import styles from './ResidentsTable.module.scss';
-import { truncate } from 'lib/utils';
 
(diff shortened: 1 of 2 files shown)


components/Search/results/ResidentsTable.tsx
```diff
@@ -25,11 +30,21 @@ const ResultEntry = (person: LegacyResident): React.ReactElement => {
     >
       <td className="govuk-table__cell">{mosaicId}</td>
       <td className="govuk-table__cell">
-        <Link href={`/people/${mosaicId}`}>
-          <a className="govuk-link govuk-custom-text-color">
+        {newTab ? (
+          <a
+            href={`/people/${mosaicId}`}
+            target="blank"
```
JohnFarrellDev, line 36: should probably be `"_blank"`

https://css-tricks.com/targetblank/

JohnFarrellDev (approved):
:+1:

PR #463, opened 2021-06-14
Bug(SCT-593): Fix duplicate team display

anabebs:
**What**  
Change worktray Team display to display multiple teams with commas between
**Why**  
Multiple teams displaying in a long string

**Anything else?**
There is an outstanding backend ticket which looks at why teams are not being removed when a worker is updated with a new team.


JohnFarrellDev:
I think you need to do an `npm i`, this will recreate the `package-lock.json` file.

JohnFarrellDev (approved):
:+1:

PR #464, opened 2021-06-14
feat(SCT-552, SCT-598): Create warning note automation tests

chrishutchinson:
**What**  
In order to boost our confidence with the warning notes form, we want to expand our automation tests. This is the first of three PRs, that adds automation tests for the 'create' form.

Future PRs will cover:
* Review warning note ([SCT-598](https://hackney.atlassian.net/browse/SCT-599))
* End warning note ([SCT-598](https://hackney.atlassian.net/browse/SCT-600))

**Why**  
[SCT-552](https://hackney.atlassian.net/browse/SCT-552), [SCT-598](https://hackney.atlassian.net/browse/SCT-598)


neilkidd (approved):
🥇

components/pages/warning-notes/AddWarningNoteForm/AddWarningNoteForm.spec.tsx
```diff
@@ -0,0 +1,506 @@
...
+    it('should not push to the summary route if there are errors on the form', async () => {
+      fireEvent.submit(screen.getByRole('form'));
+
+      await waitFor(() => {
+        expect(Router.push).not.toHaveBeenCalled();
+      });
+    });
+  });
+
+  describe('for a childrens resident', () => {
```
Renny4Real, line 487: 🎉 

Wondering if we could split up the scenarios even more? 🤔 

Most of the error messages would occur on both forms. Wondering if that could be a general scenario to consider, i.e. we expect it on the form regardless of the resident, e.g. describe('when filling the form for either adults or children'), or something similar and all the form errors for the components shared between an adult and child form could go in there.

Then the form elements and associated errors specific to the adult form could go under another scenario for `describe('for an adult resident only'` or something similar.

And then a scenario for testing things on the children side, e.g. `describe('for a childrens resident only'`  or something similar.

My thoughts are around someone new/unfamiliar reading this and seeing most of the form error validations under a describe block for an adult resident and possibly misunderstand where the difference between an adult form and child form lie. E.g. Could this test file be read as the errors occur exclusively on the adults form 🤔

components/pages/warning-notes/AddWarningNoteForm/AddWarningNoteForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import { formStepsAdult, formStepsChild } from 'data/forms/warning-note';
+
+import type { Resident, User, WarningNote } from 'types';
+
+interface FormWarningNote extends Omit<WarningNote, 'disclosedWithIndividual'> {
+  disclosedWithIndividual: 'Yes' | 'No';
+}
+
+export const AddWarningNoteForm: React.FC<{ personId: number }> = ({
```
Renny4Real, line 17: 🎉

jest.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   collectCoverageFrom: [
     'components/**/*.{js,jsx,ts,tsx}',
     'utils/**/*.{js,jsx,ts,tsx}',
+    'data/**/*.{js,jsx,ts,tsx}',
```
Renny4Real, line 5: Does this add the coverage for the tests?

pages/people/[id]/warning-notes/add/[[...stepId]].tsx
```diff
@@ -1,65 +1,20 @@
...
+        Add Warning Note
+      </h1>
+
+      <AddWarningNoteForm personId={personId} />
     </>
   );
 };
 
-export default CaseNotesRecording;
+export default AddWarningNotePage;
```
Comment (Renny4Real) [inline pages/people/[id]/warning-notes/add/[[...stepId]].tsx:20]: Would we want/need to (at some point) not use export default here and instead use export const, or can we get away with it here?

Renny4Real (approved):
🎉

components/pages/warning-notes/AddWarningNoteForm/AddWarningNoteForm.spec.tsx
```diff
@@ -0,0 +1,506 @@
...
+  );
+  fireEvent.change(
+    screen.getByLabelText('Month', {
+      selector: `[name='${fieldName}-month']`,
+    }),
+    {
+      target: {
+        value:
+          date instanceof Date
+            ? String(date.getMonth() + 1).padStart(2, '0')
```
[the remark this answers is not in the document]
chrishutchinson: That's correct yeah – docs for that here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

jest.config.js
chrishutchinson: Yep, that's right! It means the coverage report that's generated (adding `--coverage` to `yarn test`) will now include line coverage details for files in the `data/` directory

pages/people/[id]/warning-notes/add/[[...stepId]].tsx
Comment (chrishutchinson) [reply pages/people/[id]/warning-notes/add/[[...stepId]].tsx:20]: Ah, so this is a quirk of Next.js. For files that make up pages (aka in the `pages/` directory), they have to be exported as default. However, these files are never imported by us, they're used by Next.js internally, so there's no practical risk of someone importing it with the wrong name (unless someone is doing something weird by importing a page inside another component 😱!)

components/pages/warning-notes/AddWarningNoteForm/AddWarningNoteForm.spec.tsx
chrishutchinson: Really nice suggestion, I've done that in https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/464/commits/2bce97d324558a9835052ef843fdd8481ff6cca1

The test output now looks like this:

Screenshot 2021-06-16 at 10 56 36

components/pages/warning-notes/AddWarningNoteForm/AddWarningNoteForm.spec.tsx
```diff
@@ -0,0 +1,506 @@
...
+
+      await waitFor(() => {
+        screen.getByText(
+          'Enter justification for non-disclosure of warning note',
+          {
+            selector: '.govuk-error-message',
+          }
+        );
+      });
+    });
```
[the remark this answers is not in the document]
chrishutchinson: Done in https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/464/commits/2bce97d324558a9835052ef843fdd8481ff6cca1

Renny4Real (approved):
🎉

PR #467, opened 2021-06-15
feat(SCT-377): enable Warning Notes for CFS

chrishutchinson:
**What**  
We have CFS sign off for Warning Notes, so this change puts them live for CFS users/residents, but not for adults. They are still active on all residents for developer users.

**Why**  
[SCT-377](https://hackney.atlassian.net/browse/SCT-377)


components/AddForm/AddForm.tsx
```diff
@@ -13,33 +13,27 @@ const AddForm = ({ person }: { person: Resident }): React.ReactElement => {
...
-    ageContext === 'A'
-      ? [
-          {
-            text: 'Case Note Recording',
-            value: `/people/${person.id}/records/case-notes-recording`,
-          },
-          // {
-          //   text: 'Conversation 3',
-          //   value: `/people/${person.id}/records/conversation-3`,
-          // },
```
Renny4Real, line 26: Do we know why this was commented out? Are we still working on `Conversation 3`? 
Wondering if we should keep the commented out code in case someone goes looking for it? 😁

Renny4Real (approved):
🎉 🚀

components/AddForm/AddForm.tsx
chrishutchinson: I'm happy to get rid of commented code, it's always available in version control should someone need to go back to it!

PR #468, opened 2021-06-15
it works and we love that it works

jhackett1:
final tweaks needed to make the new form builder and submissions engine demoable end to end.

go to `/submissions/new` to kick off a submission

go to `/people/:id/submission/:submissionId` to review a finished submission

this _does not_ actually start redirecting users to the new form builder. so it's totally safe to launch. that will come in a few days.

### DIFF
--- a/data/flexibleForms/allForms.ts
+++ b/data/flexibleForms/allForms.ts
@@ -1596,15 +1596,14 @@ const forms: Form[] = [
     name: 'Another example',
     steps: [
       {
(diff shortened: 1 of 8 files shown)


chrishutchinson:
Would it be possible for us to try and stick to more meaningful PR titles? If possible, following the `fix`/`feat`/`chore` syntax too! 🙇

anabebs (approved):

PR #469, opened 2021-06-16
feat: improve testing for worker allocation behaviour

chrishutchinson:
**What**  
It has been raised that we were having some issues with allocating workers. As part of my investigation it seemed we didn't have many (if any) automation tests around the behaviour, especially the permissions. 

This PR:

* Refactors the logic used for rendering the allocation form to use the same permission function as rendering the "Allocate worker" button
* Adds Cypress tests for the button and add allocation form
* Adds unit tests for the updated permission method

**Why**  
To give us confidence the system works, which it does, this is just a refactor!

### DIFF
--- a/pages/people/[id]/allocations/add.tsx
+++ b/pages/people/[id]/allocations/add.tsx
@@ -5,15 +5,31 @@ import AddAllocatedWorker from 'components/AllocatedWorkers/AddAllocatedWorker/A
 import PersonView from 'components/PersonView/PersonView';
 import { useAuth } from 'components/UserContext/UserContext';
 import { isBrowser } from 'utils/ssr';
+import { canUserAllocateWorkerToPerson } from '../../../../lib/permissions';
+import { User } from '../../../../types';
+import { useResident } from '../../../../utils/api/residents';
+import ErrorMessage from '../../../../components/ErrorMessage/ErrorMessage';
+import Spinner from '../../../../components/Spinner/Spinner';
 
 const AddNewAllocationPage = (): React.ReactElement => {
   const { query, replace } = useRouter();
   const personId = Number(query.id as string);
-  const { user } = useAuth();
-  if (isBrowser() && !user?.hasAllocationsPermissions) {
+  const { user } = useAuth() as { user: User };
+  const { data: person, error } = useResident(personId);
+
+  if (error) {
+    return <ErrorMessage />;
+  }
+
+  if (!person) {
+    return <Spinner />;
+  }
+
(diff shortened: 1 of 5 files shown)


pages/people/[id]/allocations/add.tsx
```diff
@@ -5,15 +5,31 @@ import AddAllocatedWorker from 'components/AllocatedWorkers/AddAllocatedWorker/A
...
-  if (isBrowser() && !user?.hasAllocationsPermissions) {
+  const { user } = useAuth() as { user: User };
+  const { data: person, error } = useResident(personId);
+
+  if (error) {
+    return <ErrorMessage />;
+  }
+
+  if (!person) {
+    return <Spinner />;
```
Comment (Renny4Real) [inline pages/people/[id]/allocations/add.tsx:25]: What is the Spinner component? Is it the loading circle?

pages/people/[id]/allocations/add.tsx
```diff
@@ -5,15 +5,31 @@ import AddAllocatedWorker from 'components/AllocatedWorkers/AddAllocatedWorker/A
...
+  }
+
+  if (isBrowser() && !canUserAllocateWorkerToPerson(user, person)) {
```
Comment (Renny4Real) [inline pages/people/[id]/allocations/add.tsx:28]: What does `isBrowser()` check? Is it checking that the page is being loaded on a browser?

pages/people/[id]/allocations/add.tsx
```diff
@@ -5,15 +5,31 @@ import AddAllocatedWorker from 'components/AllocatedWorkers/AddAllocatedWorker/A
...
     replace(`/people/${personId}`);
```
Comment (Renny4Real) [inline pages/people/[id]/allocations/add.tsx:29]: Does `replace` cause the redirect if the person does not have permissions?

Renny4Real (approved):
🎉

pages/people/[id]/allocations/add.tsx
Comment (chrishutchinson) [reply pages/people/[id]/allocations/add.tsx:25]: Yeah that's right, this one:

Screenshot 2021-06-16 at 15 48 50

pages/people/[id]/allocations/add.tsx
Comment (chrishutchinson) [reply pages/people/[id]/allocations/add.tsx:28]: Yeah, you can't do this type of redirect from a server-side render (technically it is possible to do a redirect from server-side render but we don't use that pattern at the moment). This says: "if I'm in the browser client environment and I can't allocate workers to this person"

pages/people/[id]/allocations/add.tsx
Comment (chrishutchinson) [reply pages/people/[id]/allocations/add.tsx:29]: Yeah, that's correct. `replace()` is doing a redirect by changing the URL

PR #472, opened 2021-06-16
SCT-511 - App redirects to login when "Clear search" is used

mapleleafjack:
**What**  
Added  check for data != 1 (useSWRinfinite)

**Why**  
When searching for residents, and using the "Clear search" mechanism, the app redirects to the login page due to an error

This is not the definitive solution, but it will prevents the app from breaking when the user search twice

### DIFF
--- a/components/Search/Search.jsx
+++ b/components/Search/Search.jsx
@@ -67,10 +67,11 @@ const Search = ({
     query,
     hasQuery || showOnlyMyResults
   );
-  const results = data && {
-    records: data?.reduce((acc, d) => [...acc, ...getRecords(d)], []),
-    nextCursor: data[data.length - 1].nextCursor,
-  };
+  const results = data &&
+    data != 1 && {
+      records: data?.reduce((acc, d) => [...acc, ...getRecords(d)], []),
+      nextCursor: data[data.length - 1].nextCursor,
+    };
   const onFormSubmit = useCallback(
     (formData) => {
       const qs = formData
@@ -142,7 +143,7 @@ const Search = ({
           )
         )}
       </div>
-      {error && <ErrorMessage />}
+      {error && data != 1 && <ErrorMessage />}
 
       {type === 'people' && results && !results?.nextCursor && (
         <>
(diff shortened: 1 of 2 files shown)


chubberlisk (commented):
Looks like we have [Cypress tests that cover searching](https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/c0fef49940f698c23da93e1ef417e97cb85130aa/cypress/integration/search_for_person.ts#L3), could we add tests to cover this?

components/Search/Search.jsx
```diff
@@ -142,7 +143,7 @@ const Search = ({
           )
         )}
       </div>
-      {error && <ErrorMessage />}
+      {error && data != 1 && <ErrorMessage />}
```
chrishutchinson, line 146: What happens if for some reason there is an error (say a HTTP error) and the value of `data` is `1`? Is that scenario possible?

components/Search/Search.jsx
```diff
@@ -67,10 +67,11 @@ const Search = ({
     query,
     hasQuery || showOnlyMyResults
   );
-  const results = data && {
-    records: data?.reduce((acc, d) => [...acc, ...getRecords(d)], []),
-    nextCursor: data[data.length - 1].nextCursor,
-  };
+  const results = data &&
+    data != 1 && {
```
chrishutchinson, line 71: Probably best to use strict (in)equality: `data !== 1`. Same in your change further down too

components/Search/Search.jsx
mapleleafjack: It should be not possible, i've  tested it with a "wrong" connection string and it returns an empty array - triggering the error correctly

components/Search/Search.jsx
mapleleafjack: Done ;)

jhackett1:
what would the definitive fix be?

jhackett1 (approved):

mapleleafjack:
> what would the definitive fix be?

Hi @jhackett1,

The definitie fix would be the one in SCT-610

PR #474, opened 2021-06-17
SCT-614: Added HotJar code in Head tag

mapleleafjack:
[Link to ticket](https://hackney.atlassian.net/browse/SCT-614)

**What**  
Added HotJar code in Head tag (available in every page in production-only environment)

**Why**  
This enables HotJar analytics for the application

**Anything else?**
Code has been added in the same fashion as googletagmanager and google analytics

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -24,9 +24,10 @@ export default class AppDocument extends Document {
           `,
             }}
           />
-          <script
-            dangerouslySetInnerHTML={{
-              __html: `// Hotjar Tracking Code for https://social-care-service.hackney.gov.uk/
+          {process.env.NODE_ENV === 'production' ? (
+            <script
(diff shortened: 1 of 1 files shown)


pages/_document.tsx
```diff
@@ -24,6 +24,19 @@ export default class AppDocument extends Document {
           `,
             }}
           />
+          <script
+            dangerouslySetInnerHTML={{
+              __html: `// Hotjar Tracking Code for https://social-care-service.hackney.gov.uk/
```
chubberlisk, line 29: Do you think we want to just have this for production? Or may we want to test on staging? Just wondering if we should hardcode this or make it dependent on environment.

mapleleafjack: Good catch! Just confirmed this with Joy, it needs to go for Production only

chubberlisk (approved):

PR #476, opened 2021-06-18
refactor types to display addresses array and add tests.

anabebs:
**What**  
After a change to the back end to send the addresses array, the Person Widget was updated to display the correct postcode and address info

**Why**  
Address and Postcode were not displaying in Person Widget 

**Anything else?**
Adapted types file and resident Factory to contain addresses array.
Added a new test for handling addresses.

### DIFF
--- a/components/PersonWidget/PersonWidget.tsx
+++ b/components/PersonWidget/PersonWidget.tsx
@@ -19,6 +19,8 @@ interface Props {
 const PersonWidget = ({ person }: Props): React.ReactElement => {
   const dateOfBirth = prettyDate(person?.dateOfBirth ?? '');
 
+  const address = person?.addresses?.[0];
+
   if (person) {
     return (
       <aside className={s.aside}>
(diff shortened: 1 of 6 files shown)


components/PersonWidget/PersonWidget.tsx
```diff
@@ -19,6 +19,8 @@ interface Props {
 const PersonWidget = ({ person }: Props): React.ReactElement => {
   const dateOfBirth = prettyDate(person?.dateOfBirth ?? '');
 
+  const address = person?.addresses?.[0];
```
jhackett1, line 22: probably this should be `displayAddress` or similar just for the sake of readability

jhackett1 (approved):

components/PersonWidget/PersonWidget.tsx
anabebs: ok

JohnFarrellDev (approved):
:rocket:

PR #477, opened 2021-06-18
SCT-590 - Moved API endpoint to V1

mapleleafjack:
**What**  
Changed the Relationship call in the frontend to use the v1 endpoint

**Why**  
To keep functionality working while we work on the relationship


chubberlisk (approved):
🚀

PR #479, opened 2021-06-18
redirect to finished submission on completion

anabebs:
Co-authored-by: Ana Bebic <anabebs@users.noreply.github.com>

**What**  
create redirect on the answers page of a submission so that a user is not able to edit already submitted forms.


jhackett1 (approved):
looks grand once we get the e2e tests passing again

PR #480, opened 2021-06-18
Fix/refactor step form continue function

jhackett1:
improves the handling of the "save and continue" button at the bottom of a step form - using react state and effects to minimise the possibility of timing bugs and race conditions.

### DIFF
--- a/components/FlexibleForms/StepForm.spec.tsx
+++ b/components/FlexibleForms/StepForm.spec.tsx
@@ -81,29 +81,28 @@ describe('StepForm', () => {
     fireEvent.click(screen.getByText('Save and continue'));
 
     await waitFor(() => {
-      expect(mockPush).toBeCalledTimes(1);
+      expect(mockPush).toBeCalled();
       expect(mockPush).toBeCalledWith(`/submissions/foo`);
     });
   });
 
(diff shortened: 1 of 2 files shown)


components/FlexibleForms/StepForm.spec.tsx
```diff
@@ -86,23 +86,24 @@ describe('StepForm', () => {
...
-  it("also triggers the finish event if it's the only step", async () => {
-    render(
-      <StepForm
-        fields={mockFields}
-        onSubmit={() => true}
-        onFinish={mockFinish}
-        singleStep={true}
-      />
-    );
+  // TODO: bring this test back when we bring this functionality back
```
jhackett1, line 89: next sprint we're gonna need this functionality, so it's commented out rather than deleted

chrishutchinson: A thought – using `it.skip("...` might be better, as it gives us visibility on the fact it's skipped in our test output. A good reminder to go back and restore it later!

anabebs (approved):

components/FlexibleForms/StepForm.spec.tsx
jhackett1: sounds sensible to me!

PR #481, opened 2021-06-18
add address display to form answers page

anabebs:
**What**  
small PR to update the person widget so that it uses both address and addresses when required.
This will fix the problem of the address not displaying on the form answers page.

### DIFF
--- a/components/PersonWidget/PersonWidget.tsx
+++ b/components/PersonWidget/PersonWidget.tsx
@@ -23,7 +23,6 @@ const PersonWidget = ({ person }: Props): React.ReactElement => {
   const displayAddresses = person?.addresses?.[0];
 
   if (person) {
-    console.log(displayPostcode);
     return (
       <aside className={s.aside}>
         <h2 className={`lbh-heading-h3 ${s.title}`}>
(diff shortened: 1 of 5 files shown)


components/PersonWidget/PersonWidget.tsx
```diff
@@ -18,10 +18,12 @@ interface Props {
...
 const PersonWidget = ({ person }: Props): React.ReactElement => {
   const dateOfBirth = prettyDate(person?.dateOfBirth ?? '');
-
-  const displayAddress = person?.addresses?.[0];
+  const displayAddress = person?.address?.address;
+  const displayPostcode = person?.address?.postcode;
+  const displayAddresses = person?.addresses?.[0];
 
   if (person) {
+    console.log(displayPostcode);
```
jhackett1, line 26: errant console.log

components/PersonWidget/PersonWidget.tsx
```diff
@@ -30,14 +32,11 @@ const PersonWidget = ({ person }: Props): React.ReactElement => {
...
-
-        {displayAddress && (
-          <p className={`lbh-body-s ${s.paragraph}`}>
-            {displayAddress.addressLines}
-            <br />
-            {displayAddress.postCode}
-          </p>
-        )}
+        <p className={`lbh-body-s ${s.paragraph}`}>
+          {displayAddress || displayAddresses?.addressLines}
```
jhackett1, line 36: is it ever possible to have no address on a person at all?

jhackett1 (approved):
looks fine apart from minor gripes

components/PersonWidget/PersonWidget.tsx
anabebs: maybe but i thought all the question marks in the consts would deal with that? Do they not?

jhackett1: you'd still end up with an empty div if there is no address to show - if it's not possible for a person to have no address (if it's always a required field) then that's fine

PR #482, opened 2021-06-21
clean up form config for launch

jhackett1:
just some basic tweaks and removing wip forms so that we're ready for launch


data/flexibleForms/index.ts
```diff
@@ -1,5 +1,7 @@
-import allForms from './allForms';
 import review3c from './review3C';
-import faceOverview from './faceOverview';
+// import faceOverview from './faceOverview';
 
-export default [...allForms, review3c, faceOverview];
+export default [
+  review3c,
+  // faceOverview
```
jhackett1, line 6: we'll be adding this back in a few days

anabebs (approved):

PR #483, opened 2021-06-21
feat(SCT-636): introduce an ErrorBoundary component for handling generic errors

chrishutchinson:
**What**  
When our app errors and we don't handle them (e.g. accessing a property of an undefined object), it currently redirects the user to the login page. This isn't ideal behaviour, and often makes debugging these sort of issues more difficult.

This PR introduces an error boundary (docs: https://reactjs.org/docs/error-boundaries.html) component that wraps the app, and renders a fairly generic error message (see below).

Just need to confirm:
* [x] UI
* [x] Wording
* [x] Support link – replaced with feedback form link

**Why**
[SCT-636](https://hackney.atlassian.net/browse/SCT-636)

Screenshot 2021-06-21 at 13 07 26


components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,60 @@
...
+          tabIndex={-1}
+          data-module="govuk-error-summary"
+        >
+          <h2 className="govuk-error-summary__title" id="error-summary-title">
+            A system error has occurred
+          </h2>
+          <div className="govuk-error-summary__body">
+            <p>
+              Try your request again. If the error persists, please contact{' '}
+              <a href="#">Support</a> sharing the current URL and message below:
```
jhackett1, line 44: sentence case

jhackett1 (approved):

neilkidd (approved):
An excellent incremental improvement :shipit: !

components/ErrorBoundary/ErrorBoundary.tsx
chrishutchinson: 👍  replaced with a link to the feedback form, and converted to sentence case

PR #485, opened 2021-06-21
feat(SCT-495): introduce `<ConditionalFeature />` for feature flagging

neilkidd:
**What**
This is the first implementation of feature flags, demonstrated via an "implementation-proof" flag which allows us to verify they are operating as expected.

This change contains:
* Basic underlying data structure to support feature flags (see type `FeatureSet`).
* A React context provider (`<FeatureFlagProvider`) to store the feature flags, and make them available to the React application
* A helper component (`<ConditionalFeature />`) which allows UI to be conditionally rendered based on the active state of the feature flag
* Documentation on how to use all of the above

Coming in future PRs:
* A hook for convenient access to flag status outside of the React rendering tree
* A method for convenient access to flag status within TypeScript API code
* Feature flag expiry comments

**Why**
[SCT-495](https://hackney.atlassian.net/browse/SCT-495)


components/Layout/Footer/Footer.tsx
```diff
@@ -1,20 +1,32 @@
...
+          </a>
+          <a href="https://docs.google.com/forms/d/e/1FAIpQLScILbPD1ioKHzp1D3HN4_DKaxV2tpWLMu8upSSqNgSPCo85cg/viewform">
+            Give feedback
+          </a>
+        </nav>
+
+        <div className={s.meta}>Built and maintained by HackIT.</div>
+
+        <ConditionalFeature name="feature-flags-implementation-proof">
+          {user?.hasAdminPermissions && (
```
Renny4Real, line 23: Is there a difference between having `dev permissions` and having `admin permissions`?
What does `hasAdminPermissions` check for?

pages/_app.tsx
```diff
@@ -70,8 +81,20 @@ CustomApp.getInitialProps = async (
...
     }
   }
+
   const appProps = await App.getInitialProps(appContext);
-  return { ...appProps, pageProps: { ...appProps.pageProps, user } };
+
+  const environmentName = [
+    'social-care-service-staging.hackney.gov.uk',
+    'dev.hackney.gov.uk:3000',
+  ].includes(process.env.REDIRECT_URL || '')
```
Renny4Real, line 90: What would `.includes(process.env.REDIRECT_URL || '')` help us do?

Renny4Real (approved):
Looks really cool! 🎉

components/Layout/Footer/Footer.tsx
chrishutchinson: Looks like it checks for either `admin` or `dev` permissions under the hood:

```
hasAdminPermissions:
      groups.includes(AUTHORISED_ADMIN_GROUP) ||
      groups.includes(AUTHORISED_DEV_GROUP),
```

pages/_app.tsx
chrishutchinson: This is effectively saying: "if the value of `REDIRECT_URL` is one of our known development URLs, then set the environment name to "development". This is intentionally defensive, so that if we happened to change our production URL, this feature flag wouldn't accidentally be activated.

PR #488, opened 2021-06-21
Feature SCT-625 add require field to review 3c form

Miles-Alford:
**What**  
This PR adds the required fields on the review 3c form, Each flexible form component now also displays a red asterisk if the field is required  

**Why**  
To allow the review 3c form on the system to contain the correct required fields like on the google form.

**Anything else?**
A couple of issues exist on some of the components, for starters the `SelectField` component does not follow validation rules if it is required, furthermore any component which contains a condition will also not follow validation rules if it is set as required.


components/FlexibleForms/CheckboxField.tsx
```diff
@@ -41,7 +42,9 @@ const Field = ({
       className="govuk-fieldset"
       aria-describedby={hint && `${name}-hint`}
     >
-      <legend className="govuk-label lbh-label">{label}</legend>
+      <legend className="govuk-label lbh-label">
+        {label} {required && <span className="govuk-required">*</span>}
```
jhackett1, line 46: i believe the correct markup for these spans is:

```
<span className="govuk-required">
    <span aria-hidden="true">*</span>
    <span className="govuk-visually-hidden">required</span>
</span>
```

that way they have a proper semantic meaning rather than just a visual one.

PR #489, opened 2021-06-21
fix(SCT-251): show 'manage workers' menu item for admins

chrishutchinson:
**What**  
The "Manage workers" menu item wasn't showing for admins (only devs). This fixes that!

**Why**  
[SCT-251](https://hackney.atlassian.net/browse/SCT-251)

### DIFF
--- a/components/Layout/Header/Header.tsx
+++ b/components/Layout/Header/Header.tsx
@@ -4,6 +4,7 @@ import Link from 'next/link';
 import { useAuth } from 'components/UserContext/UserContext';
 import { getUserType } from 'utils/user';
 import Logo from './Logo';
+import { canUserManageWorkers } from '../../../lib/permissions';
 
 const HeaderComponent = ({
   serviceName,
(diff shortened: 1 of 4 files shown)


components/Layout/Header/Header.tsx
```diff
@@ -48,7 +48,10 @@ const HeaderComponent = ({
         loggedNavLinks
           .filter(({ name }) => name !== 'Forms in progress' || savedForms)
           .filter(
-            ({ name }) => name !== 'Manage workers' || user.hasDevPermissions
+            ({ name }) =>
+              name !== 'Manage workers' ||
+              user.hasDevPermissions ||
+              user.hasAdminPermissions
```
Renny4Real, line 54: Not sure I understand what's happening here 🤔 

I'm reading this as the list/array of `loggedNavLinks` is being filtered and would return elements that pass one of the conditions and the new filtered list would not contain 'Manage Workers' because of this condition `name !== 'Manage workers'` but that's not correct is it? 😅

Renny4Real (commented):
A bit confused by the code 😅

components/Layout/Header/Header.tsx
chrishutchinson: Yeah honestly this code is really really weird. I think it's saying "don't show `manage workers` unless you're a dev or admin.

Really, on second thought, I think we should modify how this works. Perhaps we should close this PR in favour of a more elegant solution! One for stand up in the morning

Renny4Real: 👍🏾

components/Layout/Header/Header.tsx
```diff
@@ -71,17 +30,24 @@ const HeaderComponent = ({
...
+                <a className="govuk-header__link">Search</a>
+              </Link>
+              <Link href="/">
+                <a className="govuk-header__link">My work space</a>
+              </Link>
+              {canUserManageWorkers(user) && (
+                <Link href="/workers">
+                  <a className="govuk-header__link">Manage workers</a>
+                </Link>
+              )}
```
Renny4Real, line 45: When I hard code the URL as an adult user, I can get to the manage worker's page. 
Would we need to have some redirect for this use case too?

Renny4Real (approved):
🎉

components/Layout/Header/Header.tsx
chrishutchinson: @Renny4Real yeah good question – at the moment none of our routes are protected from people hard coding the URL. I think we _should_ change that, but I think that's beyond the scope of this PR

PR #491, opened 2021-06-22
Feature/conditional fields can be required

jhackett1:
some fields on a flexible form might be both `required` and have a `condition`. the intended behaviour here is that the form is required _only_ when its condition is met.

this can't be done with our initial validation approach, because the `validationSchema` is created "before" the user interacts with the form, so there's no way of modifying the schema based on answers the user gives.

to get around this, we've added a second "level" of validation using formik's `validate` prop.

we've written a new validator `validateConditionalField` which will, if all is well, return an empty object `{}`.

however, if a field is:

- conditional
- required
- the condition is met
- the current value of the field is either an empty array or empty string

...then the correct error message will be added to the errors object.

eg. `foo: "This question is required"`

formik has no trouble with these two validation approaches coexisting, but we should double-check for any unintended changes.

### DIFF
--- a/lib/validators.ts
+++ b/lib/validators.ts
@@ -73,6 +73,7 @@ export const validateConditionalFields = (
       field.condition &&
       values[field.condition.id] === field.condition.value &&
       field.required &&
+      // is the current value an empty string or array?
       !values[field.id]?.length
     ) {
       errors[field.id] = getErrorMessage(field);
--- a/lib/validators.ts
+++ b/lib/validators.ts
@@ -1,4 +1,5 @@
 import * as Yup from 'yup';
+import { FormikValues, FormikErrors } from 'formik';
 import { Field } from 'data/flexibleForms/forms.types';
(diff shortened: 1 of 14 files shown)


lib/validators.ts
```diff
@@ -36,20 +46,37 @@ export const generateFlexibleSchema = (
...
+  values: FormikValues,
+  fields: Field[]
+): FormikErrors<FormikValues> => {
+  const errors: FormikErrors<FormikValues> = {};
+  fields.map((field) => {
+    if (
+      field.condition &&
+      values[field.condition.id] === field.condition.value &&
+      field.required &&
+      !values[field.id]
```
jhackett1, line 76: we're just testing for a falsy value (eg. empty string, empty array). are there situations where this won't work?

anabebs (approved):
Looks fine. still had that unsaved changes pop up a couple of times especially on pages with required questions.

Miles-Alford (approved):
👍 looks good.

PR #493, opened 2021-06-22
flexible form cleanup and bug fixes

jhackett1:
- form submit button now has a type attribute to help older browsers
- autosave delay is now 2 seconds again
- an invalid form, once validated, _won't_ now auto-return to the task list, like it was before

### DIFF
--- a/components/FlexibleForms/StepForm.tsx
+++ b/components/FlexibleForms/StepForm.tsx
@@ -72,8 +72,11 @@ const StepFormInner = ({
   const { saved, setSaved } = useAutosave();
   const router = useRouter();
 
-  if (goBackToTaskList && saved && isValid)
+  if (goBackToTaskList && saved && isValid) {
     router.push(`/submissions/${router.query.id}`);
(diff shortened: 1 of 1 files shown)


components/FlexibleForms/StepForm.tsx
```diff
@@ -72,8 +72,11 @@ const StepFormInner = ({
   const { saved, setSaved } = useAutosave();
   const router = useRouter();
 
-  if (goBackToTaskList && saved && isValid)
+  if (goBackToTaskList && saved && isValid) {
     router.push(`/submissions/${router.query.id}`);
+  } else {
```
Miles-Alford, line 77: For my own understanding, why use this syntax rather than a normal if else?

Miles-Alford (approved):

components/FlexibleForms/StepForm.tsx
jhackett1: very good point!

PR #495, opened 2021-06-22
show unfinished submissions on dashboad

jhackett1:
this overhauls the "unfinished submissions" tab of the dashboard to show cloud-backed things from the new form builder, satisfying https://hackney.atlassian.net/browse/SCT-629.

at the moment, only forms in progress that are saved in the browser's local storage will be shown on the dashboard. with this, social workers can collaborate on casework.

it takes advantage of the opportunity to implement ruksana's new designs for this part of the dashboard.

it makes use of this new api endpoint: https://hackney.atlassian.net/browse/SCT-630

we:

- add a new method to `lib/submissions.ts` to grab all unfinished submissions
- add a new `SubmissionsTable` component that displays the unfinished submissions data
- add components to handle client-side searching (just by resident id and name for now) and filtering of the submissions list
- lightly tweak the existing `SaveFormData` component so that, when the two are both displayed on the same page together, it doesn't look weird

once all forms have been moved over to the new form builder, we can remove and deprecate the `SaveFormData` component entirely and use this one for everything.

**important:** this can be safely launched _now_, in advance of the new form builder launching, because when there are no unfinished submissions, the whole component simply hides itself.

in the near future we may want to swap out my basic handwritten concatenate+filter+includes search in favour of https://lunrjs.com/docs/index.html or similar

### DIFF
--- a/components/SubmissionsTable/SubmissionRow.tsx
+++ b/components/SubmissionsTable/SubmissionRow.tsx
@@ -88,18 +88,19 @@ const SubmissionRow = ({
       {open && (
         <dl className={`${s.detailsPanel}`}>
           <div>
-            <dd>{format(new Date(lastEdited), 'dd MMM yyyy K.mm aaa')}</dd>
             <dt className="lbh-body-s">Last edited</dt>
+            <dd>{format(new Date(lastEdited), 'dd MMM yyyy K.mm aaa')}</dd>
           </div>
 
           <div>
+            <dt className="lbh-body-s">Started</dt>
             <dd>
               {format(new Date(submission.createdAt), 'dd MMM yyyy K.mm aaa')}
             </dd>
-            <dt className="lbh-body-s">Started</dt>
           </div>
 
           <div>
+            <dt className="lbh-body-s">Progress</dt>
             <dd>
               {completedSteps === totalSteps ? (
                 'Ready to finish'
@@ -111,11 +112,11 @@ const SubmissionRow = ({
                 </>
               )}
             </dd>
-            <dt className="lbh-body-s">Progress</dt>
           </div>
 
           {editors?.length > 0 && (
             <div>
+              <dt className="lbh-body-s">Edited by</dt>
               <dd>
                 <ul className="lbh-list govuk-!-margin-top-0">
                   {editors.map((editor) => (
@@ -123,7 +124,6 @@ const SubmissionRow = ({
                   ))}
                 </ul>
               </dd>
-              <dt className="lbh-body-s">Edited by</dt>
             </div>
           )}
         </dl>
--- a/components/SubmissionsTable/SubmissionRow.tsx
+++ b/components/SubmissionsTable/SubmissionRow.tsx
@@ -21,6 +21,10 @@ const SubmissionRow = ({
   const open = openRow === submission.submissionId;
 
   const form = forms.find((form) => form.id === submission.formId);
+  
+  if(!form) {
+    return null;
+  }
 
   const completedSteps = Object.keys(submission.formAnswers).length;
   const totalSteps = form?.steps?.length;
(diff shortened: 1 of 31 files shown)


components/SubmissionsTable/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  if (searchQuery) {
+    filteredSubmissions = filteredSubmissions.filter((submission) => {
+      const haystack = `${submission.residents[0].id} ${submission.residents[0].firstName} ${submission.residents[0].lastName}`;
+      if (haystack.toLowerCase().includes(searchQuery.toLowerCase()))
+        return submission;
+      return;
+    });
+  }
+
```
chrishutchinson, line 89: If I've understood your intentions correctly, I think you can consolidate this double filter and improve comprehension, something like this:

```suggestion
  const filteredSubmissions = submissions.filter((submission) => {
    // Filter out those that don't match the search term
    const haystack = `${submission.residents[0].id} ${submission.residents[0].firstName} ${submission.residents[0].lastName}`;

    if (!haystack.toLowerCase().includes(searchQuery.toLowerCase())) {
      return false;
    }

    // If showing only the current user's submissions, then filter out if the emails don't match
    if (filter === "mine" && submission.createdBy.email !== user?.email) {
      return false;
    }

    // Otherwise, this record is good to show
    return true;
  });
```

lib/submissions.spec.ts
```diff
@@ -12,6 +13,30 @@ const mockedAxios = axios as jest.Mocked<typeof axios>;
 
 const { ENDPOINT_API, AWS_KEY } = process.env;
 
+describe('getUnfinishedSubmissions', () => {
+  it('should work properly', async () => {
```
chrishutchinson, line 17: We can do better with the test names here

```suggestion
  it('should return a list of incomplete submissions', async () => {
```

chrishutchinson (changes requested):
Looking good, a few initial thoughts

components/SubmissionsTable/SubmissionRow.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const editors = submission.workers
+    .map((worker) => worker.email)
+    .filter((editor) => editor !== user.email);
+
+  return (
+    <>
+      <li className={s.row}>
+        <div className={s.person}>
+          <h3 className={s.name}>
+            <Link href={`/people/${submission.residents[0].id}`}>
```
chrishutchinson, line 40: Below you check if `residents` is null before accessing the first item, but you don't here? Is that a type issue, a mistake, or something else?

chrishutchinson (changes requested):
A bonus comment!

components/SubmissionsTable/SubmissionRow.tsx
jhackett1: the proper shape of this will be determined once we do group recording, but safe to assume it'll always be defined for now

components/SubmissionsTable/SubmissionRow.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+const SubmissionRow = ({
+  submission,
+  openRow,
+  setOpenRow,
+  user,
+}: Props): React.ReactElement => {
+  const open = openRow === submission.submissionId;
+
+  const form = forms.find((form) => form.id === submission.formId);
```
chrishutchinson, line 23: Do we want to do an explicit check here for `form === false` and return null? I can see a scenario where, for whatever reason, in the future a submission is available for a form that has been renamed or no longer exists, and I think this would blow up.

Something like:
```suggestion
  const form = forms.find((form) => form.id === submission.formId);
  
  if(!form) {
    return null;
  }
```

chrishutchinson (changes requested):
Sorry, one more quick thing as I was re-reading!

chrishutchinson (approved):
Nice work!

PR #496, opened 2021-06-22
fix(SCT-659): show warning note summaries to correct users when viewing residents

chrishutchinson:
**What**  
Warning Note summaries weren't being show to users correctly as an old feature flag was left over. This has been resolved, and new tests for the behaviour have been added.

**Why**  
[SCT-659](https://hackney.atlassian.net/browse/SCT-659)


Renny4Real (approved):
🎉

PR #497, opened 2021-06-22
SCT-645 Install google tag manager code to website 

mapleleafjack:
[Link to ticket](https://hackney.atlassian.net/browse/SCT-645)

**What**  
Added google tracking script in head and body tag

**Why**  
so we can start tracking events.

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -52,6 +52,15 @@ export default class AppDocument extends Document {
           ) : null}
         </Head>
         <body className="govuk-template__body lbh-template__body js-enabled">
+          <noscript
+            dangerouslySetInnerHTML={{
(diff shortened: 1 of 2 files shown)


pages/_document.tsx
```diff
@@ -24,6 +24,17 @@ export default class AppDocument extends Document {
           `,
             }}
           />
+          <script
+            dangerouslySetInnerHTML={{
```
Renny4Real, line 28: _dangerous_ 😎

Renny4Real (approved):
👍🏾

pages/_document.tsx
mapleleafjack: 🔫🔫🔫

PR #503, opened 2021-06-24
SCT-413 - Update relationships to support the new designs

mapleleafjack:
[Link to ticket ](https://hackney.atlassian.net/browse/SCT-413)

**What**  
The Relationship component and data structure has been changed.
Alphabetical order has been added to the Types list and Relationship person list

**Why**  
To reflect the new designs / backend response

### DIFF
--- a/components/Relationships/RelationshipElement.tsx
+++ b/components/Relationships/RelationshipElement.tsx
@@ -1,15 +1,15 @@
 import Link from 'next/link';
-import { RelationshipPerson } from 'types';
+import { RelationshipPerson, RelationshipType } from 'types';
 
 interface Props {
-  type: string;
+  type: RelationshipType;
   persons: RelationshipPerson[];
 }
 
 const RelationshipElement = ({ type, persons }: Props): React.ReactElement => {
   return (
     <div className="govuk-summary-list__row">
-      <dt className="govuk-summary-list__key">{renderTitle(type)}</dt>
+      <dt className="govuk-summary-list__key">{getTitleString(type)}</dt>
       <dd className="govuk-summary-list__value">
         <ul className="govuk-list">
           {persons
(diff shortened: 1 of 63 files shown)


components/Relationships/RelationshipElement.tsx
```diff
@@ -2,35 +2,118 @@ import Link from 'next/link';
 import { RelationshipPerson } from 'types';
 
 interface Props {
-  title: string;
-  data: RelationshipPerson[];
+  type: string;
+  persons: RelationshipPerson[];
```
chrishutchinson, line 6: `people`?

components/Relationships/RelationshipElement.tsx
```diff
@@ -2,35 +2,118 @@ import Link from 'next/link';
...
+            })}
         </ul>
       </dd>
     </div>
-  ) : (
-    <></>
   );
 };
 
+function renderTitle(title: string) {
```
chrishutchinson, line 40: Does this make more sense to be an object type rather than a massive `switch`? Something like:

```tsx
const mappings = {
  parents: 'Parents',
  parent: 'Parent',
  children: 'Children',
  // .. and so on ..
} as const;

const getTitleString = (relationshipType: keyof typeof mappings): string => {
  return mappings[relationshipType];
};
```

That magic `keyof typeof mappings` also means you get a nice type signature for your `getTitleString` method. That might be helpful, it might not, just a though 😄

components/Relationships/Relationships.tsx
```diff
@@ -17,14 +17,10 @@ const Relationships = ({ id }: Props): React.ReactElement => {
...
 
-  const shouldAppear =
-    (personalRelationships.parents &&
-      personalRelationships.parents.length > 0) ||
-    (personalRelationships.children &&
-      personalRelationships.children.length > 0) ||
-    (personalRelationships.siblings &&
-      personalRelationships.siblings.length > 0) ||
-    (personalRelationships.other && personalRelationships.other.length > 0);
+  let shouldAppear = false;
```
chrishutchinson, line 20: I _think_ this might need to be in state if you want React to properly re-render when it changes? 

```tsx
const [shouldAppear, setShouldAppear] = useState(false);
```

Then use `setShouldAppear` to change the value?

components/Relationships/Relationships.tsx
```diff
@@ -17,14 +17,10 @@ const Relationships = ({ id }: Props): React.ReactElement => {
...
+  personalRelationships.map((elm) => {
```
chrishutchinson, line 21: What's `elm`?

types.ts
```diff
@@ -213,18 +213,17 @@ export interface RelationshipPerson {
   id: number;
   firstName: string;
   lastName: string;
+  gender?: string;
```
chrishutchinson, line 216: Why optional?

types.ts
```diff
@@ -213,18 +213,17 @@ export interface RelationshipPerson {
...
 }
 
 export interface Relationship {
-  parents: Array<RelationshipPerson>;
-  siblings: Array<RelationshipPerson>;
-  children: Array<RelationshipPerson>;
-  other: Array<RelationshipPerson>;
+  type: string;
+  persons: RelationshipPerson[];
```
chrishutchinson, line 221: `people`?

components/Relationships/RelationshipElement.tsx
```diff
@@ -2,35 +2,118 @@ import Link from 'next/link';
 import { RelationshipPerson } from 'types';
 
 interface Props {
-  title: string;
-  data: RelationshipPerson[];
+  type: string;
```
chrishutchinson, line 5: Can `type` be literally any string, or is there a valid subset of `types` that could form some kind of union type?

components/Relationships/Relationships.tsx
mapleleafjack: I will rename it to "relationship" to make it clearer

types.ts
mapleleafjack: We decided that the gender is optional in the first iteration

types.ts
mapleleafjack: that is what comes from the backend response :)

components/Relationships/RelationshipElement.tsx
mapleleafjack: that is what comes from the backend response :)

components/Relationships/Relationships.tsx
mapleleafjack: Oh the page is reloaded every time so it re-renders correctly.
I will change it anyways to useState as you're right, it's probably cleaner :)

components/Relationships/RelationshipElement.tsx
mapleleafjack: There is a valid subset of available types indeed!

I will add the type "RelationshipType" so this is more robust

components/Relationships/RelationshipElement.tsx
mapleleafjack: This sounds much better, thanks!
Will definitely implement this :)

chubberlisk (approved):
Looks good to me! 🚢

PR #509, opened 2021-06-28
SCT-414 - reversed ordering

mapleleafjack:
**What**  
Reversed the ordering in the frontend for Relationships

**Why**  
The order was alphabetical but reversed

### DIFF
--- a/components/Relationships/RelationshipElement.spec.tsx
+++ b/components/Relationships/RelationshipElement.spec.tsx
@@ -33,7 +33,7 @@ describe('Relationships component', () => {
     expect(getByText('Unborn sibling')).toBeInTheDocument();
     expect(getByText('Jambi Neverborn')).toBeInTheDocument();
   });
-  it('should populate the list in alphabetical order', async () => {
+  it('should populate the list in alphabetical order (by surname/name) with same surname', async () => {
     const { queryAllByText } = render(
       <RelationshipElement {...mockedOrderedRelationship} />
     );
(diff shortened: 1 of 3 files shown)


chubberlisk (commented):
Are we able to add a test for this?

mapleleafjack:
> Are we able to add a test for this? Looks like we could do something like this: [testing-library/react-testing-library#313 (comment)](https://github.com/testing-library/react-testing-library/issues/313#issuecomment-625294327).

Hey Ting, thank you for the link, I was struggling to find documentation about testing ordering!

components/Relationships/RelationshipElement.spec.tsx
```diff
@@ -32,4 +33,14 @@ describe('Relationships component', () => {
     expect(getByText('Unborn sibling')).toBeInTheDocument();
     expect(getByText('Jambi Neverborn')).toBeInTheDocument();
   });
+  it('should populate the list in alphabetical order', async () => {
+    const { queryAllByText } = render(
+      <RelationshipElement {...mockedOrderedRelationship} />
```
chubberlisk, line 38: Shouldn't the relationships we provide be unordered so we know that the code then orders it?

When we say alphabetical order, do we mean by first name or last name? Maybe we can update the test description to be a bit more explicit.

Could we also provide another test example where the names don't have the same surname?

chubberlisk (commented):
Thanks for for looking into this! Couple of questions and suggestions.

components/Relationships/RelationshipElement.spec.tsx
mapleleafjack: Fair enough :D

I've now 
-mixed the people's order (so they're not in order in the tests)
-added different name/surname in the list
-added a proper description to the test

chubberlisk (approved):
Thank you for addressing my comments! 🚀

PR #511, opened 2021-06-28
feat(SCT-605, SCT-720): add new roles

chrishutchinson:
**What**  
A handful of requests have come in to introduce new roles to the system, this does just that!

**Why**  
[SCT-605](https://hackney.atlassian.net/browse/SCT-605), [SCT-720](https://hackney.atlassian.net/browse/SCT-720)


martapederiva (approved):
👍

PR #512, opened 2021-06-28
feat(SCT-724): add an automated release documentation tool

chrishutchinson:
BREAKING CHANGE

**What**  
(It's not a breaking change, but this _should_ bump us to `v1.0.0`!)

This PR introduces some scripts that run in CircleCI to automatically generate release notes, and tag releases in GitHub. This gives us a couple of benefits:
1. The draft release generated on merge to `main` will allow us to see what is ready to release in one place
2. Once a release to production is made, that release will be marked as "published" and a version increment will be automatically applied to our `package.json` file. The full changelog will be attached to that release in GitHub, so we will know what was released 🎉 
3. We can always see when the last release was. When there is an active production incident, it will be easy to see when the last production change was.

**Future improvements:** In future, we could automatically publish release notes into Slack, for wider team visibility 👀!

**One requirement:** this PR uses the conventional commit pattern in order to determine whether to make a major, minor, or hotfix version number change. It also uses the change type (`feat`, `fix`, `chore` etc.) to create nicely structured release notes. We will need to more rigorously follow this pattern for our changes. In general, most of the team already are (👏👏👏!), but to make it easier we may need to introduce an explicit check to ensure the pattern is followed.

**Why**  
To make it much easier to see what changes are about to be made, have been made, or were made in the past!

[SCT-724](https://hackney.atlassian.net/browse/SCT-724)

**Anything else?**

  - `release-it`: https://github.com/release-it/release-it
  - ~We need to add a `GITHUB_TOKEN` env var to CircleCI. We can use a personal token tied to one of the team's accounts but that is not particularly good practice! We will ideally use a Hackney machine user account. TBC on that~ – Done! Matt K sorted me out a token for this GitHub account: https://github.com/lbhackney-automation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,22 @@ jobs:
           name: deploy
           command: yarn build && yarn --production=true && sudo npm i -g serverless && sls deploy -s production
 
+  dry-run-generate-github-release:
+    executor: aws-cli/default
+    steps:
+      - *attach_workspace
+      - run:
+          name: generate release
+          commant: yarn release --dry-run
+
+  generate-github-release:
+    executor: aws-cli/default
+    steps:
+      - *attach_workspace
+      - run:
+          name: generate release
+          commant: yarn release
+
   assume-role-staging:
     executor: docker-python
     steps:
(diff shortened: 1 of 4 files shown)


mapleleafjack (approved):
This is really a nice feature to have!

neilkidd (approved):
'release-it`!

PR #516, opened 2021-06-29
fix: clear out old draft releases before creating a new one

chrishutchinson:
**What**  
When we create draft releases with `release-it`, instead of extending the previous draft, it creates a new one, meaning we end up with a deluge of draft releases!

I've written a small script which clears up old draft releases automatically, meaning the one `release-it` creates is always fresh, and contains all changes since the last production releases. 

This draft is then removed and replaced when a full production release is made.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,16 +88,22 @@ jobs:
     steps:
       - *attach_workspace
       - run:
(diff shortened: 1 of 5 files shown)


neilkidd (approved):
LGTM

PR #521, opened 2021-06-29
fix(SCT-661): display warning notes for all users

chrishutchinson:
**What**  
We did a fix to ensure practitioners could see warning notes for their residents last week. This extends that change to show warning notes for all users (but only users with the right permissions can `review / end` them).

**Why**  
[SCT-661](https://hackney.atlassian.net/browse/SCT-661)

### DIFF
--- a/components/WarningNote/WarningNotes.spec.tsx
+++ b/components/WarningNote/WarningNotes.spec.tsx
@@ -32,7 +32,7 @@ describe(`useWarningNotes`, () => {
     jest.spyOn(warningNotes, 'useWarningNotes').mockImplementation(() => ({
       data: [
         warningNoteFactory.build({
-          nextReviewDate: '2020-12-30',
+          nextReviewDate: '2020-11-12',
         }),
       ],
       mutate: jest.fn(),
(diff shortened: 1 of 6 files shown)


components/WarningNote/WarningNotes.spec.tsx
```diff
@@ -13,33 +28,44 @@ jest.mock('next/router', () => ({
...
-    expect(asFragment()).toMatchSnapshot();
-  });
-
-  it('should not render the review date if there is a scheduled future review date', () => {
+  beforeEach(() => {
     jest.spyOn(warningNotes, 'useWarningNotes').mockImplementation(() => ({
       data: [
         warningNoteFactory.build({
-          nextReviewDate: '2020-11-12',
+          nextReviewDate: '2020-12-30',
```
martapederiva, line 35: Why the date change?

Renny4Real, line 35: Why do we change the date here? e.g. preference? Date weirdness in tests? 😁

Renny4Real (approved):
Looks good 🎉

components/WarningNote/WarningNotes.spec.tsx
chrishutchinson: @Renny4Real: @martapederiva had exactly the same question, and I didn't have a good answer 😅 . I'm going to take a look and will report back, it's possible I changed it in error!

chrishutchinson: @Renny4Real @martapederiva: Looked like an oversight. Restoring it to the original date, things are still working as expected 👍

PR #524, opened 2021-06-30
fix: only build the app once in CI

chrishutchinson:
**What**  
At the moment our pipeline builds the app three times:
1. To run end-to-end tests
2. To deploy to staging
3. (Optionally) To deploy to production

The downside here is that _technically_ the artefact used for running our tests is not the same artefact we release to staging and production. This introduces a small opportunity for bugs to be introduced. To eliminate this, and also speed up the whole process, I have split out the build step as an independent job that runs just once, after the unit tests have passed.

_Draft while I do some testing in CI_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
       - persist_to_workspace:
           root: *workspace_root
           paths:
-            - ./build
+            - project/build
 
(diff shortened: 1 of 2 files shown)


neilkidd (approved):
LGTM - woo hoo

PR #525, opened 2021-07-01
feat(SCT-471) Add relationship

mapleleafjack:
[Ticket link](https://hackney.atlassian.net/browse/SCT-471)

**What**  
This PR adds the main necessary bits in the UI that lets the user add a relationship
As of now the feature is behind a flag for development only.

**Why**  
To let users add relationships in the system

**Things that needs to be done after**
-Style is a bit off, labels probably need to be changed
-The callback function in AddRelationship.tsx/Search.tsx needs to be changed from a prop to a Feature Provider (It currently work, but it’s probably not the proper way of doing it)
-Add more tests


chubberlisk (approved):
As discussed, we've got a number of improvements to make but we want to get this in first under a feature flag and then iterate from there!

PR #526, opened 2021-07-01
feat(SCT-738): add a `useFeatureFlag()` hook for accessing feature flags

chrishutchinson:
**What**  
Following the introduction of `<ConditionalFeature />`, this PR introduces a new hook for accessing the state of a feature flag during scripting (outside of the JSX rendering tree), for example, inside a `useEffect`.

**Why**  
[SCT-738](https://hackney.atlassian.net/browse/SCT-738)

**Anything else?**

  - Yes – [`@testing-library/react-hooks`](https://github.com/testing-library/react-hooks-testing-library)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -22,5 +22,5 @@ module.exports = {
   moduleNameMapper: {
     '^.+\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
   },
-  moduleDirectories: ['node_modules', '.'],
+  moduleDirectories: ['node_modules', '<rootDir>/node_modules', '.'],
 };
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -28,6 +28,10 @@ You can add a new feature to it following this structure:
 
 > ⚠️ **Warning**: Make sure to add an expiry date comment to the feature flag as shown above.
 
+### Using the feature flag
+
+#### React component
+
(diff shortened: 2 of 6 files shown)


jest.config.js
```diff
@@ -22,5 +22,5 @@ module.exports = {
   moduleNameMapper: {
     '^.+\\.module\\.(css|sass|scss)$': 'identity-obj-proxy',
   },
-  moduleDirectories: ['node_modules', '.'],
+  moduleDirectories: ['node_modules', '<rootDir>/node_modules', '.'],
```
chrishutchinson, line 25: GitHub issue explaining this change: https://github.com/testing-library/react-hooks-testing-library/issues/294#issuecomment-663900081

Basically, when `'.'` is used as a value here, Jest's module resolution logic looks for the nearest directory called `react`, which doesn't work in `@testing-library/react-hooks` because it finds `@testing-library/react` (which lives in a folder called `react`). This addition makes it check the root directory's `node_modules` before looking nearer to home.

Renny4Real: Did we make this change because we wanted to test the react hooks, or was this causing some issues we might have ignored before?

Renny4Real (approved):
Nice! 🎉

jest.config.js
chrishutchinson: Yeah – I've introduced `@testing-library/react-hooks` for testing the behaviour of `useFeatureFlags()`. Imported here: https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/526/files#diff-69f13c2c73ed393480bf7b704abc2b850fe2e647dc03d35d29cf5ac5c98e8702R8

PR #532, opened 2021-07-05
refactor(SCT-471): update add relationship form to use checkboxes

chubberlisk:
**What**  

- Update the add relationship form to use checkboxes instead of radio buttons.
- Remove `isInformalCarer` from request.
- Add view person button link when successful.
- Sort the relationship types in alphabetical order.
- Add tests for `utils/api/addRelationships` function and `AddRelationshipForm` component.

Before | After
-- | --
Screenshot 2021-07-05 at 17 50 28 | Screenshot 2021-07-05 at 17 50 55

**Why**  

There's been a rethink of the design to improve form for the additional fields.

**Anything else**

- Would recommend reviewing commit by commit.


jhackett1 (approved):
this looks fine in terms of what it covers, but take care building new functionality on the classic `<FormWizard/>`, because we will be deprecating this in future. try to use the new 'flexible forms' components where you can.

utils/summary.jsx
```diff
@@ -47,19 +47,22 @@ export const formatData = (componentProps, formData) => {
...
     return {
       key: name,
       title: label,
       value:
-        typeof options[0] === 'string'
+        typeof stepOptions[0] === 'string'
           ? formData[name]
-          : formData[name]?.map((data) =>
-              multiValue(
+          : formData[name]?.map((data) => {
```
JohnFarrellDev, line 59: Screenshot from 2021-07-08 09-59-49

Looks like we're trying to call .map() against a string and it throws an error when trying to create the relationship.

utils/api/relationships.ts
```diff
@@ -1,16 +1,47 @@
...
   id: number
 ): SWRResponse<RelationshipData, ErrorAPI> =>
   useSWR(`/api/residents/${id}/relationships`);
 
+interface addRelationshipFormData {
+  personId: number;
+  otherPersonId: number;
+  type: string;
+  additionalOptions?: string[] | boolean;
+  isMainCarer?: string;
```
JohnFarrellDev, line 15: Should this be a boolean?

Looks like we're representing true as "Y" and false as "N". (if the backend returns information like this we should modify the `.toResponse()` to convert the string representation to an actual boolean.

https://github.com/LBHackney-IT/social-care-case-viewer-api/blob/master/SocialCareCaseViewerApi/V1/Factories/ResponseFactory.cs#L145

JohnFarrellDev: Ah I noticed this is a post but I would still advise using boolean, if our database needs to use "Y" or "N" we can handle that when creating the Relationship object

data/forms/add-relationship.tsx
```diff
@@ -16,42 +16,42 @@ const formSteps: FormStep[] = [
...
+        component: 'Checkbox',
+        name: 'additionalOptions',
+        label: 'Select all that apply',
+        labelSize: 's',
+        options: ({ type }) => {
+          const typeOptions: Option[] = [
+            { value: 'isMainCarer', text: 'Is a main carer' },
+          ];
+
+          if (type === 'parent') {
```
JohnFarrellDev, line 40: These conditionals are cool but I wonder if it could be confusing.

Will a worker know that if a person is pregnant to add the parent relationship and then see they can choose unborn child.

I could see a worker potentially thinking someone isn't a parent yet and then never seeing the unborn child option.

JohnFarrellDev (approved):
Really cool to see you doing frontend stuff now :+1:

JohnFarrellDev:
Trixie doing code reviews

WhatsApp Image 2021-07-06 at 15 47 09

utils/summary.jsx
chubberlisk: Thanks for the spot! I've been trying to debug this and it turns out that either the `Checkbox` or `FormWizard` is returning multiple types: string, an array of strings, empty array or boolean. 

It seems like we'll have to handle that as I'm struggling to figure out which one it is and am wary of changing things in `FormWizard` especially. Let me know if you have any suggestions or spare time today, otherwise I will just have to handle the different types.

jhackett1: i have no insight to offer other than this is an excellent reason to use the new form builder instead!

chubberlisk: @JohnFarrellDev, I've pushed up the fix for this - not ideal but I think it's fine for now. 

@jhackett1, could you 🍐 with me on how we'd do that tomorrow? I'm going to merge this PR though as the functionality is there and I've branched off of this branch with some other functionality but I'd definitely be up with  seeing how the new form builder works!

utils/api/relationships.ts
chubberlisk: Yeah, this is what the service API currently sends when getting relationships and what it requests when creating a relationship. I think you're right, this should be a boolean but as I'm currently focusing on getting the functionality to work I won't address this in this PR. I'll add it to our Jira ticket for a thing to do, sorry!

data/forms/add-relationship.tsx
chubberlisk: You make a good point - definitely a thing we will look out for feedback for when we do the initial release.

JohnFarrellDev (approved):
:rocket:

PR #535, opened 2021-07-06
fix: add the `--frozen-lockfile` flag when installing deps in CI

chrishutchinson:
**What**  
We [recently had a failure when publishing a new release from CI](https://app.circleci.com/pipelines/github/LBHackney-IT/lbh-social-care-frontend/2771/workflows/1da9640e-0f9b-431e-89de-7b9d394369d7/jobs/8279) because the `yarn.lock` file had changed when the dependencies were installed in CI. This is partially because we had managed to get an incorrect `yarn.lock` file in our `main` branch, but also because we don't use the `--frozen-lockfile` flag (docs: https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-frozen-lockfile).


Renny4Real (approved):
🎉

PR #536, opened 2021-07-06
feat(SCT-538): Add safeguard adult manager decision form

Miles-Alford:
**What**  
This PR adds the safeguard adult manager decision form to the application.

**Why**  
To allow practitioners to use the form on the system

**Anything else?**
One section on the form is conditional which I am not 100% sure how to go about solving as we do not have conditional rendering sections. The section in question is the `Decision on the concern` section. The other two section's within that theme should only be filled in if they answer yes on a question within the former.


jhackett1:
this broadly looks fine, but i've left comments for small things that should be changed, including:

- sentence case everywhere
- more questions can be made conditional

combining the three steps under the "decision on the concern" theme should solve the issue you mention

i also suggest the following themes:

1. About the person
2. Decision
3. Risks and allegations
4. Feedback and next steps

jhackett1 (approved):

PR #537, opened 2021-07-06
feat(sct-330): approve submissions

jhackett1:
this adds the ability for a user to approve submissions created by another user, similar to how github pr approvals work.

in the `data/flexibleForms`config, forms can now have an `approvable` attribute.

approvable forms, once "finished" become ready for approval.

the email address of an approver can be entered - they are then emailed using notify.

a user other than the creator can then either approve the submission, or return it for edits, with a rejection reason. the original creator will then get a notify-powered email containing the reason.

this feature includes:
- three components to power the ui: `<ApprovalWidget/>` `<ApproveDialog/>` `<RejectDialog/>`
- a `lib/notify.ts` file, containing methods for interacting with notify
- new methods in `lib/submissions.ts` to make approvals and return for edits
- new api routes in `pages/submissions/[id]/approvals.ts` for the frontend to initiate the new operations
- modifications to the task list page and its header, so that a user can add an approver's email address to an approvable form.


components/ApprovalWidget/ApprovalWidget.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+      />
+    );
+    expect(
+      screen.getByText(
+        'You cannot approve your own submissions. Ask a manager or colleague for help.'
+      )
+    );
+  });
+
+  it("doesn't let a user edit their own submissions", () => {
```
Miles-Alford, line 36: Both these tests have the same name. is this intentional?

Miles-Alford (approved):
Looks good and makes sense to me

lib/submissions.spec.ts
```diff
@@ -168,3 +170,46 @@ describe('patchSubmissionForStep', () => {
...
+      `${ENDPOINT_API}/submissions/foo`
+    );
+    expect(mockedAxios.patch.mock.calls[0][1]).toEqual({
+      editedBy: 'bar',
+      submissionState: 'in_progress',
+      rejectionReason: 'test reason',
+    });
+    expect(mockedAxios.patch.mock.calls[0][2]?.headers).toEqual({
+      'x-api-key': AWS_KEY,
+    });
```
JohnFarrellDev, line 213: This could be a lot tidier using `.toHaveBeenCalledWith()`

lib/submissions.spec.ts
```diff
@@ -168,3 +170,46 @@ describe('patchSubmissionForStep', () => {
...
+    expect(mockedAxios.patch).toHaveBeenCalled();
+    expect(mockedAxios.patch.mock.calls[0][0]).toEqual(
+      `${ENDPOINT_API}/submissions/foo`
+    );
+    expect(mockedAxios.patch.mock.calls[0][1]).toEqual({
+      editedBy: 'bar',
+      submissionState: 'approved',
+    });
+    expect(mockedAxios.patch.mock.calls[0][2]?.headers).toEqual({
+      'x-api-key': AWS_KEY,
```
JohnFarrellDev, line 190: This could be a lot tidier using .toHaveBeenCalledWith()

pages/submissions/[id]/index.tsx
```diff
@@ -34,9 +34,14 @@ const TaskListPage = ({
   const completedSteps = Object.keys(formAnswers);
   const person = residents[0];
 
-  const handleFinish = async (): Promise<void> => {
+  const handleFinish = async (values?: {
```
Comment (JohnFarrellDev) [inline pages/submissions/[id]/index.tsx:37]: Why is this optional? Won't we always need the approveEmail to finish a submission

pages/api/submissions/[id]/approvals.ts
```diff
@@ -0,0 +1,50 @@
...
+    case 'POST':
+      {
+        const submission = await approveSubmission(
+          String(id),
+          String(user?.email)
+        );
+        res.json(submission);
+      }
+      break;
+    case 'DELETE':
```
Comment (JohnFarrellDev) [inline pages/api/submissions/[id]/approvals.ts:25]: Why is it using DELETE, shouldn't this be a PATCH?

lib/notify.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/** send an email notification to a creator that their submission has been rejected */
+export const notifyReturnedForEdits = async (
+  submission: Submission,
+  rejecterEmail: string,
+  host: string,
+  rejectionReason?: string
+): Promise<void> => {
+  const notifyClient = new NotifyClient(process.env.NOTIFY_API_KEY);
```
JohnFarrellDev, line 40: Do Hackney as an organisation have access to Notify and do we have a key set for staging/production?

jhackett1: we don't have keys set in production and staging yet. we should do this

components/ApprovalWidget/ApproveDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+  setOpen,
+}: Props): React.ReactElement => {
+  const [loading, setLoading] = useState<boolean>(false);
+  const [status, setStatus] = useState<string | false>(false);
+
+  const handleApprove = async () => {
+    try {
+      setLoading(true);
+      await axios.post(`/api/submissions/${submission.submissionId}/approvals`);
+      window.location.reload;
```
JohnFarrellDev, line 25: Why?
Wouldn't it make more sense to SetLoading to false and to close the dialog?

Maybe a toast message to indicate success.

jhackett1: we don't have toasts or anything in the ui yet. this is what i was meaning to follow up with in a later pr rather than making this one bigger

JohnFarrellDev:
I noticed no snapshot tests or E2E tests, do we benefit from both of these. My guess is that we would.

pages/api/submissions/[id]/approvals.ts
Comment (jhackett1) [reply pages/api/submissions/[id]/approvals.ts:25]: it just uses the method to decide what to do - this is at an `/approvals` sub-route, so POSTing or DELETing your approval makes semantic sense to me

lib/notify.ts
JohnFarrellDev: Do we have approval to get these keys within Hackney? Does seem like something you should have wanted to mention before writing the code.

I don't see how this can be approved until we know we have access to valid Notify keys within Hackney.

jhackett1: we've got a live notify team set up that's been through the process already, i can get the keys - we just gotta add them to our environments

jhackett1:
i think snapshots are a bit of a waste of time - i try to add a "renders correctly"-style test case that looks to make sure the basic elements are there but in a way that won't fail if we reword a heading, say.

end-to-end, i suggest that this be covered in the form builder's e2e test we discussed in planning

lib/submissions.spec.ts
jhackett1: can you give an example? this pattern is throughout the codebase and i agree it looks a bit ugly. toBeCalledWith doesn't seem to be able to let me choose just the headers object

jhackett1: never mind, got it

lib/notify.ts
jhackett1: now added

JohnFarrellDev (approved):
:rocket:

PR #538, opened 2021-07-06
Feat(SCT-739): Add feature flag typescript helper function

Renny4Real:
This adds a helper typescript function that will allow us to interrogate whether the given feature is active or inactive from an API / Node.js file.


pages/_app.tsx
```diff
@@ -34,18 +32,9 @@ const CustomApp = ({
...
-
-    /*
-    The feature-flags-implementation-proof has been setup to have an expiry date in the far future.
-    The FEATURE-FLAG-EXPIRES comment above will cause ESLint errors once the date in the square brackets has passed.
-    Add feature flags below following the format in the example shown.
-    */
-  };
+  const features = getFeatureFlags();
+  features['feature-flags-implementation-proof'].isActive =
+    pageProps.environmentName === 'development';
```
Renny4Real, line 37: Couldn't figure out a way to reference `pageProps` in the new `features.ts` file. 
Is there a way to do this so that we don't have to set the condition outside of the new file?
Open to any suggestions/feedback 😅

chrishutchinson: I'd pass an object into `getFeatureFlags()` which has any data we need to determine if a flag is active. Something like:

```ts
const features = getFeatureFlags({
  environmentName: pageProps.environmentName
});
```

We can extend that object in the future with anything else that might be useful (e.g. the current route)

utils/featureFlagHelpers.ts
```diff
@@ -0,0 +1,7 @@
+import { getFeatureFlags } from '../features';
+
+export const isFeatureFlagActive = (featureName: string) => {
+  const features = getFeatureFlags();
+
+  return features[featureName].isActive;
```
chrishutchinson, line 6: In our hook and component we have some more advanced checks, specifically one that checks if the `featureName` exists in `features`. We should probably look to re-use or replicate that here.

I'm wondering if it actually makes sense for this `isFeatureFlagActive` function to live with all the other feature flags stuff?

Renny4Real: Thanks 🙌🏾 . 
I wasn't sure if it should sit with others, but it's probably clearer to have them all together 👍🏾

Renny4Real: Done in 8608e2b

pages/_app.tsx
Renny4Real: Done in 4735a4f

features.ts
```diff
@@ -0,0 +1,34 @@
+import { FeatureSet } from 'lib/feature-flags/feature-flags';
+
+type options = {
+  [key: string]: any;
+};
+
+export const getFeatureFlags = (options?: options): FeatureSet => {
```
chrishutchinson, line 7: I'd structure this slightly differently, defining `featureFlags` inside the function so they can take advantage of the options / context object. Something like this:

```tsx
type FeatureFlagContext = {
  environmentName: string;
};

export const getFeatureFlags = ({ environmentName }: FeatureFlagContext): FeatureSet => {
  const featureFlags: FeatureSet = {
    /*
      The feature-flags-implementation-proof has been setup to have an expiry date in the far future.
      The FEATURE-FLAG-EXPIRES comment above will cause ESLint errors once the date in the square brackets has passed.
      Add feature flags below following the format in the example shown.
    */
    // FEATURE-FLAG-EXPIRES [3000-12-31]: feature-flags-implementation-proof
    "feature-flags-implementation-proof": {
      isActive: environmentName === "development",
    },
  };

  return featureFlags;
};
```

lib/feature-flags/feature-flags.tsx
```diff
@@ -65,3 +67,12 @@ export const ConditionalFeature: React.FC<{ name: string }> = ({
...
   return <>{children}</>;
 };
+
+/**
+ * Typescript helper function for checking whether a given feature is active or inactive
+ */
+export const isFeatureFlagActive = (featureName: string) => {
+  const features = getFeatureFlags();
+
+  return features[featureName].isActive;
```
chrishutchinson, line 77: I think we can reuse the `isFeatureActive` function here, which will mean we also check that the feature is on the object before looking for the `isActive` attribute. Something like:

```tsx
return isFeatureActive(features)(featureName);
```

chrishutchinson (changes requested):

features.ts
Renny4Real: I see. Thanks! 🙏🏾

Renny4Real: @chrishutchinson My understanding with this is that the only thing we're passing into `FeatureFlagContext` is the environment name, but this could be a list of keys & values of data to use to set conditions for other feature flags?

chrishutchinson: Yeah that's right, an object of things, but I'm proposing we should be explicit about what those things are (for now)

lib/feature-flags/feature-flags.spec.tsx
```diff
@@ -141,4 +149,34 @@ describe('feature flags', () => {
       );
     });
   });
+
+  describe('isFeatureFlagActive() helper function', () => {
```
chrishutchinson, line 153: Can we add a test for the scenario where the feature is not defined and should return false?

So: `it('should return false if the feature is not defined', () => {});`

chrishutchinson (approved):

lib/feature-flags/feature-flags.spec.tsx
Renny4Real: 👍🏾

PR #540, opened 2021-07-07
feature(SCT-519): CFS case notes form

JohnFarrellDev:
**What**  
Currently CFS case notes are processed by a google form. This form should remove the reliance on the google form and allow workers to utilise our form within the system which should be a more effective way of working.

**Anything else?**


data/flexibleForms/childCaseNote.ts
```diff
@@ -117,13 +122,11 @@ const form: Form = {
             },
           ],
         },
-
         {
           id: 'Topics',
```
jhackett1, line 126: needs to be `Tags` to match check below

jhackett1 (approved):

jhackett1:
let's update the `Submission` type to account for it now having a `tags` property?

PR #542, opened 2021-07-08
make adult manager form not be viewable on production

Miles-Alford:
**What**  
This PR makes the safeguarding adult manager concern form not viewable on production.

**Why**  
To test the form on staging first before allowing practitioners access on production


JohnFarrellDev:
One suggestion is to update the AddForm.spec.tsx.
Only change required is updating the variable `flexibleForms`, the 3 `tests.each` should then test your form renders correctly

JohnFarrellDev (approved):
Looks good

PR #545, opened 2021-07-08
feat: add redirects

jhackett1:
this improves our user experience by redirecting people from the submission view page if:

- there is no matching submission, or the submission is discarded (sent to 404)
- the submission is in progress (sent to form task list page)


JohnFarrellDev:
Shouldn't this have some tests?

jhackett1:
maybe! there don't seem to be any tests for the top-level pages - only the components, so i've been following that

JohnFarrellDev (approved):
:+1:

PR #546, opened 2021-07-08
feat(sct-802): revision timeline

jhackett1:
this adds a revision history timeline to every submission. go to `/people/:id/submissions/:submissionId` to see it.

it shows the creation date along with the most recent (up to three) edits.

it uses the `editHistory` array and `createdAt` `createdBy` information right now.

later, we should enhance this to show approvals and perhaps discards. we should also show the percentage completion and perhaps even eventually show people exactly what changed version to version.


components/RevisionTimeline/RevisionTimeline.tsx
```diff
@@ -0,0 +1,42 @@
+import { format } from 'date-fns';
+import { Submission } from 'data/flexibleForms/forms.types';
+
+interface Props {
+  submission: Submission;
+}
+
+const RevisionTimeline = ({ submission }: Props): React.ReactElement | null => {
+  // reverse the array so it's in reverse-chronological order and take the three most recent events
+  const revisions = submission.editHistory.concat([]).reverse().slice(0, 3);
```
JohnFarrellDev, line 10: Are we using .concat only to create a new array object?

I think `Array.from(submission.editHistory).reverse().slice(0, 3);` is cleaner and the code shows the intention

data/flexibleForms/forms.types.ts
```diff
@@ -92,11 +92,13 @@ export interface Submission {
...
-  }[];
+  editHistory: Revision[];
   submissionState: 'In progress' | 'Approved' | 'Discarded' | 'Submitted';
   formAnswers: FlexibleAnswers;
   tags?: string[];
 }
+
+export interface Revision {
+  worker: Worker;
+  editTime: string;
```
JohnFarrellDev, line 103: Shouldn't the type be `Date`?

components/RevisionTimeline/RevisionTimeline.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <>
+      <h2>History</h2>
+      <ol className="lbh-timeline">
+        {revisions.map((revision, i) => (
+          <li
+            className={`lbh-timeline__event lbh-timeline__event--minor ${
+              i === revisions.length - 1 && 'lbh-timeline__event--gap-below'
```
JohnFarrellDev, line 19: Would a CSS psuedoselector to select the last element make more sense here

JohnFarrellDev:
I think this is really good, we definitely need to think about what goes into EditHistory, currently any change which makes displaying the information in a way that is useful difficult.

JohnFarrellDev (approved):
Looks really good, just some minor suggestions

data/flexibleForms/forms.types.ts
jhackett1: it comes in as a string from the api - we convert it to a date just before we display it

components/RevisionTimeline/RevisionTimeline.tsx
jhackett1: all the pseudoselectors are used up! for other things! most of the timeline is actually styled pseudoselectors

PR #548, opened 2021-07-09
feat(SCT-471): prevent user from adding relationship with same person and type

chubberlisk:
**What**  

Update the relationship types to disable types in the select dropdown that the selected person is already exist.

**Why**  

To prevent duplicated relationships.

**Link to Jira ticket**

https://hackney.atlassian.net/browse/SCT-471

**Follow-up actions**

- Update add relationship page to show person selected
- Remove person from search results
- Show main carer and gender in relationships module
- Fix styling and potentially refactor to use `Formik` (thanks @jhackett1 for the walkthrough!)


jhackett1 (approved):
so the option will still be in the list but greyed out, right? is that clear enough or is it better to simply remove it from the results?

chubberlisk:
> so the option will still be in the list but greyed out, right? is that clear enough or is it better to simply remove it from the results?

Yeah, the option will still be in the list but greyed out. I think I lean more towards disabling because it would be quicker for me to go to "ah I've already added this relationship" if it was there but disabled instead of if it just wasn't there. Of course, I'm not our users so let's keep an eye on it!

PR #549, opened 2021-07-09
feat(SCT-827): printable versions of forms

jhackett1:
social care still use printed versions of forms as an aid when working with clients who are not digitally literate.

this adds a "see printable version" link to an in-progress submission, which leads to a print-friendly formatted version of the form, with everything on one page.

it involves:
- a new `<PrintableForm/>` component
- a new `/printable` page under each submission, which will trigger the browser's print dialog when opened
- a way to hide the header and footer of a page by passing a `noLayout` boolean up to next.js's app component, like how we handle "go back" links
- a new link to get to the printable page from the sidebar of a submission's task list


components/Layout/index.tsx
```diff
@@ -8,9 +8,16 @@ import Footer from './Footer/Footer';
 export interface Props {
   children: React.ReactChild;
   goBackButton?: boolean;
+  noLayout?: boolean;
```
JohnFarrellDev, line 11: Should we just make this a boolean, remove option for undefined type, default the parameter to false.

It looks like we are basically using undefined as false

pages/submissions/[id]/index.tsx
```diff
@@ -83,6 +84,12 @@ const TaskListPage = ({
         </div>
         <div className="govuk-grid-column-one-third">
           <div className={s.sticky}>
+            <Link href={`/submissions/${params.id}/printable`}>
```
Comment (JohnFarrellDev) [inline pages/submissions/[id]/index.tsx:87]: We also want the Link on finished submissions right?

pages/people/{personId}/submissions/{submissionId}

components/PrintableForm/PrintableForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+          {i + 1}. {step.name}
+        </h2>
+        <p>{step.intro}</p>
+        {step.fields.map((field) => (
+          <>
+            <h3>
+              {field.question}
+              {field.required && '*'}
+            </h3>
+            <p>{field.hint}</p>
```
JohnFarrellDev, line 24: Hint could be confusing, think an unrequird field that isn't inserted. The hint looks a bit like the answer on the form.

components/PrintableForm/PrintableForm.tsx
```diff
@@ -0,0 +1,59 @@
+import s from './PrintableForm.module.scss';
+import { Form } from 'data/flexibleForms/forms.types';
+
+interface Props {
+  form: Form;
+}
+
+const PrintableForm = ({ form }: Props): React.ReactElement => (
```
JohnFarrellDev, line 8: We need to include information about the resident in here somewhere

jhackett1: the need isn't for specific forms as far as i can tell, just to print them off and use them as they would paper forms - we're probably okay with just blank fields

components/PrintableForm/PrintableForm.tsx
jhackett1: i don't think i understand! i've tweaked the spacing so that there's less gap between a field's elements

components/Layout/index.tsx
jhackett1: done

pages/submissions/[id]/index.tsx
Comment (jhackett1) [reply pages/submissions/[id]/index.tsx:87]: maybe - i wanna see if we can get by without for now. it seems like they just want the form's questions, not the submission info per se.

Comment (jhackett1) [reply pages/submissions/[id]/index.tsx:87]: so, once a form has been submitted, they'd no longer need to print it off

components/PrintableForm/PrintableForm.tsx
JohnFarrellDev: Ok, as a preliminary version that can be modified to later include resident information seems fine

pages/submissions/[id]/index.tsx
Comment (JohnFarrellDev) [reply pages/submissions/[id]/index.tsx:87]: Right, it seems fine for preliminary purpose of allowing people to print off the form.

From what I understand part of the purpose is the printing is to print out a completed form that can then be sent to someone.

The goal is that a form can be completed on the system as normal and if for some reason some other organisation needs a paper copy the printed form with resident information can all be printed off. No filling out paperwork by hand!

As a start I'm ok with this though as it isn't clear how much functionality is required.

components/PrintableForm/PrintableForm.tsx
JohnFarrellDev: Screenshot from 2021-07-09 15-15-04

I think it is fine now with the margins, it's clear the hint isn't part of the actual answer

JohnFarrellDev (approved):
:+1:

PR #556, opened 2021-07-12
refactor(SCT-471): refactor view relationships to use a table

chubberlisk:
**What**  

- Consolidate the `Relationships` component for displaying personal relationships into one.
- Refactor the `Relationships` component to use a table instead of a table so no big visual difference atm.

Before | After
-- | --
image | image

**Why**  

We want to be able to show more data (e.g. gender and context) for relationships so the summary list component is no longer appropriate.

**Anything else?**

- Splitting this PR to make it easier to review, next PR will add new data.
- Will be looking to refactor the code and tests in the next PR.


components/pages/people/relationships/view/Relationships.tsx
```diff
@@ -67,21 +68,64 @@ const Relationships = ({ id }: Props): React.ReactElement => {
...
+                      )}
+                      <td
+                        data-testid={`related-person-name-${personRowIndex}`}
+                        key={`related-person-name-${personRowIndex}`}
+                        className={`${
+                          relationship.persons.length > 1 &&
+                          personRowIndex !== relationship.persons.length - 1
+                            ? `govuk-table__cell govuk-!-width-three-quarters ${style.noBorder}`
+                            : 'govuk-table__cell govuk-!-width-three-quarters'
+                        }`}
```
chubberlisk, line 114: Any ideas on how to do this better and with pure CSS?

components/pages/people/relationships/view/Relationships.tsx
```diff
@@ -67,21 +68,66 @@ const Relationships = ({ id }: Props): React.ReactElement => {
...
+                        )}
+                        <td
+                          data-testid={`related-person-name-${personRowIndex}`}
+                          key={`related-person-name-${personRowIndex}`}
+                          className={`${
+                            relationship.persons.length > 1 &&
+                            personRowIndex !== relationship.persons.length - 1
+                              ? `govuk-table__cell govuk-!-width-three-quarters ${style.noBorder}`
+                              : 'govuk-table__cell govuk-!-width-three-quarters'
+                          }`}
```
chubberlisk, line 116: Any ideas on how to do this better and with pure CSS? 😅

mapleleafjack: If you want to target the last element only, you can use the  ":last-child" css selector in the rule definition to change the style of the last row only - pretty sure this is fine tho

mapleleafjack (approved):
👍

PR #557, opened 2021-07-12
feat(SCT-471): update add relationship form to show select person name

chubberlisk:
**What**  

Update add relationship form to show select person name.

Before | After
--|--
image | image

**Why**  

So the practitioners don't have to remember the person they've selected to add the relationship with.

**Anything else?**

### DIFF
--- a/components/pages/people/relationships/add/AddRelationshipForm.tsx
+++ b/components/pages/people/relationships/add/AddRelationshipForm.tsx
@@ -2,6 +2,7 @@ import Seo from 'components/Layout/Seo/Seo';
 import PersonView from 'components/PersonView/PersonView';
 import FormWizard from 'components/FormWizard/FormWizard';
 import formSteps from 'data/forms/add-relationship';
+import { useResident } from 'utils/api/residents';
 import { addRelationships, useRelationships } from 'utils/api/relationships';
 import PersonLinkConfirmation from 'components/Steps/PersonLinkConfirmation';
 import RELATIONSHIP_TYPE_OPTIONS from 'data/relationships';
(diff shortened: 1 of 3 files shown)


jhackett1:
looks good and works, but we really need to use the `<PersonSelect/>` common component for this sort of thing

jhackett1 (approved):

data/forms/add-relationship.tsx
```diff
@@ -1,17 +1,30 @@
...
       title: 'Add Relationship',
       components: [
-        <h2
+        <h1
           key="subtitle-details"
           className="govuk-fieldset__legend--l gov-weight-lighter govuk-!-margin-bottom-5"
         >
           Add new relationship
+        </h1>,
+        <h2
```
jhackett1, line 19: this might be better off as `<p>` tags, because they aren't really headings. you can use class names to make things visually look a particular way

components/pages/people/relationships/add/AddRelationshipForm.tsx
```diff
@@ -31,11 +32,20 @@ const AddRelationshipForm: React.FC<{
...
+    useResident(secondPersonId);
 
-  if (!relationships && !error) return <Spinner />;
+  const { data: relationships, error: relationshipsError } =
+    useRelationships(personId);
 
-  if (error || relationships === undefined) return <ErrorMessage />;
+  if (!secondPerson && !secondPersonError) return <Spinner />;
+
+  if (secondPersonError || secondPerson === undefined) return <ErrorMessage />;
```
jhackett1, line 43: is there an error message we can show to the user here? `<ErrorMessage/>` can take a string as an argument

data/forms/add-relationship.tsx
chubberlisk: Ahhhh, good shout - have updated!

components/pages/people/relationships/add/AddRelationshipForm.tsx
chubberlisk: Ah okay, we can update these to have more specific error messages for each hook.

PR #559, opened 2021-07-13
feat(SCT-814): overhaul person view

jhackett1:
the first of a sequence of pull requests for a more mature, scalable and navigable person view: https://hackney.atlassian.net/browse/SCT-814

we did a lot of early user research on different approaches to the person view. this one was the winner

this one:

- splits the person view out into tabs (timeline, allocations, relationships, details)
- redesigns the old records table into a timeline view, which also shows information about in-progress submissions against that person
- populates the other tabs (i.e. everything other than timeline) with the existing components as-is. this looks a little repetitive right now but it's best to keep the size of this pr down

among the ux improvements, we:

- show a summary of a resident's circumstances right up by their name (birthdate, id, allocations)
- show the counts of relationships and allocations right there in the menu
- make it easier and more obvious to "add something new" to a case
- create a secondary action area near the main side nav, for other common actions (eg. "send an email/text message"?)
- synthesise together records/cases and submissions data into a cohesive timeline, to give users a clear idea of where a case is at

in later prs we'll:
- fix up the other tabs to make everything look more visually consistent and tightly integrated
- improve the way the "add something new" button works, with an improved hierarchy of case recording possibilities
- enhance the timeline further to show next steps and more filtering options

Screenshot 2021-07-13 at 11-32-24 Josephine Tester Social care Hackney Council
^ flavour of the new timeline view


components/NewPersonView/Layout.tsx
```diff
@@ -0,0 +1,161 @@
...
+  person: Resident;
+  children: React.ReactChild;
+}
+
+const summariseAllocations = (allocations: Allocation[]): string | null => {
+  if (allocations?.length === 1)
+    return ` · Allocated to ${allocations[0].allocatedWorker}`;
+  if (allocations?.length === 2)
+    return ` · Allocated to ${allocations[0].allocatedWorker} and 1 other`;
+  if (allocations?.length > 1)
```
JohnFarrellDev, line 46: I know the previous if statement would already have returned but I think saying `if (allocations?.length > 2)` would make more sense.

components/NewPersonView/PersonTimeline.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { Case } from 'types';
+import useSearch from 'hooks/useSearch';
+import s from './index.module.scss';
+import FilterButton, { Filter } from './FilterButton';
+import { Submission } from 'data/flexibleForms/forms.types';
+import SearchBox from 'components/SubmissionsTable/SearchBox';
+import UnfinishedSubmissionsEvent from './UnfinishedSubmissions';
+import { normaliseDateToISO } from 'utils/date';
+import Event from './Event';
+import { mockSubmission } from 'factories/submissions';
```
JohnFarrellDev, line 12: never used

components/NewPersonView/PersonTimeline.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const minorForms = ['ASC_case_note', 'Child Case Note'];
```
JohnFarrellDev, line 14: What do we decide is minor vs major?

It would be more sensible to store this information in the root level data folder than in our component code.

components/NewPersonView/UnfinishedSubmissions.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const completedSteps = Object.keys(sub.formAnswers).length;
+  const totalSteps = sub.form?.steps?.length;
+
+  return (
+    <li key={sub.submissionId}>
+      <Link href={`/submissions/${sub.submissionId}`}>
+        {sub?.form?.name || sub.formId}
+      </Link>{' '}
+      <p className="lbh-body-xs">
+        {!isNaN(completedSteps) &&
```
JohnFarrellDev, line 19: It's typically better to use `Number.isNaN(...)` if you actually want to check for NaN value.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

components/NewPersonView/UnfinishedSubmissions.tsx
```diff
@@ -0,0 +1,55 @@
...
+          !isNaN(totalSteps as number) &&
+          `${Math.round(
+            (completedSteps / Number(totalSteps)) * 100
```
JohnFarrellDev, line 22: Would totalSteps ever be undefined, if so Number(undefined) gives us `NaN`,

pages/api/submissions/index.ts
```diff
@@ -1,31 +1,39 @@
...
           socialCareId,
           String(user?.email)
         );
-
         res.json(submission);
       }
       break;
     case 'GET':
-      {
+      if (req.query.includeSubmissions) {
```
JohnFarrellDev, line 25: I think this code is fine, there seems to be some issue on staging currently with getting unfinished submissions, should investigate that before merging this in.

pages/people/[id]/index.tsx
```diff
@@ -1,21 +1,93 @@
-import { useRouter } from 'next/router';
+import { getResident } from 'lib/residents';
+import Layout from 'components/NewPersonView/Layout';
+import { GetServerSideProps } from 'next';
+import { Resident, User } from 'types';
```
Comment (JohnFarrellDev) [inline pages/people/[id]/index.tsx:4]: `User` not being used

pages/people/[id]/index.tsx
```diff
@@ -1,21 +1,93 @@
...
+import { useCasesByResident } from 'utils/api/cases';
+import { Case } from 'types';
+import PersonTimeline from 'components/NewPersonView/PersonTimeline';
+import Spinner from 'components/Spinner/Spinner';
+import ErrorMessage from 'components/ErrorMessage/ErrorMessage';
+import { useUnfinishedSubmissions } from 'utils/api/submissions';
+import { canManageCases } from 'lib/permissions';
+import { useAuth } from 'components/UserContext/UserContext';
```
Comment (JohnFarrellDev) [inline pages/people/[id]/index.tsx:12]: not being used

pages/people/[id]/index.tsx
```diff
@@ -1,21 +1,93 @@
...
 };
 
 PersonPage.goBackButton = true;
 
+export const getServerSideProps: GetServerSideProps = async ({
+  params,
+  req,
+}) => {
+  const person = await getResident(Number(params?.id));
+  const user = await isAuthorised(req);
```
Comment (JohnFarrellDev) [inline pages/people/[id]/index.tsx:66]: `isAuthorised` doesn't return a promise, we don't need to await right?

JohnFarrellDev (approved):
I think this is really good, just minor suggestions and let's see what is going on with not being able to get the unfinished submissions on staging

pages/people/[id]/index.tsx
```diff
@@ -1,21 +1,93 @@
...
+          />
+        ) : (
+          <p>No events to show</p>
+        )
+      ) : casesError ? (
+        <ErrorMessage label={casesError.message} />
+      ) : (
+        <Spinner />
+      )}
+    </Layout>
```
Comment (chubberlisk) [inline pages/people/[id]/index.tsx:55]: What do you think about feature flagging this? So if it's on, use this improved person view and if it's not, use the old `PersonDetailsPage`. This is quite a big visual change so I wonder if we'll need approval before it goes out but tbh I think it's always good to put a feature flag just in case.

utils/api/submissions.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { useUnfinishedSubmissions } from './submissions';
+import * as SWR from 'swr';
+
+jest.mock('swr');
+
+describe('residents APIs', () => {
+  describe('useResident', () => {
```
chubberlisk, line 7:
```suggestion
  describe('useUnfinishedSubmissions', () => {
```

utils/api/submissions.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+    it('should work properly', () => {
```
chubberlisk, line 8: Looks like we're also doing some filtering on the response we get back. Could we add some unit tests around this?

utils/date.ts
```diff
@@ -63,3 +63,17 @@ export const objectDateToString = (
...
+    const [date, time] = str.split(' ');
+    const [days, month, year] = date.split('/');
+    const [hours, minutes, seconds] = time.split(':');
+    return `${year}-${month}-${days}T${hours.padStart(2, '0')}:${minutes}${
+      seconds ? `:${seconds}` : ''
+    }`;
+  } else {
+    return str;
+  }
+};
```
chubberlisk, line 79: Could something like Moment.js help us here? https://momentjs.com/docs/#/parsing/special-formats/

chubberlisk (commented):
Had a quick looksies as I know John has reviewed already. Tested locally and the new design is a nice breakdown of information! 💯 Some comments below.

components/NewPersonView/PersonTimeline.tsx
jhackett1: yes, we should do that, but at the moment i don't have a good sense of what's major and minor - it will probably be a longer list than just these two.

pages/api/submissions/index.ts
jhackett1: was to do with api disruption - working fine now

utils/date.ts
jhackett1: we already have a date library in the app - moment is a very large library so would be best to avoid adding a second one. i think the ideal here is to fix the apis to always expose a standard iso format so we don't need this logic at all!

PR #560, opened 2021-07-13
feat(SCT-471): add main carer, gender and details to view relationships

chubberlisk:
**What**  

Add if a related person is a main carer and display their gender and any details added to the relationship.

Before | After
--|--
image | image

**Why**  

We need to display what we have as fields in the add relationship form.

**Anything else?**


mapleleafjack (approved):
👫

PR #563, opened 2021-07-13
SCT-828 - Public ENV variable

mapleleafjack:
**What**  
Added the NEXT_PUBLIC_ENV variable in serverless.yml with value ${opt:stage}.
The final value should be "Staging" or "Production" based on the stage.

**Why**  
This is necessary for [this PR](https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/562) so Cypress can read the new variable and pass the tests for the new feature/fix, which is currently failing as Staging doesn't have this variable set


chubberlisk (approved):

mapleleafjack:
Documentation around this parameter has been added in the linked PR (where the parameter is effectively used)

PR #565, opened 2021-07-14
Feat: Finalise cfs case note form

JohnFarrellDev:
**What**  
Finalise some changes for CFS case note form.

Add title field.
Move positions of tags.
Make repeater group minimized by default.

**Why**  

Based on feedback from practitioner group.

**Anything else?**


components/AddForm/AddForm.spec.tsx
```diff
@@ -39,7 +39,7 @@ const flexibleForms: [string, boolean, boolean][] = [
   ['FACE overview assessment', false, false],
   ['Safeguarding Adult Concern Form', false, false],
   ['Safeguarding adult manager decision on concern', false, false],
-  ['Child Case Note', false, false],
+  ['Case note', false, false],
```
jhackett1, line 42: we're replacing this component with `<AddFormDialog/>` shortly. should adjust this if that gets merged first

JohnFarrellDev:
Add virtual visit for kind of note

Add conditional field for if child/children seen, if child/children seen alone

components/FlexibleForms/RepeaterGroupField.tsx
```diff
@@ -35,6 +38,8 @@ const RepeaterGroupField = ({
     touched: FormikTouched<FormikValues>;
   } = useFormikContext();
 
+  const [hidden, setHidden] = useState(hiddenOnStart ?? false);
```
jhackett1, line 41: this seems to do the trick! though i wonder if we could do it in a stateless way by modifying line 82 on the `lib/utils/generateInitialValues.ts` file - the `getInitialValues` function. we could do something like:

```
if(field.initiallyBlank){
initialValues[field.id] = []
} else {
initialValues[field.id] = generateInitialValues(field.subfields || [], person),
}

jhackett1 (approved):
looks fine - left a suggestion on a possible cleaner implementation, and let's add "virtual visit" as a visit subtype

PR #568, opened 2021-07-14
chore: update person details

jhackett1:
make the `<PersonDetails/>` component visually consistent with the new person view

it also:
- uses sentence case on all form labels
- updates terminology to be consistent with the rest of the app (eg. the neutral "social care id" rather than "person id" or "mosaic id")
- splits nhs number off into its own panel, ready for the new health and medical fields we will soon be adding
- removes the `<PersonDetails/>` component from the `<FormWizard/>` and instead uses the new `<PersonWidget/>` to be consistent with our new forms

Capture


JohnFarrellDev (approved):
:+1:

PR #569, opened 2021-07-14
Fix for "add relationship" link after page design change

mapleleafjack:
**What**  
Fixed the "add a new relationship" button link so it works again

**Why**  
After the page redesign, the link was not working anymore


chubberlisk (approved):
🕵️

jhackett1:
we need to do the same to the add allocations button, oops

PR #571, opened 2021-07-14
feat(SCT-823): no task list if single step form

JohnFarrellDev:
**What**  
Do not show the task list for forms with a single step

**Why**  
it is confusing to see a task list with a single step

**Anything else?**

---

the implementation involves:

- adding a conditional redirect to the step page if a form has only a single step
- conditionally calling a `finishForm` function if the submit button is pressed, the form is valid, and the form has only one step
- bug fixes and improved error handling to our back-end api and `finishSubmissions` function

this appears to be functional and sufficiently fast to launch, but it would probably benefit from refactoring. the code of the submissions pages and `<StepForm/>` is getting quite spaghetti.


jhackett1:
> having a single step thing is harder than it looks. will probably take the rest of the day
2:20
skipping the task list is fine - the tough bit is making the button at the end of the form:
save changes (answers and group recording)
clean up any outstanding autosave so as not to block route changes
mark the submission finished
redirect to the right place
2:21
i might use it as an opportunity to refactor the way the submission pages operate
2:21
will report back

jhackett1:
after trying several different approaches, i'm now convinced that the cleanest approach is _not_ to use the task builder. at least not the same way we've used it for other forms.

it's too much effort and weird code to  avoid the task list element - instead, i've made a new route at `/case-note` which uses the _primitives_ of the form builder directly. it seems a much cleaner approach, at least for the tiny number of forms that have only a single step.

jhackett1 (approved):

PR #573, opened 2021-07-14
chore(sct-814): fix allocations layout

jhackett1:
a follow-up pr to make allocations visually consistent with the new person view

that's pretty much it!

it:
- changes the ui for allocating someone to a simple button, rather than a whole empty panel, which is less confusing
- simplifies the text by making the name of the worker the heading of the panel
- use the same styling as personal details

Screenshot 2021-07-14 at 18 02 30


chubberlisk (approved):
💯

PR #575, opened 2021-07-15
feat(SCT-471): update relationship type options to match when viewing

chubberlisk:
**What**  

Update relationship type options to match when viewing.

**Why**  

This ensures that the dropdown options for relationship types when adding a relationship match
how they are written when viewing relationships but singular. We do this by refactoring
and having a single mapping list.

**Anything else?**


jhackett1 (approved):

jhackett1:
impeccable use of sentence case!

PR #578, opened 2021-07-15
chore(sct-858): fix error when there are no events on the timeline

jhackett1:
fixes: https://hackney.atlassian.net/browse/SCT-858

also improves the visual appearence of the timeline when there are more events to load

and adds a new test case to make sure the problem can't reccur.


Miles-Alford (approved):
👍

PR #581, opened 2021-07-15
chore(sct-823): clean up and fix child case note

jhackett1:
https://hackit-lbh.slack.com/archives/C021Q4SDAV8/p1626365692369000

it solves these small issues:

- it is a little odd that the title and “what happened?” fields are the same size.
- how did we select the four items that get radio buttons? visits and correspondence make sense. not sure i understand why management oversight is included. help me out?
- if something is in the set of radio buttons can we remove it from the topics list?
- can we limit the items in “topics” to just the 22 that are in the google form “types” list?
- do we need some sort of hint text in the topics module about being able to type in there and being able to create your own custom titles? not sure how discoverable those are, and they are both really really useful features.
- i keep getting this error where i finish the note, hit “finish and send” and the get a “There was a problem finishing the submission” red warning box. if i hit the “go back button” it takes me to the completed form output page.
- is it possible to clean up the output view so that we remove any fields that were empty? especially the type fields at the top.

as well as:

- adding a `default` attribute for string-type fields, so we can have date fields default to today's date, for instance
- cleaning up the appearance and usability of tag and combobox fields
- has tag fields accept a `choices` prop, like other similar fields, rather than having a single hardcoded list
- improves `<FlexibleAnswers/>` to hide falsy fields and empty array fields
- removes code for the older, superseded child case note implementation


rs2416 (approved):
I can’t properly review it since I am having that permissions on the local environment - but looking at the code I’m happy.

PR #582, opened 2021-07-16
chore: fix the go back button behaviour

jhackett1:
>i think the “go back” button needs some reconsideration soon. it is just a browser back button right now, but seems like it should behave a little differently (more like a breadcrumb)

> the back button works for me maybe one third of the time

>i know the “go back” pattern is a gds classic but it is more than a little problematic for a case management system

>we can improve the behaviour by hardcoding the link to go back to

remaining problem: how can we get dynamic information at the **top level** of a page component file?

eg. at the moment we say whether a page should show the back button by putting in `PersonPage.goBackButton = true`. how can we instead include specific information like the id of a person to go back to?

### DIFF
--- a/components/Layout/BackButton/BackButton.spec.tsx
+++ b/components/Layout/BackButton/BackButton.spec.tsx
@@ -17,4 +17,11 @@ describe('BackButton', () => {
     fireEvent.click(button);
     expect(history.back).toHaveBeenCalled();
   });
+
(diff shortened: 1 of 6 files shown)


chrishutchinson:
Closing follow [this discussion on Slack](https://hackit-lbh.slack.com/archives/C021Q88C953/p1633012276265200). The branch will stay, so we can restore it later if needed.

PR #584, opened 2021-07-16
fix(SCT-471): radio button styling when selecting related person

chubberlisk:
**What**  

Fix radio button styling when selecting a person for a relationship.

Before | After
--|--
image | image

**Why**  

So it matches visually with other radio buttons.

**Anything else?**


mapleleafjack (approved):
🔘

components/Search/results/RelationshipTable.tsx
```diff
@@ -14,15 +14,23 @@ const ResultEntry = ({
...
+              id={mosaicId}
+              name="personId"
+              type="radio"
+              value={mosaicId}
+              onChange={(e) => {
+                callback(e.target.value);
+              }}
+              className={`govuk-radios__input`}
+            />
+            <label className="govuk-label govuk-radios__label">
```
jhackett1, line 28: needs a `htmlFor={mosaicId}` attribute to be valid html

jhackett1 (changes requested):
one minor a11y/semantic html fix

components/Search/results/RelationshipTable.tsx
chubberlisk: Ahhhhh, thanks for the spot!

jhackett1 (approved):

PR #586, opened 2021-07-16
refactor(SCT-481): move factory building into test file

chubberlisk:
**What**  

Refactor `components/pages/people/relationships/view/Relationships.spec.tsx` to move factory building into the test file and renamed test descriptions.

**Why**

So that the set up required for a test and what the test is doing is clearer.

**Anything else?**


mapleleafjack (approved):
Much cleaner!

🚀

PR #591, opened 2021-07-19
Delete relationships UI fixes

mapleleafjack:
**What**  
Changed the Relationship UI to use UseState react hook to update data visualisation,
tidied up the code

**Why**  
UseState lets us re-render the module without refreshing the page

**Anything else?**

We now need to update the relationship count in the left hand side


chubberlisk (approved):
👍 as discussed we want to improve how we update the relationships shown when removed but we get things working first and then iterate!

PR #592, opened 2021-07-19
feat(SCT-519): activate child case notes

jhackett1:
this makes child case notes visible to social workers, and removes the equivalent google form.

it also improves the group recording widget so that someone can now search by name or id in the same box. as a temporary workaround, it does this by making three seperate requests to the api, respectively searching for matches on:
- id
- first name
- last name

all the results are then concatenated and deduplicated.

along with replacing tags with checkboxes

also make some minor visual and ux adjustments to the timeline


components/GroupRecording/GroupRecordingWidget.tsx
```diff
@@ -58,8 +80,8 @@ const GroupRecordingWidget = ({
           .includes(parseInt(String(person.mosaicId)));
       }
     })
-    // only show six max
-    .slice(0, 5);
+    // // only show six max
```
JohnFarrellDev, line 83:
```suggestion
    // only show nine max
```

components/NewPersonView/Layout.tsx
```diff
@@ -84,16 +84,18 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
...
-      text: 'Add case note',
-      href: `/people/${person.id}/case-note`,
-    },
     {
       text: 'Add warning note',
       href: `/people/${person.id}/warning-notes/add?id=${person.id}`,
     },
   ];
 
+  if (person.contextFlag === 'C')
```
JohnFarrellDev, line 93: I really dislike that we determine in the component if a link to a form is displayed or not compared to our boolean flags.

As we saw before multiple forms got deployed when they shouldn't have been and I had to rewrite that section of code to tidy it up and make it manageable.

JohnFarrellDev: There is also no test to determine if this can be seen or not depending on groups the user belongs to!

components/NewPersonView/Layout.tsx
```diff
@@ -90,6 +90,12 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
...
+    secondaryNavigation.unshift({
```
JohnFarrellDev, line 94: code is fine but lacks tests

jhackett1: added

JohnFarrellDev (approved):

PR #593, opened 2021-07-19
fix: don't prematurely launch cfs case notes

jhackett1:


components/NewPersonView/Layout.tsx
```diff
@@ -84,16 +84,18 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
...
-      text: 'Add case note',
-      href: `/people/${person.id}/case-note`,
-    },
     {
       text: 'Add warning note',
       href: `/people/${person.id}/warning-notes/add?id=${person.id}`,
     },
   ];
 
+  if (person.contextFlag === 'A')
```
chubberlisk, line 93: Could we use a feature flag instead please? https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/main/.github/CONTRIBUTING.md#react-hook

jhackett1: this isn't supposed to be a feature flag - we _want_ to show that link in all conditions on an adult person

JohnFarrellDev: As we saw last time with the `AddForm` component (I believe) this will quickly get too messy and either need to be rewritten or forms will accidentally be released.

Also it is a mess to debug, we need to be using the flags on the forms `isViewableByAdults` and `isViewableByChildrens`

chubberlisk: Ahhhh sorry @jhackett1, I get it now! 🤦

jhackett1: @JohnFarrellDev this isn't one of our flexible forms - this is the legacy form that we're taking out. there's no flags to check for

JohnFarrellDev (approved):

PR #596, opened 2021-07-20
feat(SCT-900): pass the auditing params with the GET case note by ID API request

chrishutchinson:
**What**  
We want to add our new auditing process to the "GET case note by ID" request. This does just that!

**Why**  
[SCT-900](https://hackney.atlassian.net/browse/SCT-900)


mapleleafjack (approved):
✅

PR #599, opened 2021-07-20
chore(SCT-519): overhaul asc case notes

jhackett1:
redo the asc case notes in the new form builder, remove old implementation.

it does this by:
- using the lead/first resident's service context to load either the adult or child case note forms in `/pages/people/[id]/case-note`
- adding a stripped down form config for `adult-case-note`, which combines the current live version and the approved childrens' version.
- simplifies the "add case note" links in the side nav to use the one single link for everyone.
- strips out some old code


components/NewPersonView/Layout.spec.tsx
```diff
@@ -115,33 +115,4 @@ describe('Layout', () => {
     );
     expect(screen.queryByText('Timeline')).toBe(null);
   });
-
-  it('shows the correct case note link for a child resident', () => {
```
jhackett1, line 119: these are now irrelevant because we no longer have any conditional logic here

data/flexibleForms/adultCaseNote.ts
```diff
@@ -0,0 +1,189 @@
...
+        {
+          id: 'Topics',
+          question: 'Topics',
+          hint: 'Help colleagues find this note. Add as many as you need.',
+          type: 'checkboxes',
+          condition: {
+            id: 'Type',
+            value: 'Something else',
+          },
+          choices: [
```
jhackett1, line 95: taken from current live list, with duplicates from radios above removed.

data/flexibleForms/adultCaseNote.ts
```diff
@@ -0,0 +1,189 @@
+import { Form } from './forms.types';
+import { format } from 'date-fns';
+
+const form: Form = {
+  id: 'adult-case-note',
+  name: 'Case note',
+  groupRecordable: true,
+  isViewableByAdults: true,
```
JohnFarrellDev, line 8: Just to confirm, this has been approved to go out into prod?

data/flexibleForms/adultCaseNote.ts
JohnFarrellDev: I don't think it will be possible to query by these with the current setup. I do have some thoughts around tagging/topics and how we can do it though.

pages/people/[id]/case-note.tsx
```diff
@@ -151,7 +155,8 @@ export const getServerSideProps: GetServerSideProps = async ({
   params,
   query,
 }) => {
-  let submission;
+  let submission: Submission;
+
   if (query.submissionId) {
     submission = await getSubmissionById(String(query.submissionId));
```
Comment (JohnFarrellDev) [inline pages/people/[id]/case-note.tsx:161]: we should just assert that query.submissionId is a string, we don't need to actually call the String constructor which is adding real runtime code.

`query.submissionId as string`

JohnFarrellDev (approved):
LGTM, only minor comments and making sure we got deployment approval

PR #600, opened 2021-07-20
Security Configuration: Disable CSP

Miles-Alford:
**What**  
This PR primarily removes the `Content-Security-Policy` security header from the application. Furthermore the baseline scan config has now been amended by locking in all the currently succeeding rules to fail if they are broken in the future.

**Why**  
The reason for this change is that the default CSP is too strict, it blocks important third party sites which are used within the application such as google fonts and other's such as Hotjar. The reason for the added security rules to the baseline config is that it makes it easier to debug the problem if these rules begin to fail in the future.

The two security headers still in the next config should still protect against the original security issues brought up from the penetration test.

**Anything else?**
The CSP should be re-added once the rules for it has been agreed on by Hackney.


neilkidd (approved):
Fantastic description!

PR #605, opened 2021-07-21
feat: datetime field

jhackett1:
this adds support for a combined date and time field to the flexible form builder.

it does this by:
- adding a `<DateTimeField/>` component, and rendering it in `<FlexibleField/>`
- augmenting the `generateFlexibleSchema` and `generateInitialValues` functions to support the new field
- supplementing the `default` attribute on a field to support an array of strings, so a default date and time can be specified like:

```
        {
          id: 'Date of event',
          question: 'When did this happen?',
          type: 'datetime',
          required: true,
          default: [
            format(new Date(), 'yyyy-MM-dd'),
            format(new Date(), 'HH:00'),
          ],
        },
```

Screenshot 2021-07-21 at 15 02 52

### DIFF
--- a/components/FlexibleForms/DateTimeField.spec.tsx
+++ b/components/FlexibleForms/DateTimeField.spec.tsx
@@ -1,0 +1,92 @@
+import DateTimeField from './DateTimeField';
+import { Formik, Form } from 'formik';
+import { render, screen } from '@testing-library/react';
+
+const mockSubmit = jest.fn();
+
+describe('TextField', () => {
+  it('renders two text inputs with the right types', () => {
+    render(
+      <Formik
+        onSubmit={mockSubmit}
+        initialValues={{
+          foo: false,
+        }}
+      >
+        {({ touched, errors }) => (
+          <Form>
+            <DateTimeField
+              touched={touched}
+              errors={errors}
+              name="foo"
+              label="Label text"
+              hint="Hint text"
+            />
+          </Form>
+        )}
+      </Formik>
+    );
+
+    expect(screen.getByText('Label text'));
+    expect(screen.getByText('Hint text'));
+    expect((screen.getByLabelText('Date') as HTMLInputElement).type).toBe(
+      'date'
+    );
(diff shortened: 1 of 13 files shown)


components/FlexibleForms/DateTimeField.spec.tsx
```diff
@@ -0,0 +1,92 @@
+import DateTimeField from './DateTimeField';
+import { Formik, Form } from 'formik';
+import { render, screen } from '@testing-library/react';
+
+const mockSubmit = jest.fn();
+
+describe('TextField', () => {
```
JohnFarrellDev, line 7: DateTimeField

components/FlexibleForms/DateTimeField.tsx
```diff
@@ -0,0 +1,88 @@
...
+      className={`govuk-input lbh-input ${s.date}`}
+      aria-describedby={hint && `${name}-hint`}
+      type="date"
+    />
+
+    <label htmlFor={`${name}-time`} className="govuk-visually-hidden">
+      Time
+    </label>
+    <RawField
+      data-testId="datetime-box"
```
JohnFarrellDev, line 78: Do we ever use these data-testIds? I don't think so.

data/flexibleForms/forms.types.ts
```diff
@@ -27,8 +28,8 @@ export interface Field {
   className?: string;
   /** on conditional fields, required value is only respected when all conditions are met */
   required?: boolean;
-  /** give an initial, default value for string-type fields */
-  default?: string;
+  /** give an initial, default value for datetime and string-type fields */
+  default?: string | string[];
```
JohnFarrellDev, line 32: Am i right in thinking the datetime would always be 2 strings, we could keep our types more restrictive using a tuple [string, string]

jhackett1: makes sense

JohnFarrellDev (approved):

PR #606, opened 2021-07-21
feat(SCT-471): add headings to relationships table

chubberlisk:
**What**  

- Add headings to relationships table.
- Fix bug where if you selected a relationship type with a single checkbox option e.g. Aunt / uncle, friend and selected main carer, it didn't send it as part of the request so main carer never showed up when you went to view the relationship

Before | After
--|--
image|image

**Why**  

This was part of feedback received about relationships.

**Anything else?**


mapleleafjack (approved):
[🚀]

PR #614, opened 2021-07-22
feat(SCT-918): timeline improvements

jhackett1:
small improvements to the timeline:
- show a title in all cases, rather than defaulting to "view"
- for **OLDER CASE NOTES ONLY**, show the title or description, if it's available (doing this for newer submissions needs api changes)
- for **OLDER CASE NOTES ONLY**, search case note content as well as name and author
- handle the datetime format returned by google docs
- show a more prominent banner if a person is restricted

Screenshot 2021-07-22 at 11 42 57

Screenshot 2021-07-22 at 13 05 30

### DIFF
--- a/components/Cases/CaseLink.tsx
+++ b/components/Cases/CaseLink.tsx
@@ -21,10 +21,10 @@ const getWarningNoteDetailsPageName = (form_name: string) => {
   }
 };
 
-const getLink = (
+export const getLink = (
   recordId: string,
   { form_name_overall, ...caseFormData }: CaseFormData
-) => {
+): string | null => {
   const fileName = (form as Record<string, unknown>)[form_name_overall]
     ? form_name_overall
     : null;
--- a/components/Cases/CaseLink.tsx
+++ b/components/Cases/CaseLink.tsx
@@ -1,56 +1,7 @@
 import Link from 'next/link';
 import forms from 'data/flexibleForms';
-import * as form from 'data/forms';
-import {
-  AllocationCaseFormData,
-  CaseFormData,
-  DeallocationCaseFormData,
-  WarningNoteCaseFormData,
-} from 'types';
-
-const getWarningNoteDetailsPageName = (form_name: string) => {
-  switch (form_name) {
-    case 'Warning Note Created':
-      return 'note-created';
-    case 'Warning Note Reviewed':
-      return 'note-reviews';
-    case 'Warning Note Ended':
-      return 'note-end';
-    default:
-      return null;
-  }
-};
-
-export const getLink = (
-  recordId: string,
-  { form_name_overall, ...caseFormData }: CaseFormData
-): string | null => {
-  const fileName = (form as Record<string, unknown>)[form_name_overall]
-    ? form_name_overall
-    : null;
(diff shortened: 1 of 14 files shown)


components/Cases/CaseLink.tsx
```diff
@@ -21,10 +21,10 @@ const getWarningNoteDetailsPageName = (form_name: string) => {
   }
 };
 
-const getLink = (
+export const getLink = (
```
JohnFarrellDev, line 24: this function should be unit tested

components/NewPersonView/Event.tsx
```diff
@@ -24,7 +26,9 @@ const Event = ({ event }: Props): React.ReactElement => {
     String(event?.dateOfEvent || event?.caseFormTimestamp)
   );
 
-  const displayName = event?.formName || event?.caseFormData?.form_name_overall;
+  const displaySnippet = event?.caseFormData?.case_note_title
+    ? `${event?.caseFormData?.case_note_title}: ${event?.caseFormData?.case_note_description}`
+    : false;
```
JohnFarrellDev, line 31: Is this tested?

components/NewPersonView/Event.tsx
```diff
@@ -33,21 +37,20 @@ const Event = ({ event }: Props): React.ReactElement => {
...
       <h3 className="lbh-heading-h3">
-        <CaseLink
-          formName={event.formName}
-          externalUrl={event.caseFormUrl}
-          caseFormData={event.caseFormData}
-          recordId={event.recordId}
-          personId={event.personId}
-        >
-          {displayName}
-        </CaseLink>
```
JohnFarrellDev, line 44: Do we even need the CaseLink component now, it only exists in the <CaseTable> component which doesn't seem to be used anywhere?

components/NewPersonView/Layout.spec.tsx
```diff
@@ -115,4 +119,20 @@ describe('Layout', () => {
     );
     expect(screen.queryByText('Timeline')).toBe(null);
   });
+
+  it('tells unauthorised users when a resident is restricted', () => {
```
JohnFarrellDev, line 123: Same test but for when canManageCases(user, person) returns true.

utils/date.ts
```diff
@@ -73,6 +73,9 @@ export const normaliseDateToISO = (str: string): string => {
     return `${year}-${month}-${days}T${hours.padStart(2, '0')}:${minutes}${
       seconds ? `:${seconds}` : ''
     }`;
+  } else if (/[0-9]{1,2}\/[0-9]{1,2}\/[0-9]{4}/.test(str)) {
```
JohnFarrellDev, line 76: Probably want a test for this too

JohnFarrellDev (approved):
I reckon the code is fine but would like to see some more tests (who knows maybe writing the tests will expose some minor oversights/bugs too)

components/NewPersonView/Event.tsx
jhackett1: it's still used in a few places - i am keen to deprecate it

PR #619, opened 2021-07-23
feat: refactor submission components

jhackett1:
iterate on the `<SubmissionPanel/>` component and its use on the dashboard:
- improved content hierarchy
- it now shows a bar across the bottom indicating completion percentage
- it is more obviously actionably (big "resume" button)
- uses a "details" dialog that can be used in more places in the app
- remove "submissions" language in favor of "work in progress" which seems to perform better in research
- streamline the ui on the "work in progress" page to remove the radio buttons - instead using a separate "my work" page

idea is to re-use these components on the person view overview tab too.

Screenshot 2021-07-23 at 10-38-12 Dashboard Social care Hackney Council


anabebs (approved):
Looks good!

PR #624, opened 2021-07-26
feature(sct-772): media

jhackett1:
mvp frontend for media

this includes:
- a new page at `/people/:id/media` to show all media stored for a particular resident
- new stub methods in `/lib/media.ts` for retrieving media for a resident or by id
- new components in `/components/Media` for displaying different kinds of media

it's behind a feature flag called `media`, so it's **safe to launch** in its current state

Screenshot 2021-07-27 at 10 56 35
Screenshot 2021-07-27 at 10 56 25

### DIFF
--- a/components/Media/MediaDetailDialog.tsx
+++ b/components/Media/MediaDetailDialog.tsx
@@ -1,5 +1,4 @@
 import Dialog from 'components/Dialog/Dialog';
-import s from './index.module.scss';
 import { Media } from 'types';
 import { format } from 'date-fns';
 
--- a/components/Media/MediaDetailDialog.tsx
+++ b/components/Media/MediaDetailDialog.tsx
@@ -2,22 +2,13 @@ import Dialog from 'components/Dialog/Dialog';
 import { Media } from 'types';
 import { format } from 'date-fns';
 import s from './index.module.scss';
+import { humanFileSize } from 'utils/media';
 
 interface Props {
   media: Media;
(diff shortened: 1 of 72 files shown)


components/Media/MediaDetailDialog.tsx
```diff
@@ -0,0 +1,58 @@
...
+import s from './index.module.scss';
+
+interface Props {
+  media: Media;
+  isOpen: boolean;
+  onDismiss: () => void;
+}
+
+const humanFileSize = (bytes: number): string => {
+  if (bytes == 0) {
```
JohnFarrellDev, line 13: Probably want to use triple equality `===`

components/Media/MediaDetailDialog.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return '0.0 B';
+  }
+  const e = Math.floor(Math.log(bytes) / Math.log(1024));
```
JohnFarrellDev, line 16: This seems overly complex and I think has some unexpected-ish behaviour, I'd expect `1024000` to return 1MB not 1000.0 KB for example.

I would advise moving this into a seperate utility module and testing the code.

We also don't handle negatives.

chrishutchinson:
Closing follow [this discussion on Slack](https://hackit-lbh.slack.com/archives/C021Q88C953/p1633012276265200). The branch will stay, so we can restore it later if needed.

PR #625, opened 2021-07-26
fix(sct-857): auto save stuck on hidden conditionals

jhackett1:
https://hackney.atlassian.net/browse/SCT-857

solves the problem by:

- removing the separate `validateConditionalFields` and combine the logic into our main `generateFlexibleSchema` function
- extract `applyRequired` into its own function to stop duplication
- make every form have `conditions: Condition[]` rather than allowing some to be a plain object
- add some new tests to specifically cover the buggy behaviour

**the only interesting stuff here is in `lib/validators.ts`**


JohnFarrellDev (approved):
Works as expected :+1:

PR #626, opened 2021-07-26
fix: add form data to submissions earlier on, so we can search on it

jhackett1:
it's ideal to be able to type in the name of a form on the dashboard and see the submissions that match it.

> the filter field on the submissions page seems to work strangely. for instance, if i type “review” it filters out the review 3c forms (meaning, they no longer show) in “just mine” but i see other people’s in “everyone”

this wasn't possible before because we were just querying what came back from the api, which had only the machine-readable `formId`.

- we now use a memoised function to add the form to each submission, before applying searching logic
- we modify components lower down the tree to use the form that already exists, rather than re-searching for it

### DIFF
--- a/components/SubmissionsTable/index.spec.tsx
+++ b/components/SubmissionsTable/index.spec.tsx
@@ -1,8 +1,9 @@
-import { fireEvent, render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { mockSubmission } from 'factories/submissions';
 import SubmissionsTable from './index';
 import { useAuth } from 'components/UserContext/UserContext';
 import { mockedResident } from 'factories/residents';
+import { Form } from 'data/flexibleForms/forms.types';
 
 jest.mock('components/UserContext/UserContext');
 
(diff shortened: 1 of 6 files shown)


components/SubmissionsTable/index.spec.tsx
```diff
@@ -42,6 +43,16 @@ describe('SubmissionsTable', () => {
     expect(screen.getByText('No results to show.'));
   });
 
+  it('lets you search for a submission by form name', async () => {
+    render(<SubmissionsTable submissions={[mockSubmission]} />);
+
+    fireEvent.change(screen.getByRole('searchbox'), {
+      target: { value: 'Sandbox form' },
+    });
```
Renny4Real, line 51: Just to check I understand this: when we mock a submission in our tests, this submission is related to the form with an id of `foo` saved [here](https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/f8cb4b2db9a976c96307336b1dfddba1297c5cb4/data/flexibleForms/foo.ts) and that form has the name: `Sandbox form`, which is why we search for it in this test?

components/SubmissionsTable/index.tsx
```diff
@@ -22,20 +23,28 @@ export const SubmissionsTable = ({
...
-          !user?.hasUnrestrictedPermissions &&
-          submission.residents.every((resident) => resident.restricted === 'Y')
-        )
-          return false;
+      submissions
+        // augment each one with its form
+        .map((submission) => ({
+          ...submission,
+          form: forms.find((form) => form.id === submission.formId),
+        }))
```
Renny4Real, line 31: Is this somehow pairing a submission with its original form?

Renny4Real (approved):
Looks good 🎉 . 
Just a couple of questions

components/SubmissionsTable/index.tsx
jhackett1: yes - the forms are json files stored in the git repo. the submission stores a `formId` which we can relate back to a particular form, particularly the human-readable name of the form

components/SubmissionsTable/index.spec.tsx
jhackett1: exactly. were we to ever change that form this test would break

PR #627, opened 2021-07-27
fix: canonical submission urls on the dashboard

jhackett1:
not all submissions follow the standard `/submissions/:id` url format any more. case notes, in particular, don't.

at the moment, it's possible to go to a task list version of the case note from the dashboard, which shouldn't be possible, because of this.

this pr adds a `generateSubmissionUrl` method that can handle those exceptions.


JohnFarrellDev (approved):
:+1:

PR #629, opened 2021-07-27
fix(SCT-946): reload after approval, panel approval or reject

jhackett1:
https://hackney.atlassian.net/browse/SCT-946

improves user experience by refreshing the page after a submission changes state, making sure that the ui reflects the latest state (eg. approved, in progress, submitted)


JohnFarrellDev (approved):
Going to approve because it solves the issue and is simple enough. I do think in general SPAs shouldn't ever need a page reload to create the desired UI change.

PR #633, opened 2021-07-29
feat(SCT-961): record date of event from asc form

JohnFarrellDev:
**What**  
Allows our flexible forms to identify a field as being the `dateOfEvent`, when patching our answers we then can supply the datetime value from the form.

**Why**  
For our ASC case note form the date of event is currently just when the form was created but the event happened sometime in the past. This then messes up the timeline for the resident for example where the form appears in the timeline when it was created, not when the event occurred.

**Anything else?**

### DIFF
--- a/data/flexibleForms/forms.types.ts
+++ b/data/flexibleForms/forms.types.ts
@@ -69,6 +69,9 @@ export interface Form {
   tags?: string[];
   /** override the automatically generated /submissions/new... url, for edge case forms */
   canonicalUrl?: (socialCareId: number) => string;
+  dateOfEvent?: {
+    associatedId: string;
+  };
 }
 
 export interface RepeaterGroupAnswer {
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,5 @@
+process.env.TZ = 'UTC';
+
 module.exports = {
   collectCoverageFrom: [
     'components/**/*.{js,jsx,ts,tsx}',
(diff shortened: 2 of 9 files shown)


pages/submissions/[id]/steps/[stepId].tsx
```diff
@@ -50,17 +47,6 @@ const StepPage = ({
...
-    setStatus: (message: string) => void
-  ): Promise<void> => {
-    try {
-      await axios.post(`/api/submissions/${params.id}`);
-      router.push(`/people/${residents[0].id}/submissions/${params.id}`);
-    } catch (e) {
-      setStatus(e.toString());
-    }
-  };
-
```
Comment (JohnFarrellDev) [inline pages/submissions/[id]/steps/[stepId].tsx:63]: unused function

Comment (jhackett1) [reply pages/submissions/[id]/steps/[stepId].tsx:63]: this is from when we were thinkng about doing single-step forms on the same file. now no longer needed!

data/flexibleForms/forms.types.ts
```diff
@@ -69,6 +69,9 @@ export interface Form {
   tags?: string[];
   /** override the automatically generated /submissions/new... url, for edge case forms */
   canonicalUrl?: (socialCareId: number) => string;
+  dateOfEvent?: {
+    associatedId: string;
```
jhackett1, line 73: why an object, rather than directly referencing a key somewhere in formAnswers?

JohnFarrellDev: I think this way is quite explicit making for readable code. Also means if we need anything other than the associated id we just add another property.

jest.config.js
```diff
@@ -1,3 +1,5 @@
+process.env.TZ = 'UTC';
```
JohnFarrellDev, line 1: This sets the timezone for our tests to UTC, I didn't want tests to fail once BST ends.

mapleleafjack (approved):

PR #639, opened 2021-08-06
Fix: Parse datetime correctly from FlexibleForms

JohnFarrellDev:
**What**  
We were mistakenly sending the minutes from the input datetime as seconds, this fixes that

**Anything else?**


martapederiva (approved):
LGTM

PR #640, opened 2021-08-06
Sct 425/chore remove date search

anabebs:
The date input on search has a bug which reverses the date and month inputs when sent to the back end. 
This input may be replaced with a date picker. The search functionality may have a overhaul in the future. Therefore, we have decided to comment out an not use the date input until we have a clearer way of moving forward with search.


mapleleafjack (approved):
❌ 📅 ❌

JohnFarrellDev:
I'd advise to delete code instead of commenting it out, our git history keeps track of old code so if we ever need it again it is still there

[Clean code](https://github.com/ontiyonke/book-1/blob/master/%5BPROGRAMMING%5D%5BClean%20Code%20by%20Robert%20C%20Martin%5D.pdf) Page 99 of pdf or 68 of the actual book

PR #641, opened 2021-08-06
Fix: Address mismatch between creating a new note and a person's details

Renny4Real:
**What**  
When a person's address is updated, the change is reflected in the UI when viewing the person details page. If previous addresses were saved on the person and you created a new case note or submission, the address displayed in the widget was not the same as the one found on the details page.

**Why** 
When creating a new case note or submission, the widget pulled the first address from this list of addresses for a person, and the first address may not always be the most recent address saved on a person.

This changes the code to retrieve the address we want to use when a person has a list of addresses associated with them and gets the last address added instead of the first.

We also added `data_test_id` attributes to make testing the change easier.

### DIFF
--- a/components/PersonWidget/PersonWidget.spec.tsx
+++ b/components/PersonWidget/PersonWidget.spec.tsx
@@ -54,6 +54,40 @@ describe('PersonWidget', () => {
     ).toBeNull();
   });
 
+  it('should return the last address if there are multiple addresses saved on the person', () => {
+    const firstAddress = {
+      addressLines: '100 Example Street',
+      postCode: 'E1 8HW',
+    };
+
+    const newAddress = {
+      addressLines: '123 TestCode Lane',
+      postCode: 'SE1 7DT',
+    };
+
+    const resident = residentFactory.build();
+    resident.addresses = [firstAddress, newAddress];
+
+    const { getByTestId, asFragment } = render(
+      <PersonWidget person={resident} />
+    );
+
+    expect(getByTestId('resident-last-address')).toHaveTextContent(
+      '123 TestCode Lane'
+    );
+    expect(getByTestId('resident-last-address')).toHaveTextContent('SE1 7DT');
+
+    expect(getByTestId('resident-last-address')).not.toHaveTextContent(
+      '100 Example Street'
+    );
+
+    expect(getByTestId('resident-last-address')).not.toHaveTextContent(
+      'E1 8HW'
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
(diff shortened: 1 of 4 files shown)


components/PersonWidget/PersonWidget.spec.tsx
```diff
@@ -54,6 +54,40 @@ describe('PersonWidget', () => {
...
   });
 
+  it('should return the last address if there are multiple addresses saved on the person', () => {
+    const firstAddress = {
+      addressLines: '100 Example Street',
+      postCode: 'E1 8HW',
+    };
+
+    const newAddress = {
+      addressLines: '123 TestCode Lane',
```
martapederiva, line 64: Love TestCode Lane :)

chrishutchinson (approved):
Looking good!

components/PersonWidget/PersonWidget.spec.tsx
```diff
@@ -54,6 +54,40 @@ describe('PersonWidget', () => {
...
+
+    expect(getByTestId('resident-last-address')).not.toHaveTextContent(
+      '100 Example Street'
+    );
+
+    expect(getByTestId('resident-last-address')).not.toHaveTextContent(
+      'E1 8HW'
+    );
+
+    expect(asFragment()).toMatchSnapshot();
```
chrishutchinson, line 88: Small nitpick – the `expect` statements in your test case above are pretty explicit, you can probably skip having the snapshot too

Renny4Real: I see. 🤔 Thanks!
I wasn't sure if needed the snapshot 😅 . When's the best time to use a snapshot?

chrishutchinson: There's a few schools of thoughts around snapshots. My general rules of thumb are:

* Snapshots are good for high level pages, where you want to be alerted when a smaller change in a sub-component is likely to have an impact
* Snapshots are good when it's hard to be specific with your `expect` statements, perhaps it's a complex DOM element (like a `<select>`, `<form>` or something)
* Snapshots can be good for complex data objects (e.g. a nested set of JSON) as it's just doing a string match, so you don't have to check every property explicitly

* Snapshots are bad when there are clear and concise `expect` statements which more clearly define the intent of the code under test
* Snapshots are bad when the DOM is going to change _a lot_ (you'll just be constantly updating them, better to aim for some general expect statements)
* Snapshots are bad when there isn't a lot of DOM – they encourage lazy code reviews (not maliciously, but they're easy to skip over!), better to have explicit expect statements

As with most things, used in moderation, they can be effective!

PR #642, opened 2021-08-06
Fix: Send correct formId for case notes

JohnFarrellDev:
**What**  
We were always sending the formId as `child-case-note` even when the form was actually for an adult so should be `child-case-note`

**Why**  
Bug fix.

**Anything else?**


JohnFarrellDev:
For full context see discussion here: https://hackit-lbh.slack.com/archives/C021W8CESLU/p1628252738055700

mapleleafjack (approved):

PR #643, opened 2021-08-06
Fix: Person timeline test failing as tested value is dynamic from the date

JohnFarrellDev:
**What**  
The `PersonTimeline.spec.tsx` `renders a list of clickable submissions` test started failing. This is due to the testing an output from the function `safelyFormatDistanceToNow`. Now refers to the current time, the test wasn't mocking out the current time though so this test has started failing, added a mock to Date.now that fixes this.

**Why**  
We want our test to actually work and not block the pipeline :p

**Anything else?**


mapleleafjack (approved):
📅

PR #646, opened 2021-08-11
Fix: Update endpoint call to request for in-progress submissions created two weeks from today

Renny4Real:
We seem to be hitting a limit on the size of the response from the API, this should limit the amount to submission created in the last two weeks as a temp fix.


JohnFarrellDev (approved):
:boom:

PR #647, opened 2021-08-11
Fix(SCT-983): Title handling

martapederiva:
**What**  
Frontend is now using the form name for the Casenote title and correctly mapping the form id.

**Why**  

**Anything else?**


JohnFarrellDev (approved):
:+1:

PR #648, opened 2021-08-12
Fix(SCT-1080): Change content of questions in cfs case notes form

JohnFarrellDev:
**What**  
[Ticket](https://hackney.atlassian.net/browse/SCT-1080)

Just need to change the text for a couple of questions based on feedback.
Also make sure that we now capture the date of event for the child case note.

**Anything else?**


martapederiva (approved):
✔️ 👏

PR #653, opened 2021-08-13
Fix (Sct 994): Case notes date validation

martapederiva:
**What**  
This PR is about not allowing for case notes to be created for a future date.

**Why**  
Case notes are reflections of something that has already happened.

**Anything else?**

No
No
N/A


jhackett1:
elegant!

jhackett1 (approved):

JohnFarrellDev (approved):
:+1:

PR #654, opened 2021-08-16
Fix (SCT-1086): Clean up test suit

acodeninja:
**What**  
When running the test suite there are a number of errors, warnings and  unexpected output. On investigation I found a number of issues:

* a number of async describe blocks.
* non-unique react mapped keys in mocked data.
* key props missing from some nested components.
* some unneeded debug logging in tests.
* parse date test fails on MacOS.
* promises not waited on, resulting in some tests never being executed or logging after execution
* error boundary tests with logging to cli

**Why**  
This can be a sign of smelly tests, so was worth investigating to ensure the tests were doing what we expect.

**Anything else?**

* No third party libraries added.
* Added a new helper method `getByTextIgnoringTags` which will get text on a page that is split across several html tags.

### DIFF
--- a/ci-scripts/delete-draft-github-releases.spec.js
+++ b/ci-scripts/delete-draft-github-releases.spec.js
@@ -8,6 +8,13 @@ describe('#deleteDraftGitHubReleases()', () => {
     delete global.process.env.GITHUB_TOKEN;
 
     jest.resetAllMocks();
+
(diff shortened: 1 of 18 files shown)


mapleleafjack (approved):
✅✅✅

jhackett1:
much better

jhackett1 (approved):

PR #655, opened 2021-08-16
feat(sct-1024): Load and display case statuses 

anabebs:
Screenshot 2021-08-16 at 11 50 12
**What**  
This is the PR for the first part of the creating front end journey for cse status types. The styling will need further refinement once we know exactly what is expected.

This PR displays the case status on user 43. 

The back end of the next.js front end for this get endpoint has been completed along with its tests, factories and types.

We have also added a feature flag for case status.


components/pages/people/case-status/view/case-status.tsx
```diff
@@ -0,0 +1,38 @@
...
+const CaseStatusView = ({ person }: Props): React.ReactElement => {
+  const { data: caseStatusData, error } = GetCaseStatus(person.id);
+
+  if (error) {
+    return (
+      <ErrorMessage label="There was a problem with getting case status." />
+    );
+  }
+
+  if (!caseStatusData) {
```
jhackett1, line 19: this probably isn't an appropriate use of that component, since it's visually much larger than the data it placeholds. either return nothing, or a much smaller spinner/skeleton ui

types.ts
```diff
@@ -318,3 +318,17 @@ export type WarningNote = DisclosedNote | UndisclosedNote;
...
+
+export interface PersonCaseStatus {
+  personId: number;
+  caseStatuses: Array<CaseStatus>;
+}
+
+export interface CaseStatus {
+  id: number;
+  type: string;
+  subType: string;
```
jhackett1, line 330: can we be more explicit about the possible values here? use enums?

utils/api/caseStatus.ts
```diff
@@ -0,0 +1,8 @@
+import useSWR, { SWRResponse } from 'swr';
+
+import type { PersonCaseStatus, ErrorAPI } from 'types';
+
+export const GetCaseStatus = (
```
jhackett1, line 5: convention: should be `getCaseStatus`

components/pages/people/case-status/view/case-status.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+const person = mockedResident;
+
+describe('CaseStatusView component', () => {
+  it('displays the casestatus of a person', async () => {
+    jest.spyOn(caseStatusApi, 'GetCaseStatus').mockImplementation(() => ({
+      data: mockedPersonCaseStatusFactory.build({
+        personId: person.id,
+        caseStatuses: [
+          mockedCaseStatusFactory.build({
+            type: 'CIN',
```
jhackett1, line 31: is this test content? `CIN` isn't what we're going with, i hope!

components/NewPersonView/Layout.tsx
```diff
@@ -13,6 +13,8 @@ import { canManageCases } from 'lib/permissions';
 import AddFormDialog from 'components/AddFormDialog/AddFormDialog';
 import { useState } from 'react';
 import Banner from 'components/FlexibleForms/Banner';
+import CaseStatusView from 'components/pages/people/case-status/view/case-status';
+import { ConditionalFeature } from 'lib/feature-flags/feature-flags';
```
chubberlisk, line 17: Are these needed? Doesn't look like they're used.

features.ts
```diff
@@ -20,7 +20,10 @@ export const getFeatureFlags = ({
     'person-cases': {
       isActive: environmentName === 'production',
     },
-
+    // FEATURE-FLAG-EXPIRES [2021-09-31]: person-timeline
```
chubberlisk, line 23:
```suggestion
    // FEATURE-FLAG-EXPIRES [2021-09-31]: case-status
```

utils/api/cases.spec.ts
```diff
@@ -1,62 +1,162 @@
-import axios from 'axios';
-
-import * as casesAPI from './cases';
+import * as relationshipsAPI from './relationships';
```
chubberlisk, line 1: Not sure about these changes, looks like we're changing the `cases` to have `relationships` tests. Probs some Git weirdness?

components/pages/people/case-status/view/case-status.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+      }),
+      isValidating: false,
+      mutate: jest.fn(),
+      revalidate: jest.fn(),
+    }));
+
+    const { queryByText } = render(<CaseStatusView person={person} />);
+
+    expect(queryByText(/CIN/)).not.toBeInTheDocument();
+  });
```
chubberlisk, line 59: The test description is different to the expectation we have. Are we expecting to just not see any tags i.e. no "CIN" or a "No case status found for this person"?

lib/casestatus.ts
```diff
@@ -0,0 +1,16 @@
+import axios from 'axios';
```
chubberlisk, line 1: Could we rename this file to `lib/caseStatus.ts` to follow our naming convention? Similarly for the spec file.

chubberlisk (commented):
Since it's under a feature flag, not too worried about presentation of the tags yet as we can refine that iteratively. Added some comments for things I've noticed.

components/pages/people/case-status/view/case-status.tsx
anabebs: removed spinner

types.ts
anabebs: I think these are being redefined by the team and should be strings in the future.

utils/api/caseStatus.ts
anabebs: issue with my linter. Jack is having a look.

components/NewPersonView/Layout.tsx
anabebs: changed

features.ts
anabebs: done

lib/casestatus.ts
anabebs: done

mapleleafjack (approved):

PR #657, opened 2021-08-16
fix: remove broken search/filtering from person timeline

JohnFarrellDev:
**What**  
Searching and filtering is not working correctly and is all done by the frontend which is not sensible anyway.
Removed the code.

**Why**  
Code don't work.

**Anything else?**


martapederiva (approved):
👍

PR #658, opened 2021-08-17
fix: datetime validate time correctly

JohnFarrellDev:
**What**  
Update validation for datetimes that must be in the past, can accept a future time if date is past.

**Why**  
To allow for visits/casenotes to record datetimes properly!

**Anything else?**


martapederiva (approved):
👍

PR #660, opened 2021-08-17
fix: do not validate PR title on main branch

chrishutchinson:
**What**  
We don't care about PR titles on the `main` branch, so this ignores that branch in the CircleCI config.


JohnFarrellDev (approved):
:+1:

PR #661, opened 2021-08-17
feat(SCT-1089): Loading form values from the database

mapleleafjack:
**What**  
Added necessary communication API bits for getting Form values / enums from the database for Case Status.

**Why**  
We're storing various case status types definition in the database.
To keep the FE aligned, we created a new GET endpoint in the database that serves the data.
This PR adds the necessary frontend bits to load that endpoint and use the data to build forms.

```
import { GetFormValues } from 'utils/api/caseStatus';

const {
  data: { fields: caseStatuses} =  {}
} = GetFormValues("CIN");

if (caseStatuses){
  console.log(caseStatuses);
  console.log(caseStatuses[0]);
}
```


lib/caseStatus.ts
```diff
@@ -15,3 +15,13 @@ export const getCaseStatusByPersonId = async (
   );
   return data;
 };
+
+export const GetFormValues = async (type: string): Promise<FormFields> => {
+  const { data }: { data: FormFields } = await axios.get(
+    `${ENDPOINT_API}/case-status/form-options/${type}`,
```
JohnFarrellDev, line 21: I think this is fine but I'm not sure if it fully matches the backend API?

```
[ProducesResponseType(typeof(GetCaseStatusFieldsResponse), StatusCodes.Status200OK)]
        [HttpGet]
        [Route("{type?}")]
        public IActionResult GetCaseStatusTypeFields([FromRoute] GetCaseStatusFieldsRequest request)
        {
            try
            {
                return Ok(_getCaseStatusFieldsUseCase.Execute(request));
            }
            catch (CaseStatusNotFoundException exception)
            {
                return NotFound(exception.Message);
            }
        }
```

I think it should just be 

```
[Route("{type}")]
        public IActionResult GetCaseStatusTypeFields(string type)
```

Maybe this API call is working fine because .net is managing to figure out what was intended if if the code is a bit odd.

mapleleafjack: Hi John, I'll have a look at the backend code and do the necessary edits then!

I haven't made that endpoint so I can't really say why, but I'll make those edits and checks if everything works :)

lib/caseStatus.spec.ts
```diff
@@ -29,3 +32,27 @@ describe('case status APIs', () => {
...
+
+      expect(mockedAxios.get).toHaveBeenCalled();
+      expect(mockedAxios.get.mock.calls[0][0]).toEqual(
+        `${ENDPOINT_API}/case-status/form-options/CIN`
+      );
+      expect(mockedAxios.get.mock.calls[0][1]?.headers).toEqual({
+        'x-api-key': AWS_KEY,
+      });
+
+      expect(data).toEqual(mockedData);
```
JohnFarrellDev, line 55:
```
expect(mockedAxios.get).toHaveBeenCalledWith(
        'https://dr03nduqxh.execute-api.eu-west-2.amazonaws.com/staging/api/v1/residents/123/casestatuses',
        { headers: { 'x-api-key': 'Lcu7lh4K4p9Ol4I9eAXgJa2D8xb8scpw6E8GJPKB' } }
      );
      expect(data).toEqual(caseStatus);
```

JohnFarrellDev: I think it is tidier to take a few of those assertions and move it into a single ```.toHaveBeenCalledWith(...)```

JohnFarrellDev (approved):
Looks good, only one comment on a test that I think should be addressed

lib/caseStatus.spec.ts
mapleleafjack: Thank you John, will edit the test!

PR #662, opened 2021-08-17
fix(SCT-962): Added styles for status type view

mapleleafjack:
**What**  
- Added necessary styling bits for displaying case status types in Person view.
- Test for Case Status Types adjusted
- Tests for Layout adjusted to include a FeatureFlag provider (necessary for the component)

**Why**  
It was not displaying correctly and in the incorrect section.

Screenshot 2021-08-17 at 16 15 23


chubberlisk (approved):
Definitely an improvement from before but just want to flag that from the current designs it seems the tags near the person name will just be the high level one so either "Child in need" or "Looked after child" and then somewhere else we'll have more details.

image

Okay to merge and get in though as it's under a feature flag! 👍

PR #663, opened 2021-08-18
fix: make child case note text box larger

jhackett1:
our implementation already passes extra field data as props to the input element, so all we need to do to support bigger textareas is:

- add a `rows` value to the form config
- add that to the `Field` type so typescript doesn't throw a tant


JohnFarrellDev (approved):
looks good

PR #664, opened 2021-08-19
fix: use eslint to remove all unused imports

JohnFarrellDev:
**What**  
utilise eslint plugin that removes all unused imports

**Why**  
They annoy me :p 

**Anything else?**

- Have you added any new third-party libraries? Yes, the eslint plugin


martapederiva (approved):
👍

PR #666, opened 2021-08-20
fix(SCT-1082): handle new api response from query submissions

JohnFarrellDev:
**What**  
The new API response from the GET /api/v1/submissions no longer returns EditHistory or FormAnswers. It does now return a time for LastEdited and also CompletedSteps. This PR updates various components that expected a Submission and now get an InProgressSubmission with the different properties being passed in.

**Why**  
This slightly over halves the API response size due to reducing the amount of JSON being passed back. This should hopefully allow /work-in-progress to not 500 error anymore.

**Anything else?**


components/RevisionTimeline/MiniRevisionTimeline.tsx
```diff
@@ -1,14 +1,34 @@
...
+  inProgressSubmission: InProgressSubmission;
 }
 
-const RevisionTimeline = ({ submission }: Props): React.ReactElement | null => {
+const RevisionTimeline = ({
+  inProgressSubmission,
+}: Props): React.ReactElement | null => {
+  const [submission, setSubmission] = useState<Submission>();
+
+  useEffect(() => {
```
chrishutchinson, line 19: We usually use SWR for fetching data in components, instead of calling axios directly from here. I think there's some good reasons for doing that too. One of which is that it has built in pagination logic etc., but there's definitely other pros.

Any reason why we're not doing that here?

data/flexibleForms/forms.types.ts
```diff
@@ -121,9 +126,16 @@ export interface Submission {
   residents: Resident[];
   workers: Worker[];
   editHistory: Revision[];
-  submissionState: 'In progress' | 'Approved' | 'Discarded' | 'Submitted';
+  submissionState:
+    | 'In progress'
+    | 'Approved'
+    | 'Discarded'
+    | 'Submitted'
+    | 'Panel Approved';
```
chrishutchinson, line 134: It _might_ be nice to use an `enum` here rather than a union of strings:

```ts
enum SubmissionState {
    InProgress = "In progress",
    Approved = "Approved",
    Discarded = "Discarded",
    Submitted = "Submitted",
    PanelApproved = "Panel Approved"
};

submissionState: SubmissionState;
```

chrishutchinson (changes requested):

martapederiva (approved):
Chris said we're good to go

components/RevisionTimeline/MiniSubmissionTimeline.spec.tsx
```diff
@@ -1,60 +1,133 @@
...
+      expect(screen.getByText('foo.bar@hackney.gov.uk', { exact: false }));
+      expect(screen.getByText('21 Jun 2021', { exact: false }));
+    });
   });
 
-  it('only shows up to three events', () => {
+  it('only shows up to three events', async () => {
+    const mockSubmissionClone = JSON.parse(
+      JSON.stringify(mockSubmission)
+    ) as Submission;
```
chrishutchinson, line 45: What's going on here – stringify then parse? 🤔

chrishutchinson: Ah, I see you're cloning here. Normally our fixtures use the `.build()` pattern so you can create a fresh one at the start of the test, rather than having to worry about mutation issues.

Could we take a look at doing the same thing here?

data/flexibleForms/mapFormIdsToDisplayName.ts
```diff
@@ -1,4 +1,10 @@
 export const mapIdToType: Record<string, string> = {
   'adult-case-note': 'Case note',
   'child-case-note': 'Case note',
+  foo: 'Sandbox form',
+  'review-3c': 'Review of Care and Support Plan (3C)',
+  'FACE overview assessment': 'FACE overview assessment',
```
chrishutchinson, line 6: These don't really feel like IDs? "FACE overview assessment" doesn't at least. Is that the intention?

chrishutchinson: It seems more logical to me that we'd have one mapping object for this and the other use cases, something like:

```ts
type FormDefinition = {
  form: Form;
  displayName: string;
  id: string;
};

const formMap = Record<string, FormDefinition> = {
  'adult-case-note': {
    form: formAsObjects.adultCaseNote,
    displayName: "Case note",
    id: "adult-case-note",
  },
  // ...
}
```

chrishutchinson (changes requested):

components/RevisionTimeline/MiniSubmissionTimeline.spec.tsx
JohnFarrellDev: Yeah, I'll make this change, messy way to clone instead of using build

data/flexibleForms/mapFormIdsToDisplayName.ts
JohnFarrellDev: Agreed on the single mapping.

The "id" thing is a bit messy, as "FACE overview assessment" is not currently released there is nothing stopping me just changing its id to be hyphenated like the rest of the forms.

chrishutchinson (approved):

PR #669, opened 2021-08-23
feat(SCT-1111): Case status response structure change

mapleleafjack:
**What**  
The Backend response for Case Status changed.
Frontend has been adjusted accordingly

**Why**  
To match the Backend response

**Anything else?**

Tests, factories and types changed / updated

### DIFF
--- a/components/CaseStatus/CaseStatusView.tsx
+++ b/components/CaseStatus/CaseStatusView.tsx
@@ -29,7 +29,7 @@ const CaseStatusView = ({ person }: Props): React.ReactElement => {
               key={status.id}
             >
               {getTypeString(status.type as keyof typeof valueMapping)} -{' '}
-              {status.fields.map((val) => val.selectedOption.name)}
+              {status.fields.map((val) => val.selectedOption.name).join(',')}
             </span>
           ))}
         </section>
--- a/components/CaseStatus/CaseStatusView.tsx
+++ b/components/CaseStatus/CaseStatusView.tsx
@@ -29,7 +29,7 @@ const CaseStatusView = ({ person }: Props): React.ReactElement => {
               key={status.id}
             >
               {getTypeString(status.type as keyof typeof valueMapping)} -{' '}
-              {status.subType}
+              {status.fields.map((val) => val.selectedOption.name)}
             </span>
           ))}
         </section>
--- a/factories/caseStatus.ts
+++ b/factories/caseStatus.ts
@@ -1,5 +1,10 @@
 import { Factory } from 'fishery';
-import { CaseStatus, PersonCaseStatus } from 'types';
+import {
+  CaseStatus,
+  PersonCaseStatus,
+  FieldOption,
+  CaseStatusFields,
+} from 'types';
 
 export const mockedPersonCaseStatusFactory = Factory.define<PersonCaseStatus>(
   ({ sequence }) => ({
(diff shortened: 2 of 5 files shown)


lib/caseStatus.ts
```diff
@@ -7,9 +7,11 @@ const headers = { 'x-api-key': AWS_KEY };
 export const getCaseStatusByPersonId = async (
   personId: number
 ): Promise<PersonCaseStatus> => {
-  const { data, error }: { data: PersonCaseStatus; error: Error } =
-    await axios.get(`${ENDPOINT_API}/residents/${personId}/casestatuses`, {
+  const { data }: { data: PersonCaseStatus; error: Error } = await axios.get(
```
JohnFarrellDev, line 10: Should instead of type `Error` should it be `AxiosError`?

JohnFarrellDev: Also do we even need Error?

could we just write 

`const { data } = await axios.get<PersonCaseStatus>(
    `${ENDPOINT_API}/residents/${personId}/casestatuses`,
    {
      headers,
    }
  );`

factories/caseStatus.ts
```diff
@@ -12,9 +17,20 @@ export const mockedCaseStatusFactory = Factory.define<CaseStatus>(
   ({ sequence }) => ({
     id: sequence,
     type: 'foo',
-    subType: 'bar',
-    StartDate: '2021-1-1',
-    EndDate: '2021-1-1',
-    Notes: 'a lot to note',
+    fields: [mockedStatusField.build()],
+    startDate: '2021-1-1',
+    endDate: '2021-1-1',
```
JohnFarrellDev, line 22: Are we going to be returning ISO formatted date strings from our API (I hope so).

If so we should probably make sure our mocked strings are in ISO-8601 format.

lib/caseStatus.ts
mapleleafjack: Not really needed, will remove it!

components/CaseStatus/CaseStatusView.tsx
```diff
@@ -29,7 +29,7 @@ const CaseStatusView = ({ person }: Props): React.ReactElement => {
               key={status.id}
             >
               {getTypeString(status.type as keyof typeof valueMapping)} -{' '}
-              {status.subType}
+              {status.fields.map((val) => val.selectedOption.name)}
```
JohnFarrellDev, line 32: I might not understand correctly the context but this will return an array of strings inside JSX.

If you want to print them (comma separated) something like `{status.fields.map((val) => val.selectedOption.name).join(',')}` might be what you want?

Maybe what you have currently is fine and I'm just lacking context.

factories/caseStatus.ts
mapleleafjack: Changed them to ISO

components/CaseStatus/CaseStatusView.tsx
mapleleafjack: So, does not really matter for now as we got only one result.

I've added your suggestion anyways as it's a good one, and will help when we're gonna load multiple values

JohnFarrellDev (approved):
:+1: (just a heads up we haven't done the db migration changes yet, at 3 myself Lawrence and Tuomo I think are jumping on a call)

PR #670, opened 2021-08-24
feat(SCT-1076): Use query params to filter submissions server side

JohnFarrellDev:
**What**  
A bunch of our API calls were doing filtering by taking the response and then calling `.filter`. Instead we now pass filter values into our query parameters so that the filtering can be done server side.

**Why**  
Benefits include

- cleaner code
- smaller response size when requesting data
- will work better with pagination, how can you filter correctly if you only return a paginated subset of data

**Anything else?**

- Have you added any new third-party libraries? no
- Are any new environment variables or configuration values needed? no
- Anything else in this PR that isn't covered by the what/why? no

### DIFF
--- a/components/NewPersonView/PersonTimeline.tsx
+++ b/components/NewPersonView/PersonTimeline.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { formatDistanceToNow } from 'date-fns';
 import { Case } from 'types';
 import s from './index.module.scss';
-import { Submission } from 'data/flexibleForms/forms.types';
+import { InProgressSubmission } from 'data/flexibleForms/forms.types';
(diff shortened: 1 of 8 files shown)


JohnFarrellDev:
If you want to test this manually I would suggest going to the /pages/work-in-progress.tsx

line 34 is 

```
const submissions = await getInProgressSubmissions(user?.permissionFlag);
```

change to 

```
const submissions = await getInProgressSubmissions(
    'A',
    1,
    'john.farrell@hackney.gov.uk'
  );
```

You can mess around with the parameters but should be able to see we can filter by the age context (a resident on the form belongs to the age context submitted, the resident id or the worker email.

martapederiva (approved):
🆒

PR #671, opened 2021-08-25
feat(SCT-1100): Display case statuses

mapleleafjack:
**What**  
This PR tidies up the main UI visualization for Case Status Types.
Style changed a bit as well

**Why**  
Before this, in presence of multiple case status connected to the user, they were displayed all under the user's name, one per subtype.
Now it displays only the main status type. 
A following PR will add the main component which will hold the detailed information

Screenshot 2021-08-25 at 11 28 46

### DIFF
--- a/components/CaseStatus/CaseStatusView.tsx
+++ b/components/CaseStatus/CaseStatusView.tsx
@@ -24,7 +24,7 @@ const CaseStatusView = ({ person }: Props): React.ReactElement => {
       {groupByType(caseStatusData.caseStatuses, (data: any) => data.type).map(
         (status) => (
           <span
-            className="govuk-tag lbh-tag govuk-!-margin-right-1 govuk-!-margin-top-2"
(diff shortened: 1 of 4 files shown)


components/CaseStatus/CaseStatusView.tsx
```diff
@@ -21,28 +21,39 @@ const CaseStatusView = ({ person }: Props): React.ReactElement => {
...
+            {getTypeString(status as keyof typeof valueMapping)}
+          </span>
+        )
       )}
     </>
   );
 };
 
+function groupByType(list: any, keyGetter: any) {
+  const map = new Array<string>();
```
JohnFarrellDev, line 39: Feels a bit confusing for the variable to be called map which is a string[] when the Map type exists

JohnFarrellDev:
```
function groupByType(
  allCasesStatues: CaseStatus[]
): [keyof typeof valueMapping] {
  return Array.from(new Set(allCasesStatues.map((el) => el.type))) as [
    keyof typeof valueMapping
  ];
}
```

JohnFarrellDev (approved):
:+1:

PR #672, opened 2021-08-25
Fix(SCT-935): dynamically extend form textarea

Miles-Alford:
**What**  
This PR allows the textarea component on forms to dynamically extend if a user types a large amount of text or adds paragraphs.

**Why**  
To allow users to see all the information they are inputting into a text area without having to either click and drag the textarea to extend it or content being covered by scrolling.

**Anything else?**
Nothing else needed

### DIFF
--- a/components/FlexibleForms/TextField.spec.tsx
+++ b/components/FlexibleForms/TextField.spec.tsx
@@ -83,4 +83,34 @@ describe('TextField', () => {
     );
     expect(screen.getByText('Example error'));
   });
+
+  it('should render a textarea when the as prop is specified', () => {
+    render(
+      <Formik
+        onSubmit={mockSubmit}
+        initialValues={{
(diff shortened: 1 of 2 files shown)


components/FlexibleForms/TextField.spec.tsx
```diff
@@ -83,4 +83,34 @@ describe('TextField', () => {
     );
     expect(screen.getByText('Example error'));
   });
+
+  it('should render a textarea when the as prop is specified', () => {
```
JohnFarrellDev, line 87: Could we have the same test but to show that input is rendered when the as prop isn't provided

Miles-Alford: 👍 done

JohnFarrellDev (approved):
Nice work, well done on getting through this

PR #674, opened 2021-08-27
fix(SCT-1126): nan percent complete showing in unfinished submissions

JohnFarrellDev:
**What**  
Utilise our mapping form id to form object.
Remove form property from our submissions as it is no longer added via mutation at the API level.

**Why**  
We want to show the actual percentage complete or handle errors appropriately when we can not compute % complete.

**Anything else?**


Miles-Alford (approved):
👍

PR #675, opened 2021-08-31
feat(SCT-1125): Release new person timeline

JohnFarrellDev:
**What**  

Updating the view when you go to see a residents page to incorporate a timeline

before

Screenshot from 2021-08-31 10-46-23-redacted-info

after

Screenshot from 2021-08-31 11-03-07

**Why**  

Improvements to the UI/UX

**Anything else?**


features.ts
```diff
@@ -12,14 +12,6 @@ export const getFeatureFlags = ({
     'feature-flags-implementation-proof': {
       isActive: environmentName === 'development',
     },
-    // FEATURE-FLAG-EXPIRES [2021-08-31]: person-timeline
-    'person-timeline': {
-      isActive: environmentName === 'development',
```
chrishutchinson, line 17: Typically to 'release' a feature flag you keep the flag in place, and just change the `isActive` logic to `true`. That is the least destructive, lowest code change, and lowest risk way to launch a new feature. We then later come back and tidy up the flags etc.

In this specific case it's small enough not to really matter, but probably worth thinking about for the future.

pages/people/[id]/index.tsx
```diff
@@ -39,21 +37,13 @@ const PersonPage = ({ person }: Props): React.ReactElement => {
       {events ? (
         events.length > 0 ? (
           <>
-            <ConditionalFeature name="person-timeline">
```
Comment (chrishutchinson) [inline pages/people/[id]/index.tsx:42]: No tests need to change now this is a permanent feature?

Comment (JohnFarrellDev) [reply pages/people/[id]/index.tsx:42]: Good point, will make the tests.

I have noticed something odd, most of our components are unit tested but everything that exists in /pages isn't?

features.ts
JohnFarrellDev: Right, that makes sense. Will keep in mind for next release.

pages/people/[id]/index.tsx
Comment (chrishutchinson) [reply pages/people/[id]/index.tsx:42]: Yeah, I started pulling things out of `/pages` into their own "page components" in the `/components` directory. This one is a good example: https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/main/pages/people/%5Bid%5D/details.tsx#L14.

Once they're in `/components` they can be tested just like anything else. For me, the files in `/pages` in a Next.js app should be super thin, just manage any server-rendering logic and SEO titles etc., and otherwise just refer out to other components for the vast majority of their UI.

Comment (Miles-Alford) [reply pages/people/[id]/index.tsx:42]: Out of curiosity is it better to keep `/pages` files thin? and if so why?

Comment (chrishutchinson) [reply pages/people/[id]/index.tsx:42]: @Miles-Alford It's really for testing purposes, which in and of itself isn't a great reason.

The biggest issue is the way that, typically, we aim to keep tests close to the component. This is mostly convenient for us as Engineers. So, if we put our tests in the `/pages` folder alongside a component, the directory tree might look something like this:

```
pages/
├─ person/
│  ├─ __tests__/
│  │  ├─ [id].test.tsx
│  ├─ [id].tsx
```

>Here `/pages/person/__tests__/[id].test.tsx` is the test file for `/pages/person/[id].tsx`

In Next.js, folders and file names in the `/pages` directory create the URL routing system, so to get to the person page for ID 10, the URL in your browser would be:
```
https://your.domain/person/10
```

However, the `/__tests__` directory and files would _also_ make URLs, annoyingly, the following URL would also work 
for the person with ID 10 (**and cause an error**):
```
https://your.domain/person/__tests__/10.test
```

This is not ideal, in fact, it's specifically behaviour we don't want.

Really, keeping files in `/pages` thin, and moving components out to a separate `/components` folder is to get around the issue of keeping tests close to the components causing unexpected routes.

components/NewPersonView/PersonHistory.tsx
```diff
@@ -0,0 +1,53 @@
...
+    setSize,
+    error: casesError,
+    isValidating,
+  } = useCasesByResident(personId);
+  const { data: submissionsData } = useUnfinishedSubmissions(personId);
+
+  const events = [] as Case[];
+  for (let i = 0; casesData !== undefined && i < casesData.length; i++) {
+    events.push(...casesData[i].cases);
+  }
```
chrishutchinson, line 26: Nit pick: would I be right in saying this `for` loop could go further down, underneath the `<ErrorMessage />` if statement?

chrishutchinson (approved):

pages/people/[id]/index.tsx
Comment (Miles-Alford) [reply pages/people/[id]/index.tsx:42]: Thank you for the detailed answer! @chrishutchinson

components/NewPersonView/PersonHistory.tsx
JohnFarrellDev: good point, save some CPU cycles :)

PR #679, opened 2021-09-09
feat(SCT-1024): Add case status UI

mapleleafjack:
[Link to ticket](https://hackney.atlassian.net/browse/SCT-1024)

**What**  

Added:
- Link "Add a case status" in the person view(this is visible only for children and if the person don't have any other active CIN status, behind feature flag)
- Flexible Form bits for Case Status
- Pages: Add case status, Case status review add
- Confirmation banner in `Details` page
- Communication bits to POST a case status
- Added tests, fixed previous existing ones
- Added factories

Everything is under the 'case-status' Feature flag

**Why**  
This flow let us add a Case Status in the database through the UI

**Anything else?**
- validation for `date` flexible field now support the `isfutureDateValid` parameter (true/false)
- Refactored API function calls (following the same model as the previous existing ones)
- added `CaseStatusMapping` class (that maps case statuses types to its text - "CIN" to "Child in need") and `AddCaseStatusFormData` class (that holds the POST information)

link | form | review | confirmation banner
---|---|---|---
Screenshot 2021-09-10 at 13 04 49 | Screenshot 2021-09-10 at 13 04 59 | Screenshot 2021-09-10 at 13 05 10 | Screenshot 2021-09-10 at 13 09 55


lib/validators.ts
```diff
@@ -101,9 +101,17 @@ export const generateFlexibleSchema = (
       );
     } else if (field.type === 'timetable') {
       shape[field.id] = Yup.object();
+    } else if (field.type === 'date' && field.isfutureDateValid === false) {
+      shape[field.id] = Yup.string().test(
+        'Validate date is present or past',
+        'Date cannot be in the future',
```
chubberlisk, line 107: So I understand the second string but what's the first string for?

pages/people/[id]/case-status/add/review.tsx
```diff
@@ -0,0 +1,95 @@
...
+          </Banner>
+        )}
+        <h1 className="govuk-fieldset__legend--l gov-weight-lighter">
+          Review case status details
+        </h1>
+
+        <FlexibleAnswers answers={displayValue} />
+
+        <div>
+          Do you want to add this case status?
```
Comment (chubberlisk) [inline pages/people/[id]/case-status/add/review.tsx:71]:
```suggestion
          <p className="lbh-body">Do you want to add this case status?</p> 
```

This looked quite small to me and different sized from the text above, I think it's because we're missing the `<p>` tag.

pages/people/[id]/details.tsx
```diff
@@ -4,17 +4,40 @@ import { GetServerSideProps } from 'next';
...
+    );
+  });
+
+  return (
+    <>
+      {showConfirmation && (
+        <ConditionalFeature name="case-status">
+          <ConfirmationBanner title={'Flagged status added'}>
+            <></>
+          </ConfirmationBanner>
```
Comment (chubberlisk) [inline pages/people/[id]/details.tsx:32]:
```suggestion
          <ConfirmationBanner title={'Flagged status added'} />
```
I think we should be able to do this, if a component doesn't have children.

pages/people/[id]/details.tsx
```diff
@@ -4,17 +4,40 @@ import { GetServerSideProps } from 'next';
...
   person: Resident;
 }
 
-const PersonAllocationsPage = ({ person }: Props): React.ReactElement => (
-  <Layout person={person}>
-    <PersonDetails person={person} />
-  </Layout>
-);
+const PersonAllocationsPage = ({ person }: Props): React.ReactElement => {
+  const [showConfirmation, setShowConfirmation] = useState(false);
```
Comment (chubberlisk) [inline pages/people/[id]/details.tsx:17]: Since this is specific for case status, maybe something like `showAddedCaseStatusConfirmation` is clearer?

pages/people/[id]/case-status/add/review.tsx
```diff
@@ -0,0 +1,95 @@
...
+const ReviewCaseStatusForm = (): React.ReactElement => {
+  const [status, setStatus] = useState('');
+
+  const router = useRouter();
+  const { user } = useAuth() as { user: User };
+  const personId = Number(router.query.id as string);
+  const formAnswers = router.query;
+  const valueMapping = new CaseStatusMapping();
+
+  const submitAnwers = async () => {
```
Comment (chubberlisk) [inline pages/people/[id]/case-status/add/review.tsx:23]:
```suggestion
  const submitAnswers = async () => {
```

pages/people/[id]/case-status/add/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { Form, Formik, FormikHelpers, FormikValues } from 'formik';
+import { generateInitialValues } from 'lib/utils';
+import { generateFlexibleSchema } from 'lib/validators';
+import CASE_STATUS from 'data/flexibleForms/caseStatus';
+import FlexibleField from 'components/FlexibleForms/FlexibleFields';
+import PersonView from 'components/PersonView/PersonView';
+import Button from 'components/Button/Button';
+import Link from 'next/link';
+
+const AddNewCaseStatus = (): React.ReactElement => {
```
Comment (chubberlisk) [inline pages/people/[id]/case-status/add/index.tsx:12]: What do you think about extracting this into a component so we can have more confidence that we've correctly configured the form? Like how we've done with the add relationship form, we can then have unit tests are conditional viewing, validation, disabled buttons etc.

lib/validators.ts
mapleleafjack: The first string it's the test name/description 😄

pages/people/[id]/case-status/add/review.tsx
Comment (mapleleafjack) [reply pages/people/[id]/case-status/add/review.tsx:71]: Good catch! Updated

pages/people/[id]/details.tsx
Comment (mapleleafjack) [reply pages/people/[id]/details.tsx:32]: Fixed!

This was because it was holding a link before (which I've removed for now as it's gonna be there in the next iteration)

pages/people/[id]/details.tsx
Comment (mapleleafjack) [reply pages/people/[id]/details.tsx:17]: Renamed the variable

pages/people/[id]/case-status/add/review.tsx
Comment (mapleleafjack) [reply pages/people/[id]/case-status/add/review.tsx:23]: Whoops

pages/people/[id]/case-status/add/index.tsx
Comment (mapleleafjack) [reply pages/people/[id]/case-status/add/index.tsx:12]: That's actually not a bad idea, I'll move everything to a component

Comment (mapleleafjack) [reply pages/people/[id]/case-status/add/index.tsx:12]: Content extracted into components and tests added 😄

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
```diff
@@ -0,0 +1,67 @@
+import AddCaseStatusForm from './AddCaseStatusForm';
+import { render, fireEvent } from '@testing-library/react';
+import { residentFactory } from 'factories/residents';
+
+const mockedResident = residentFactory.build();
+
+describe('ReviewAddRelationshipForm - CIN', () => {
```
chubberlisk, line 7:
```suggestion
describe('AddCaseStatusForm - CIN', () => {
```

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(getByText('Child in need')).toBeInTheDocument();
+  });
+
+  it('should disable the submit button when not completed', () => {
+    const { getByTestId } = render(
+      <AddCaseStatusForm personId={mockedResident.id} prefilledField={{}} />
+    );
+
+    expect(getByTestId('submit_button')).toHaveAttribute('disabled');
```
chubberlisk, line 21:
```suggestion
    expect(getByTestId('submit_button')).toBeDisabled();
```

This provides a clearer expectation explanation when the test fails, see https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-assertion.

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+        personId={mockedResident.id}
+        prefilledField={{
+          type: 'CIN',
+          notes: 'this is a note',
+          startDate: '2020-01-01',
+        }}
+      />
+    );
+
+    expect(getByTestId('submit_button')).not.toHaveAttribute('disabled');
```
chubberlisk, line 48: Similarily here -
```suggestion
    expect(getByTestId('submit_button')).not.toBeDisabled();
```

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { generateInitialValues } from 'lib/utils';
+import { generateFlexibleSchema } from 'lib/validators';
+import FlexibleField from 'components/FlexibleForms/FlexibleFields';
+import Button from 'components/Button/Button';
+import Link from 'next/link';
+import { useRouter } from 'next/router';
+
+const AddCaseStatusForm: React.FC<{
+  personId: number;
+  prefilledField: any;
```
chubberlisk, line 12: Minor: What do you think about `prefilledFields` instead? So it's clear it's multiple values.

chubberlisk (approved):
Thanks for addressing my feedback! 💯

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.tsx
mapleleafjack: Changed to plural!

PR #680, opened 2021-09-09
feat(SCT-1100): display case statuses

1JW1:
[link to ticket](https://hackney.atlassian.net/browse/SCT-1100)

**What**  

- created case status details component
- created the style sheet and tests for the component
- added feature flag to the component
- updated the `ExpandDetails` component to follow the new design and updated tests and snapshots
- fixed typo from `CPP` to `CP`

An example can be found on the user id `78` an `43`

Screenshot 2021-09-09 at 13 46 03

**Why**  

I created a Case Status Details component to display each case status connected to a child.


mapleleafjack (approved):
🚀🚀🚀

PR #683, opened 2021-09-14
fix(SCT-1024): Added null check to Case Status

mapleleafjack:
**What**  
Fixed a null pointer exception

**Why**  
A null pointer exception was getting triggered by the Add case status page


1JW1 (approved):
Nice, looks like the necessary change has been made to resolve the issue

PR #684, opened 2021-09-14
fix(SCT-1155): Align endpoint names

mapleleafjack:
## Link to JIRA ticket

[SCT-1555](https://hackney.atlassian.net/browse/SCT-1155)

### *What is the problem we're trying to solve*

Case status endpoint names have various way of spelling "case status".
This PR aligns them to "case-statuses"

`${ENDPOINT_API}/residents/${personId}/casestatuses` becomes `${ENDPOINT_API}/residents/${personId}/case-statuses`
`${ENDPOINT_API}/case-status/form-options/${type}` is now `${ENDPOINT_API}/case-statuses/form-options/${type}`


martapederiva (approved):
🔥

PR #685, opened 2021-09-15
fix: Remove sorting logic for tables that is never actually used

JohnFarrellDev:
**What**  
Some frontend logic exists for sorting by column name and if the sort should be ascending or descending. At the current moment though we don't ever sort these tables so I have deleted the code as it adds complexity without any functionality.

**Why**  
Cleaner code/Less code == less bugs maybe

**Anything else?**


mapleleafjack (approved):
🧹🧹🧹

PR #686, opened 2021-09-15
fix: Removed feature flagged part that's not working

mapleleafjack:
**What**  
The feature flag didn't work properly, so I'm removing the code that shouldn't be in the release

**Why**  
I was expecting the feature flag to not execute the code, but it looks like it's doing it regardless.
I'll re-introduce after checking what went wrong


JohnFarrellDev (approved):
:+1:

PR #687, opened 2021-09-15
fix: Unfinished submissions showing 0 for no reason when there are no submissions

JohnFarrellDev:
**What**  

Screenshot from 2021-09-15 17-12-07

^ should not be showing the 0...

**Why**  

It's a mistake, JSX if check meant to be comparing length and only showing unfinished submissions when the length is greater than 0, when you do something like 
```
{unfinishedSubmissionRes?.items?.length && ...}
```

in JSX if the left side of the && expression is true the right side JSX will get rendered, however if the left side is false(y) it actually displays 0, what you need to do is 

```
{unfinishedSubmissionRes?.items?.length > 0 && ...}
```

However, in this case we didn't even need the length check, if the length is 0 then we map over an empty list and render nothing which allowed reduction in nested if clauses.

**Anything else?**

### DIFF
--- a/components/NewPersonView/Layout.tsx
+++ b/components/NewPersonView/Layout.tsx
@@ -148,16 +148,6 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
           </p>
 
           <ConditionalFeature name="case-status">
-            <span hidden>
-              {casestatus &&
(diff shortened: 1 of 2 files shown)


martapederiva (approved):
⚛️ 🆒

PR #688, opened 2021-09-15
fix: Fetch unfinished submissions as paginated

JohnFarrellDev:
**What**  
Make a sensible paginated request for unfinished submissions as we currently only display 4.

**Why**  

Before

Screenshot from 2021-09-16 00-31-24

After

Screenshot from 2021-09-16 00-31-45

Admittedly I hope there aren't people on the system with 100's (242) of unfinished submissions like this staging user.

http://dev.hackney.gov.uk:3000/people/1

See a 89% decrease in response time.
See a 96% decrease in response size.

**Anything else?**

### DIFF
--- a/components/NewPersonView/UnfinishedSubmissions.spec.tsx
+++ b/components/NewPersonView/UnfinishedSubmissions.spec.tsx
@@ -60,42 +60,13 @@ describe('UnfinishedSubmissions', () => {
     expect(screen.getAllByRole('link').length).toBe(3);
   });
 
-  it('truncates a long list', () => {
+  it('calculates item left based on paginated count', () => {
     jest
       .spyOn(submissionHooks, 'useUnfinishedSubmissions')
       .mockImplementation(() => {
         const response = {
           data: {
             items: [
-              mockInProgressSubmission,
-              mockInProgressSubmission,
-              mockInProgressSubmission,
-              mockInProgressSubmission,
-              mockInProgressSubmission,
-              mockInProgressSubmission,
-            ] as InProgressSubmission[],
-            count: 6,
-          },
-        } as SWRResponse<Paginated<InProgressSubmission>, ErrorAPI>;
-
-        return response;
-      });
-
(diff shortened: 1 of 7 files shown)


components/NewPersonView/UnfinishedSubmissions.spec.tsx
```diff
@@ -60,7 +60,7 @@ describe('UnfinishedSubmissions', () => {
     expect(screen.getAllByRole('link').length).toBe(3);
   });
 
-  it('truncates a long list', () => {
```
JohnFarrellDev, line 63: We don't truncate a list now as we already return a paginated list of what we want. Reason for removal of .slice in UnfinishedSubmissions component.

components/NewPersonView/UnfinishedSubmissions.spec.tsx
```diff
@@ -71,35 +71,6 @@ describe('UnfinishedSubmissions', () => {
...
-        return response;
-      });
-
-    render(<UnfinishedSubmissions personId={1} />);
-
-    expect(screen.getByText('and 2 more'));
-    expect(screen.getAllByRole('listitem').length).toBe(5);
-  });
-
-  it('calculates item left based on response count', () => {
```
JohnFarrellDev, line 90: We don't use the count from array length to determine the total number of unfinished submissions for a user, the count is now part of the paginated response.

components/NewPersonView/UnfinishedSubmissions.tsx
```diff
@@ -43,7 +43,7 @@ const UnfinishedSubmissionsEvent = ({
     data: unfinishedSubmissionRes,
     isValidating,
     error,
-  } = useUnfinishedSubmissions(personId);
+  } = useUnfinishedSubmissions(personId, 1, 4);
```
JohnFarrellDev, line 46: I could see in the future allowing the user to load more, but this maintains the current functionality of just displaying the first 4 unfinished submissions

neilkidd (approved):
LGTM

PR #689, opened 2021-09-16
fix(SCT-1216): Case status feature flag

mapleleafjack:
**What**  
Case status feature flag failed yesterday as I put some logic in a `<ConditionalFeature>` block, expecting it to not be executed.
This logic now has been moved and adjusted with an appropriate check (by using the `isFeatureActive` function)

**Why**  
Added a better feature flag check for the Case Status logic.

### DIFF
--- a/components/NewPersonView/Layout.tsx
+++ b/components/NewPersonView/Layout.tsx
@@ -14,7 +14,10 @@ import { canManageCases } from 'lib/permissions';
 import AddFormDialog from 'components/AddFormDialog/AddFormDialog';
 import { useState } from 'react';
 import Banner from 'components/FlexibleForms/Banner';
-import { ConditionalFeature } from 'lib/feature-flags/feature-flags';
+import {
+  ConditionalFeature,
+  useFeatureFlags,
+} from 'lib/feature-flags/feature-flags';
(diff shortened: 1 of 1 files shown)


components/NewPersonView/Layout.tsx
```diff
@@ -103,6 +105,20 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
       href: `/people/${person.id}/case-note`,
     },
   ];
+
+  if (
+    isFeatureActive('case-status') &&
+    casestatus &&
+    casestatus.caseStatuses &&
+    !groupCaseStatusByType(casestatus.caseStatuses).has('CIN') &&
```
Renny4Real, line 113: Is `!groupCaseStatusByType(casestatus.caseStatuses).has('CIN')`, checking that the case statuses retrieved do not include `CIN`?

Renny4Real (approved):
👍🏾

components/NewPersonView/Layout.tsx
mapleleafjack: The check disables the "Add a case status" if there currently are active CIN for the user.
The main reason for this is because you can't have more than one CIN, the form would return an error in case you try to add another one.

It needs to be removed/adjusted when we add other case status types to the flow

PR #690, opened 2021-09-17
fix: prune request for unfinished submissions

JohnFarrellDev:
**What**  

Goes with: https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/449

The API can now return a smaller/faster response for listing our unfinished submissions.
There's two places this request is made.

First any person page with the UnfinishedSubmissions component but also the work-in-progress page that lists all the unfinished submissions for a worker/all workers.

**Why**  
To improve API speed. Everything should feel "near instant"

**Anything else?**

### DIFF
--- a/components/Form/types.ts
+++ b/components/Form/types.ts
@@ -57,6 +57,7 @@ export interface GenericField {
   isMulti?: boolean;
   isMultiInit?: boolean;
   isMultiTrigger?: string;
+  buttonStyle?: string;
 }
 
 export interface Checkbox extends GenericField {
--- a/components/FormWizard/DynamicInputMulti.jsx
+++ b/components/FormWizard/DynamicInputMulti.jsx
@@ -16,12 +16,14 @@ const DynamicInputMulti = ({
   hint,
   isMultiTrigger = 'Add a new one',
   isMultiInit = true,
+  buttonStyle,
   ...otherProps
 }) => {
   const multiPresent = initialInputData.length || 0;
   const [counter, setCounter] = useState(
     multiPresent === 0 && isMultiInit ? 1 : 0
   );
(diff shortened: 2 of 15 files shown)


chubberlisk (approved):
Makes sense to me! The only feedback I have would require restructuring the tests in a way where a test is only expecting what's relevant to it. E.g. we have a test `only queries for submissions using the requested age context` so the only thing I'd see is relevant to that is that the URL `mockedAxios.get` has been called has the `ageContext` query parameter. The rest of the parameters aren't of concern in that test. Therefore, when there a new query parameter is needed we just add a new test and don't have to update all other tests. Also, it means our tests would document what exactly it's asking for from the service API.

JohnFarrellDev:
> Makes sense to me! The only feedback I have would require restructuring the tests in a way where a test is only expecting what's relevant to it. E.g. we have a test `only queries for submissions using the requested age context` so the only thing I'd see is relevant to that is that the URL `mockedAxios.get` has been called has the `ageContext` query parameter. The rest of the parameters aren't of concern in that test. Therefore, when there is a new query parameter is needed we just add a new test and don't have to update all other tests. Also, it means our tests would document what exactly it's asking for from the service API.

I get your point but that would mean rewriting the test to not use `.toHaveBeenCalledWith` with our mocked axios call.
We'd have to get the mocked Axios query another way and then test that the string contains the correct ageContext query.
Testing that a string contains some text instead of being an explicit value potentially makes our test less strict. Could our test keep passing even though ageContext is somehow breaking the query.

I think the default parameters or hard coded queries are fine, we wouldn't need to update the existing tests unless the new query is hard-coded or provides default params which seems unlikely.

PR #691, opened 2021-09-17
feat(SCT-1083):  Phone types label improvements

martapederiva:
**What**  
Phone number has now a hint, phone type is now called notes and has a hint instead of placeholder, different style to the "add another" button.

**Why**  
Users want a clearer labelling of the phone number fields when adding a new person

**Anything else?**

### DIFF
--- a/components/Form/types.ts
+++ b/components/Form/types.ts
@@ -57,6 +57,7 @@ export interface GenericField {
   isMulti?: boolean;
   isMultiInit?: boolean;
   isMultiTrigger?: string;
+  buttonStyle?: string;
 }
 
 export interface Checkbox extends GenericField {
--- a/components/FormWizard/DynamicInputMulti.jsx
+++ b/components/FormWizard/DynamicInputMulti.jsx
@@ -16,12 +16,14 @@ const DynamicInputMulti = ({
   hint,
   isMultiTrigger = 'Add a new one',
   isMultiInit = true,
+  buttonStyle,
   ...otherProps
 }) => {
   const multiPresent = initialInputData.length || 0;
   const [counter, setCounter] = useState(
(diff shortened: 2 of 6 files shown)


components/Form/types.ts
```diff
@@ -57,6 +57,7 @@ export interface GenericField {
   isMulti?: boolean;
   isMultiInit?: boolean;
   isMultiTrigger?: string;
+  buttonStyle?: string;
```
Renny4Real, line 60: Why are we adding a `buttonStyle` prop?

components/FormWizard/DynamicInputMulti.jsx
```diff
@@ -65,7 +67,12 @@ const DynamicInputMulti = ({
       ))}
       <div className="govuk-!-margin-top-3 govuk-!-margin-bottom-5">
         <button
-          className="govuk-link"
+          className={
+            buttonStyle === 'button'
```
Renny4Real, line 71: This confuses me slightly.

We set a `buttonStyle` prop equal to 'button' within the className property object, but the type is also set to "button" on line 75, all within a button component. 

I might also not understand these lines of code well, so happy for things to stay if we find that we need them. We could refactor, e.g. rename stuff to make it more straightforward 🤔

Renny4Real (commented):
Looks good, just a couple of questions to understand the code better. 🙏🏾

components/Form/types.ts
martapederiva: this is to handle a different CSS style for this button as per users request
the button is created dynamically by a component used in various places throughout the app
the prop will pass the CSS class itself (thanks @mapleleafjack), if not the usual class will be applied

JohnFarrellDev (approved):
:+1:

PR #692, opened 2021-09-20
chore: Upgrade set-value to version 4.0.1 or later

Renny4Real:
Dependabot raised a security alert [here](https://github.com/LBHackney-IT/lbh-social-care-frontend/security/dependabot/yarn.lock/set-value/open) about the set-value package needing an upgrade to a version that has been patched to fix a vulnerability.

It seems Dependabot was unable to do this automatically, so this PR manually attempts to fix the issue raised.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "restana": "^4.9.1",
     "serve-static": "^1.14.1",
     "serverless-http": "^2.7.0",
+    "set-value": "^4.1.0",
     "swr": "^0.5.6",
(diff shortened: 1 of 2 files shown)


JohnFarrellDev:
The link for the security alert no longer works

Renny4Real:
@JohnFarrellDev 🤔 Are you able to see the alert when you go to the repo?

JohnFarrellDev (approved):

PR #695, opened 2021-09-23
feat(SCT-1055): Add CP option to the case status add form UI

Renny4Real:
**What**  
This adds changes that would allow a user to add a CP case status with a specific category to a child resident.

**Why**  
Users want to be able to add and view a CP case status saved on a child resident.

### DIFF
--- a/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
+++ b/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
@@ -5,7 +5,7 @@ import { residentFactory } from 'factories/residents';
 const mockedResident = residentFactory.build();
 
 describe('AddCaseStatusForm - CIN', () => {
-  it('displays the form', () => {
+  it('displays CIN type option in the form', () => {
     const { getByText } = render(
       <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
     );
@@ -13,7 +13,7 @@ describe('AddCaseStatusForm - CIN', () => {
     expect(getByText('Child in need')).toBeInTheDocument();
   });
 
-  it('should disable the submit button when not completed', () => {
+  it('should disable the submit button when no type has been selected', () => {
     const { getByTestId } = render(
       <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
     );
@@ -28,7 +28,6 @@ describe('AddCaseStatusForm - CIN', () => {
 
     fireEvent.click(getByText('Child in need'));
 
-    expect(getByText('Child in need')).toBeInTheDocument();
     expect(getByText('Start Date')).toBeInTheDocument();
     expect(getByText('Notes')).toBeInTheDocument();
   });
@@ -85,19 +84,18 @@ describe('AddCaseStatusForm - CP', () => {
     ).not.toBeInTheDocument();
   });
 
-  it('displays start date and category when selecting CP', () => {
+  it('displays start date and category question when CP is selected', () => {
     const { getByText } = render(
       <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
     );
 
     fireEvent.click(getByText('Child protection'));
 
-    expect(getByText('Child protection')).toBeInTheDocument();
     expect(getByText('Start Date')).toBeInTheDocument();
     expect(getByText('Category of child protection plan')).toBeInTheDocument();
   });
 
-  it('displays the category options when selecting CP', () => {
+  it('displays the category options when CP is selected', () => {
     const { getByText } = render(
       <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
     );
@@ -108,7 +106,21 @@ describe('AddCaseStatusForm - CP', () => {
     expect(getByText('Physical abuse')).toBeInTheDocument();
     expect(getByText('Emotional abuse')).toBeInTheDocument();
     expect(getByText('Sexual abuse')).toBeInTheDocument();
+  });
+
+  it('pre-select CP and fills the other fields', () => {
+    const { getByText } = render(
+      <AddCaseStatusForm
+        personId={mockedResident.id}
+        prefilledFields={{
+          type: 'CP',
+          category: 'C1',
+        }}
+      />
+    );
+
+    expect(getByText('Child protection')).toBeInTheDocument();
     expect(getByText('Start Date')).toBeInTheDocument();
-    expect(getByText('Category of child protection plan')).toBeInTheDocument();
+    expect(getByText('Neglect')).toBeInTheDocument();
   });
 });
--- a/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
+++ b/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
@@ -75,7 +75,7 @@ describe('AddCaseStatusForm - CP', () => {
     expect(getByText('Child protection')).toBeInTheDocument();
   });
 
-  it('displays start date and protection categories when selecting CP', () => {
+  it('displays start date when selecting CP', () => {
     const { getByText } = render(
       <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
     );
--- a/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
+++ b/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
@@ -65,3 +65,24 @@ describe('AddCaseStatusForm - CIN', () => {
     expect(getByText('this is a note')).toBeInTheDocument();
   });
 });
+
+describe('AddCaseStatusForm - CP', () => {
+  it('displays the form', () => {
+    const { getByText } = render(
+      <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
+    );
+
+    expect(getByText('Child protection')).toBeInTheDocument();
+  });
+
+  it('displays start date and protection categories when selecting CP', () => {
+    const { getByText } = render(
+      <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
+    );
+
+    fireEvent.click(getByText('Child protection'));
+
+    expect(getByText('Child protection')).toBeInTheDocument();
+    expect(getByText('Start Date')).toBeInTheDocument();
+  });
+});
--- a/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
+++ b/components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
@@ -75,7 +75,17 @@ describe('AddCaseStatusForm - CP', () => {
     expect(getByText('Child protection')).toBeInTheDocument();
   });
 
-  it('displays start date when selecting CP', () => {
+  it('does not display category question unless CP is clicked the form', () => {
+    const { queryByText } = render(
+      <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
+    );
+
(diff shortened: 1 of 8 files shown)


Renny4Real:
TO DO:
- We grab the value of field options via an API call and use them to change the form fields depending on whether `CIN` or `CP` is selected. However, a UI bug occurs where if a user is reviewing what they want to add and then goes back to the add step , the field for selecting the categories options is repeated. This goes aways after a refresh but not sure if there is something we could do code-wise to stop this behaviour.

Renny4Real:
> **Thoughts?**
> We grab the value of field options for `CP` via an API call and use them to add extra form fields if `CP` is selected.
> 
> However, a UI bug occurs where if a user is reviewing what they want to add and then goes back to the add step , the question for choosing the `CP` category gets repeated.
> 
> This goes aways after a refresh but not sure if there is something we could do code-wise to stop this behaviour.

Fixed in cf77d0f

JohnFarrellDev:
> **Thoughts?**
> We grab the value of field options for `CP` via an API call and use them to add extra form fields if `CP` is selected.
> 
> However, a UI bug occurs where if a user is reviewing what they want to add and then goes back to the add step , the question for choosing the `CP` category gets repeated.
> 
> This goes aways after a refresh but not sure if there is something we could do code-wise to stop this behaviour.

Could we please get the questions from the API, I know this annoying considering you've already done this work but we're way over complicating the backend/db architecture by storing the questions in the database.

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
```diff
@@ -65,3 +65,24 @@ describe('AddCaseStatusForm - CIN', () => {
...
+
+describe('AddCaseStatusForm - CP', () => {
+  it('displays the form', () => {
+    const { getByText } = render(
+      <AddCaseStatusForm personId={mockedResident.id} prefilledFields={{}} />
+    );
+
+    expect(getByText('Child protection')).toBeInTheDocument();
+  });
+
```
JohnFarrellDev, line 77: Should we test the validation?

Start date can be set in the past (at least on chrome)

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.tsx
```diff
@@ -6,20 +6,52 @@ import FlexibleField from 'components/FlexibleForms/FlexibleFields';
...
+import { Choice, Field } from 'data/flexibleForms/forms.types';
 
 const AddCaseStatusForm: React.FC<{
   personId: number;
   prefilledFields: any;
 }> = ({ personId, prefilledFields }) => {
   const router = useRouter();
-  const form_fields = CASE_STATUS.steps[0].fields;
+
+  const form_fields = [...CASE_STATUS.steps[0].fields];
```
JohnFarrellDev, line 18: ```react_devtools_backend.js:4049 Warning: Encountered two children with the same key, `startDate`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.```

there is a shared id meaning the key is being shared I think

components/CaseStatus/AddCaseStatusForm/ReviewAddCaseStatusForm.tsx
```diff
@@ -41,18 +49,33 @@ const ReviewAddCaseStatusForm: React.FC<{
...
+        if (elm === elmField.name) {
+          elmField.options.map((option) => {
+            if (option.name === formAnswers[elm]) {
+              diplayObj[elmField.description] = option.description;
+            }
+          });
+        }
+      });
+    });
+  }
```
JohnFarrellDev, line 74: Can you explain this please

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
JohnFarrellDev, line 77: Test the validation works as expected?

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.tsx
```diff
@@ -29,9 +61,7 @@ const AddCaseStatusForm: React.FC<{
         pathname: `/people/${personId}/case-status/add/review`,
         query: {
           personId: personId,
-          type: values.type,
-          startDate: values.startDate,
-          notes: values.notes,
+          ...values,
         },
       });
     } catch (e) {
```
JohnFarrellDev, line 67: Can we handle this TS error please.

```
catch (e: unknown) {
      if (typeof e === 'string') {
        setStatus(e.toString());
      }
    }
```

Renny4Real:
> > **Thoughts?**
> > We grab the value of field options for `CP` via an API call and use them to add extra form fields if `CP` is selected.
> > However, a UI bug occurs where if a user is reviewing what they want to add and then goes back to the add step , the question for choosing the `CP` category gets repeated.
> > This goes aways after a refresh but not sure if there is something we could do code-wise to stop this behaviour.
> 
> Could we please get the questions from the API, I know this annoying considering you've already done this work but we're way over complicating the backend/db architecture by storing the questions in the database.

@JohnFarrellDev I don't understand your suggestion 😞 . Could you clarify 🙏🏾 ? Are you saying the questions also stored in the API? Which API are we referring to? 
Also I've been able to resolve the initial UI bug but not sure if what you're suggesting is a change in the scope of this ticket? 🤔

utils/api/caseStatus.ts
```diff
@@ -24,4 +24,4 @@ export const useCaseStatuses = (
 export const useFormValues = (
   type: string
```
JohnFarrellDev, line 25: If there are only 3 possible types ("CIN, "CP" and "LAC") could we utilise an Enum here

JohnFarrellDev:
> > > **Thoughts?**
> > > We grab the value of field options for `CP` via an API call and use them to add extra form fields if `CP` is selected.
> > > However, a UI bug occurs where if a user is reviewing what they want to add and then goes back to the add step , the question for choosing the `CP` category gets repeated.
> > > This goes aways after a refresh but not sure if there is something we could do code-wise to stop this behaviour.
> > 
> > 
> > Could we please get the questions from the API, I know this annoying considering you've already done this work but we're way over complicating the backend/db architecture by storing the questions in the database.
> 
> @JohnFarrellDev I don't understand your suggestion  . Could you clarify 🙏🏾 ? Are you saying the questions also stored in the API? Which API are we referring to?
> Also I've been able to resolve the initial UI bug but not sure if what you're suggesting is a change in the scope of this ticket?

Basically we should not be making this request `http://dev.hackney.gov.uk:3000/api/casestatus/form-options?type=CP`, there's no benefit to storing the form-options in the database which then has to be fetched by an API call. 

It creates a lot of unnecessary complexity. 

I know that wasn't your design decision but I would really appreciate if we got the fields for child-protection by just having the data in the frontend.

Renny4Real:
> > > > **Thoughts?**
> > > > We grab the value of field options for `CP` via an API call and use them to add extra form fields if `CP` is selected.
> > > > However, a UI bug occurs where if a user is reviewing what they want to add and then goes back to the add step , the question for choosing the `CP` category gets repeated.
> > > > This goes aways after a refresh but not sure if there is something we could do code-wise to stop this behaviour.
> > > 
> > > 
> > > Could we please get the questions from the API, I know this annoying considering you've already done this work but we're way over complicating the backend/db architecture by storing the questions in the database.
> > 
> > 
> > @JohnFarrellDev I don't understand your suggestion  . Could you clarify 🙏🏾 ? Are you saying the questions also stored in the API? Which API are we referring to?
> > Also I've been able to resolve the initial UI bug but not sure if what you're suggesting is a change in the scope of this ticket?
> 
> Basically we should not be making this request `http://dev.hackney.gov.uk:3000/api/casestatus/form-options?type=CP`, there's no benefit to storing the form-options in the database which then has to be fetched by an API call.
> 
> It creates a lot of unnecessary complexity.
> 
> I know that wasn't your design decision but I would really appreciate if we got the fields for child-protection by just having the data in the frontend.

Let's share this with the rest of the team and get some thoughts.

I don't know why the form options and questions were saved in the database and would want to check this. ( I'm not convinced that there's no benefit yet, I would like to see if I can understand the decision first)

My initial approach involved hard-coding the questions in the frontend, and then I got a bit stuck 😅 ; @mapleleafjack recommended calling the API to grab the form values needed from there. This also makes me think we should probably brainstorm together or share and understand this as a team. Make sure we're all on the same page, even if the changes are only for the `CP` values.

What do you think?

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
Renny4Real: Is this the validation on the start date?

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.tsx
Renny4Real: I think this was because we duplicated the start date form field so that it appear when you selected `CIN` or `CP`.

The start date field should appear if you selected either `CIN` and `CP`.

We duplicated the field as a work around because the start date form field had a condition set to appear only when you select a`CIN` and the conditions field on form fields doesn't handle OR operations. So essentially we override it depending on what was clicked. 

See [slack thread](https://hackit-lbh.slack.com/archives/C021Q88C953/p1632153326209500) for the conversation that inspired this but I'm starting to wonder if the start date needs to hide before a case status option is selected. 🤔 Maybe it can always been shown?

components/CaseStatus/AddCaseStatusForm/ReviewAddCaseStatusForm.tsx
Renny4Real: I think I could rename some of the variables to be clearer.

It loops through the form options retrieved from the API call to check if there are any matches in the the form answers and get the human readable description.

For example, if the user selected `CP` as the first option and then chose `Neglect` as a category of CP.

In the above case, the query sent to the review stage, has `CP` as the `type` parameter but `C1` as the category. The `Neglect` option the user sees is the description of the category selected i.e `C1`.

It might be easier to explain in person: The structure of our API request body to create a CP/LAC case status does not take the description `Neglect`, it takes the name of the option selected which is `C1`. This doesn't pass the work `Neglect` in the query parameters.

So the code above, loops through the options retrieved from the API, finds a match in the form answers, and if there is, rather than display the short hand code, returns the description so that we can display it to the user at the review stage.

Open to suggestions on better ways to do this.

The way this was coded was also in anticipation of LAC using the same process, so open to feedback on approach too.

utils/api/caseStatus.ts
Renny4Real: I'm not sure

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.tsx
Renny4Real: Fixed in 0dff9d4

components/CaseStatus/AddCaseStatusForm/AddCaseStatusForm.spec.tsx
Renny4Real: @JohnFarrellDev Which validation were you referring to here?

Renny4Real:
@tony-griffin and I tried to add a test for checking the submit button was disabled for CP until you filled it in but we had some trouble with setting the test up correctly. 

For some reason, even when we copied and pasted the exact same test (for the disabled submit button) from the `CIN` describe block and put it in the `CP` describe block, the test failed in the `CP` describe block.
 
Not sure if anyone has any thoughts on this or knows what might be happening? 🤔

martapederiva (approved):
✔️

PR #696, opened 2021-09-27
fix: add a content security policy

acodeninja:
**What**  

* compatibility with nextjs 'unsafe-inline' for styles and scripts
* compatibility with google tag manager and GA
* lock-down of usage in frames
* allow loading fonts from google fonts cdn
* allow loading HotJar custom styles and scripts

**Why**  
Currently, there is no content security policy leaving the application open to remote inclusion attacks.

**Anything else?**

- Have you added any new third-party libraries? Nope
- Are any new environment variables or configuration values needed? Nope

This should be tested in staging before rolling to production as it may lead to the site not functioning correctly once deployed. I've covered all the situations I could think of and tested in the local dev environment.


mapleleafjack (approved):
I guess next step is checking if everything works in staging :D

neilkidd (approved):

PR #698, opened 2021-09-28
chore: When case status array is empty, don't show the case status title

Renny4Real:
**What**  
This adds a condition for when the case status array is empty.

**Before**
Screenshot 2021-09-28 at 15 30 02

**After**
Screenshot 2021-09-28 at 15 30 19

**Why**  
The details page would show the case statuses title even if there were no case statuses. This checks if the array is empty as well as if it's undefined before returning the UI for case statuses.

**Anything else?**


martapederiva (approved):
🔥

PR #699, opened 2021-09-28
feat: add redirect functionality to the edit person details form

jhackett1:
if you add a `redirectUrl` query parameter to the edit person details form, it will redirect to that url upon completion, rather than to the normal confirmation page.

for example, `/people/1/edit?redirectUrl=http://example.com`

this lets us embed editing person details in other flows and user journeys, such as starting a new workflow.

because the edit person details form still uses the old form wizard that we're in the process of deprecating, this code might look a bit messy.

we:

- updated the `<SummaryStep/>` component to redirect on submit if there is a `redirectUrl` param in the url
- augmented the `<FormWizard/>` component to preserve a `redirectUrl` param as the user moves through steps
- improved and simplified test coverage of the old form wizard


acodeninja (approved):
Looks good to me.

PR #700, opened 2021-09-28
feat(SCT-1057, SCT-1161):  Create UI for editing and ending a person case status 

mapleleafjack:
Link to ticket 
[Create UI for editing a Case Status](https://hackney.atlassian.net/browse/SCT-1161)
[Create UI for ending a Case Status](https://hackney.atlassian.net/browse/SCT-1057)

**What**  
-Changed `components/ExpandDetails/ExpandDetails.tsx` so it supports link in the dropdown (as per design)
-Added folder  `pages/people/[id]/case-status/[caseStatusId]/edit` and pages `edit.tsx`, `index.tsx`, `review.tsx`.
-Added components `components/CaseStatus/EditCaseStatusForm/ChooseEditCaseStatusForm.tsx`, `components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx` and  `components/CaseStatus/EditCaseStatusForm/ReviewEditCaseStatusForm.tsx` + tests
-Added  `data/flexibleForms/caseStatus` folder that holds all the various CaseStatus forms
-Added flexible forms definition: `chooseEdit`, `editCaseStatus`, `endCaseStatus` 
-Added  `CaseStatusMapping` enum that holds the long string information for case statuses 
-Changed the check that makes appear the main link for adding a case status (now it will hide if there's at least one)
-`caseStatusConfirmationText` for the main confirmation banner is now dynamic

**Why**  
To let the user edit and end a case status of type CIN and CP

**Anything else?**

-Added a new validation for flexible forms date field: you can add a `startDate` property and the datepicker will return a user error if the date selected is before this date
-Adjusted the way the Form fields loads the default for select dropdowns


types.ts
```diff
@@ -365,13 +365,28 @@ export enum ChildProtectionCategoryOptions {
...
+export interface CaseStatusFormValue {
+  name: string;
+  selected: string;
+}
+
+export interface EditCaseStatusFormData {
+  personId: number;
+  caseStatusID: number;
+  values?: CaseStatusFormValue[];
+  startDate?: string;
```
JohnFarrellDev, line 384: I don't think the start date is optional

types.ts
```diff
@@ -365,13 +365,28 @@ export enum ChildProtectionCategoryOptions {
...
 
+export interface CaseStatusFormValue {
+  name: string;
+  selected: string;
+}
+
+export interface EditCaseStatusFormData {
+  personId: number;
+  caseStatusID: number;
+  values?: CaseStatusFormValue[];
```
JohnFarrellDev, line 383: I don't think values is optional.

types.ts
JohnFarrellDev, line 384: I don't think we can change the start date.

pages/people/[id]/details.tsx
```diff
@@ -16,19 +16,24 @@ interface Props {
...
+      setCaseStatusConfirmationText(String(router.query.message));
+    }
   });
 
   return (
     <>
       {showAddedCaseStatusConfirmation && (
         <ConditionalFeature name="case-status">
-          <ConfirmationBanner title={'Flagged status added'} />
+          <ConfirmationBanner title={caseStatusConfirmationText} />
```
Comment (JohnFarrellDev) [inline pages/people/[id]/details.tsx:36]: Would this work instead

```
<ConfirmationBanner
            title={String(router.query.message) ?? 'Flagged status added'}
          />
```

Then we can remove a state from the component and reduce the useEffect

Comment (JohnFarrellDev) [reply pages/people/[id]/details.tsx:36]: I also use an assertion instead of creating a new string when possible, this saves us actually having to create the new string object.

pages/people/[id]/details.tsx
```diff
@@ -16,19 +16,24 @@ interface Props {
...
 
   useEffect(() => {
-    setshowAddedCaseStatusConfirmation(
-      Boolean(router.query.flaggedStatus) && person.contextFlag === 'C'
-    );
+    setshowAddedCaseStatusConfirmation(Boolean(router.query.flaggedStatus));
+    if (router.query.message) {
+      setCaseStatusConfirmationText(String(router.query.message));
+    }
   });
```
Comment (JohnFarrellDev) [inline pages/people/[id]/details.tsx:30]:
```
useEffect(() => {
    setshowAddedCaseStatusConfirmation(Boolean(router.query.flaggedStatus));
  }, [router.query.flaggedStatus]);
```

pages/api/casestatus/index.ts
```diff
@@ -34,6 +34,22 @@ const endpoint: NextApiHandler = async (
               .json({ message: 'Unable to add the casestatus' });
       }
       break;
+    case 'PATCH':
+      try {
+        await patchCaseStatus(Number(req.query?.caseStatusId), req.body);
+        res.status(StatusCodes.OK).end();
+      } catch (error) {
```
JohnFarrellDev, line 41: Can we explicitly define the type of the error?

lib/validators.ts
```diff
@@ -101,6 +101,15 @@ export const generateFlexibleSchema = (
       );
     } else if (field.type === 'timetable') {
       shape[field.id] = Yup.object();
+    } else if (field.type === 'date' && field.startDate != null) {
```
JohnFarrellDev, line 104: !==

components/CaseStatus/AddCaseStatusForm/ReviewAddCaseStatusForm.tsx
```diff
@@ -48,7 +46,7 @@ const ReviewAddCaseStatusForm: React.FC<{
   };
 
   const displayObj: any = {
```
JohnFarrellDev, line 48: can we remove the any here? Just let the type be inferred correctly?

JohnFarrellDev (approved):
I think this is mostly fine but it's also quite a lot of changes for a single PR.

PR #701, opened 2021-10-01
feat(SCT-1195): add panel on flag hover

1JW1:
**What**  
Added the Tip component and used it for the case status types

Screenshot 2021-10-01 at 15 22 21

**Why**  
To display information on the case status on hover and give the user another way to access details section


mapleleafjack (commented):
🚀🚀🚀

mapleleafjack (approved):

PR #704, opened 2021-10-04
feat(SCT-1215, SCT-1227): Edit and End LAC case status

mapleleafjack:
**What**  
- Added UI bits for editing end ending LAC case status.
- Added enum types for LAC 
- Flexible form now loads enum types dynamically instead of them being hardcoded
- Tidied up code
- Added tests

**Why**  
To edit / end a LAC case status

| Main view | 
| ------------ |
| Screenshot 2021-10-04 at 13 15 23 | 
| Screenshot 2021-10-04 at 13 01 45 |

| Edit stage | End stage | 
| ---------- | ----------|
|  Screenshot 2021-10-04 at 13 01 49 | Screenshot 2021-10-04 at 13 01 52 |


components/CaseStatus/EditCaseStatusForm/ReviewEditCaseStatusForm.spec.tsx
```diff
@@ -48,6 +98,54 @@ describe('ReviewEditCaseStatusForm', () => {
...
+      getByText(
+        'E4A: Returned home to live with parent(s), relative(s), or other person(s) with parental responsibility as part of the care planning process (not under a special guardianship order or residence order or (from 22 April 2014) a child arrangement order).'
+      )
+    ).toBeInTheDocument();
+    expect(
+      getByText('Do you want to end this case status?')
+    ).toBeInTheDocument();
+  });
+
   it('displays start date and notes for the edit', () => {
```
tony-griffin, line 149: I'm unsure of the value of this tests now that you've updated the other test for Edit in CIN? (Line 8)

Could we add the check for the startDate into that test instead?

tony-griffin (approved):
I've left a few comments, more questions really, but all looks good.

components/CaseStatus/EditCaseStatusForm/ReviewEditCaseStatusForm.spec.tsx
mapleleafjack: It mainly checks the general visualisation of the component!

Checking the startDate is another way of doing that, either way they're CIN-related properties only :)

PR #706, opened 2021-10-04
feature: Use ajax spider in addition to traditional zap baseline scan

neilkidd:
**What**  

Add further scanning to the baseline security scan by enabling the Ajax spider **in addition** to the traditional one.


acodeninja (approved):
Seems like a good value add to me!

PR #707, opened 2021-10-05
fix: update content security policy

mapleleafjack:
**What**  

- Added CSP directive `'sha256-TGjSkcFoVCRdq4Hp3hcETnPc9cikdBGvkz/08Bjzy0I='` for the `style-src-elem`element
- Added `in.hotjar.com` to the `connect-src` directive 
- Update tests

**Why**  
Added CSP string to `contentSecurity.ts`


acodeninja (approved):
Look right to me

PR #708, opened 2021-10-05
feat(SCT-1235): Write UI for adding LAC

tony-griffin:
**What**  

- Added UI radio button to allow selection of LAC case status
- Added LAC properties to the case status flexible form
- Define post object to use for adding a case status, this will be sent to the backend
- Added enum types for LAC
- Added tests for AddCaseStatusForm & ReviewAddCaseStatusForm

**Why**  
To add a LAC case status

| Add case status page | 
| ------------ |
Screenshot 2021-10-05 at 12 45 00

|  Add case status page - LAC selected | Review page - adding LAC selected | 
| ------------ | ------------ |
Screenshot 2021-10-05 at 12 45 17 | Screenshot 2021-10-05 at 12 45 35


mapleleafjack (approved):
🚀🚀🚀

PR #709, opened 2021-10-05
feat(SCT-1224): END a case status

mapleleafjack:
**What**  
Moved the `patch` endpoint in its own folder
Adjusted the caseStatusID parameter so it gets it correctly

**Why**  
This PR let us END a case status.


tony-griffin (approved):
Looks good

PR #720, opened 2021-10-15
refactor: retrieval of workflows pilot URL

chubberlisk:
**What**  

- Add a new React context called `AppConfigContext` to store `NEXT_PUBLIC` like values. It'll store values retrieved server-side.
- Refactor all places where we used `NEXT_PUBLIC_WORKFLOWS_PILOT_URL` to use the hook to get the value from the new context.
- A new `WORKFLOWS_PILOT_URL` env var has been added to CircleCI config.
- **New:** documentation for how to expose an environment variable to the browser.

**Why**  

We added a new env var `NEXT_PUBLIC_WORKFLOWS_PILOT_URL` and updated the serverless config but yesterday when we launch the pilot we realised that prod was pointing to our staging. That addition didn’t have an effect as Next.js resolves `NEXT_PUBLIC` env vars at build time. We did a quick fix by changing the value in CircleCI to point to prod but that means staging is currently incorrect.

**Anything else?**

- A bit of a big diff as we had to update all unit tests to use new provider.
- Would recommend review commit by commit as it'll be easier to follow through, although let me know if it'll be easier to review with the refactoring as a separate PR.
- Looking for feedback for how best to handle the error thrown if config value isn't found.


components/Dashboard/DashboardWrapper.tsx
```diff
@@ -43,38 +44,50 @@ interface Props {
...
-                >
-                  Workflows
-                </NavLink>
-              </ConditionalFeature>
-            </ul>
-          </nav>
-          <MyData />
+  try {
+    navLink = (
+      <NavLink href={getConfigValue('workflowsPilotUrl') as string}>
```
chrishutchinson, line 54: Instead of casting here, we could make `getConfigValue` take a restricting generic type argument, something like:

```tsx
getConfigValue<string>("workflowsPilotUrl");
```

Happy to pair with you to go over how to do this

components/Dashboard/DashboardWrapper.tsx
```diff
@@ -43,38 +44,50 @@ interface Props {
...
-                <NavLink
-                  href={process.env.NEXT_PUBLIC_WORKFLOWS_PILOT_URL as string}
-                >
-                  Workflows
-                </NavLink>
-              </ConditionalFeature>
-            </ul>
-          </nav>
-          <MyData />
+  try {
```
chrishutchinson, line 52: I would probably do this slightly differently than you have here. The `try` / `catch` is a little odd inside the React component.

Is there an issue if we just inline it further down, like so:

```tsx
<ConditionalFeature name="workflows-pilot">
  <NavLink href={getConfigValue('workflowsPilotUrl') as string}>Workflows</NavLink>
</ConditionalFeature>
```

components/NewPersonView/Layout.tsx
```diff
@@ -67,6 +68,36 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
...
+        )}/workflows/new?social_care_id=${person.id}`}
+        className="lbh-link lbh-body-s lbh-link--no-visited-state"
+      >
+        Start workflow
+      </a>
+    );
+  } catch (error) {
+    workflowsNavLink = null;
+    startWorkflowNavLink = null;
+  }
```
chrishutchinson, line 99: Same comment about the `try`/`catch`, this feels wrong. Maybe worth discussing so I can understand what you were aiming for?

pages/_app.tsx
```diff
@@ -109,9 +114,16 @@ CustomApp.getInitialProps = async (
...
+  const workflowsPilotUrl = process.env.WORKFLOWS_PILOT_URL;
+
   return {
     ...appProps,
-    pageProps: { ...appProps.pageProps, user, environmentName },
+    pageProps: {
+      ...appProps.pageProps,
+      user,
+      environmentName,
+      workflowsPilotUrl,
```
chrishutchinson, line 125: Instead of passing the individual values here, I'd probably create the `appConfig` object in advanced, and pass that through instead. Something like:

```ts
const appConfig = {
  workflowsPilotUrl: process.env.WORKFLOWS_PILOT_URL
};

return {
  ...appProps,
  pageProps: {
    // ...
    appConfig
  }
};
```

components/Dashboard/DashboardWrapper.tsx
chubberlisk: We decided it's fine as is as we don't gain too much from it in this case.

pages/_app.tsx
```diff
@@ -109,9 +112,18 @@ CustomApp.getInitialProps = async (
     ? 'development'
     : 'production';
 
+  const appConfig = {
```
chrishutchinson, line 115: It feels like it's probably worth a comment here warning that the values put in this object are passed to the client app, and therefore are publicly available in the browser. Maybe:

```suggestion
  // Warning! Any values placed in this object are passed to the client app, and will be made public in the browser. Do not share any sensitive or secret  values through this.
  const appConfig = {
```

lib/appConfig.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const useAppConfig: () => {
+  getConfigValue: (identifier: string) => string | number;
+} = () => {
+  const appConfig = useContext(AppConfigContext);
+
+  if (appConfig === undefined)
+    throw new Error(
+      'A <AppConfigProvider /> must be provided as a parent of this component'
+    );
```
chrishutchinson, line 29: Nit pick, but we should use braces! It's too easy to indent incorrectly and this goes horribly wrong

```suggestion
  if (appConfig === undefined) {
    throw new Error(
      'A <AppConfigProvider /> must be provided as a parent of this component'
    );
  }
```

lib/appConfig.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return {
+    getConfigValue: (identifier) => {
+      const configValue = appConfig[identifier];
+
+      if (configValue === undefined)
+        throw new Error(`A value for ${identifier} is not defined`);
```
chrishutchinson, line 36: Nit pick, could we clarify the error a little more?

```suggestion
        throw new Error(`A value for ${identifier} is not defined in the app config`);
```

pages/_app.tsx
chrishutchinson: Probably best to add a similar warning to the docs you've written too!

chrishutchinson (changes requested):
Docs are fantastic, nice work. Just a few other small things and we're good to go

PR #729, opened 2021-10-28
feat(SCT-1393): Added check for date picker so it can't be an invalid date + UI fix

mapleleafjack:
**What**  

- Set default date picker value when editing case status and updating a child situation.
- Added `currentCaseStatusStartDate` parameter in the URL of the flow
- Updated the `CaseStatusDetails` component so it shows the correct start/end date range
- Added and adjusted unit tests

**Why**  
The date picker was not being set correctly (as in value and validation) when editing and updating a case status.
The main Case Status widget was not showing the correct start/end date.


components/CaseStatus/CaseStatusDetailsTable.tsx
```diff
@@ -9,17 +9,141 @@ interface Props {
   styleType?: string;
   status: CaseStatus;
   answers?: CaseStatusAnswerDisplay;
+  sheduledStatus?: CaseStatusAnswerDisplay[];
```
rachel-hutchinson, line 12: This seems to be misspelt, might be worth fixing but not a blocker

mapleleafjack: Fixed :)

components/CaseStatus/CaseStatusDetailsTable.spec.tsx
```diff
@@ -44,6 +44,23 @@ describe('CaseStatusDetailsTable component', () => {
...
+          startDate: '2021-09-09',
+          endDate: '2022-09-09',
+          answers: [],
+        })}
+      />
+    );
+
+    const elements = getByTestId('start_end_date');
+    expect(elements).not.toBeNull();
+    expect(queryByText('09 Sept 2022')).toBeInTheDocument;
```
rachel-hutchinson, line 62: Does this need to check for 'Sep' or 'Sept'?

mapleleafjack: Hi Rachel, it depends if this is ran on a Unix or a Windows machine :) 

I've made them for Unix so this should pass (let me know in case it doesn't on your environment!)

components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
```diff
@@ -61,6 +75,50 @@ const EditCaseStatusForm: React.FC<{
...
+            'yyyy-MM-dd'
+          );
+        }
+      });
+    }
+  }
+
+  if (caseStatusType == 'LAC' && action == 'edit') {
+    if (pastCaseStatusStartDate && pastCaseStatusStartDate !== 'undefined') {
+      const caseStatusStartDate = new Date(pastCaseStatusStartDate);
```
rachel-hutchinson, line 109: Would it make more sense to call this const pastStatusStartDate?

components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
```diff
@@ -61,6 +75,50 @@ const EditCaseStatusForm: React.FC<{
...
+      caseStatusStartDate.setDate(caseStatusStartDate.getDate() + 1);
+
+      form_fields.map((field: any) => {
+        if (field.id === 'startDate') {
+          field.startDate = format(new Date(caseStatusStartDate), 'yyyy-MM-dd');
+
+          field.default = format(new Date(caseStatusStartDate), 'yyyy-MM-dd');
```
rachel-hutchinson, line 116: I think it might make more sense to the user if the default was currentCaseStatusStartDate

mapleleafjack: Good one!
Changed the default to +1 day

PR #730, opened 2021-10-29
feat(SCT-1381): add new CSF roles to date/roles.ts

tony-griffin:
**What**  

Added  **Independent Chair** & **Practice Development Manager** to the CSFRoles array.

These roles are used when creating or editing a worker.


rachel-hutchinson (approved):
Looks good to me

PR #732, opened 2021-11-02
fix(SCT-1362): Case Status error handling in UI

mapleleafjack:
**What**  
Next.JS API routes now returns correctly Case Status error messages sent from the backend to the frontend.
The error definitions are passed from the API routes to the try/catch block when calling the endpoint.

**Why**  
Before this it wasn't clear what the error was when an operation went wrong - all that was displayed was the error number, which doesn't make much sense to the end user.


tony-griffin (approved):
LGTM

tony-griffin:
Rachel and I were wondering if it is worth doing a test for a response error from the API for ReviewAddCaseStatusForm?

PR #733, opened 2021-11-02
fix: Case Status failing test

mapleleafjack:
**What**  
FIxed the non-working test

**Why**  
The async function that fired the event for the test was not behaving properly:
It was running correctly for a while both in circleci and in the local environment, then it suddenly stopped working.

Running the code in a synchronous way makes the test run correctly.

**Anything else?**


martapederiva (approved):
LGTM

PR #734, opened 2021-11-02
feat(SCT-1030): Mash dashboard 

rs2416:
Changes include the following:

- turning the bulk of team assignments code into a component called 'MashDashboard.tsx' so it is easier to create tests for.
- a component called 'MainCard.tsx' was created that allows the right card (per stage e.g. initial decision etc.) to be conditionally rendered.
- created a component per cards per tab/stage (e.g. initial decision).


components/MashCards/MainCard.tsx
```diff
@@ -0,0 +1,80 @@
+import { MashReferral } from 'types';
+import ContactCard from './ContactCard';
+import InitialDecisionCard from './InitialDecisionCard';
+import ScreeningCard from './ScreeningCard';
+import FinalDecisionCard from './FinalDecisionCard';
+
+interface Props {
+  filter: string;
+  mashReferrals: MashReferral | any;
```
JohnFarrellDev, line 9: why the | any?

components/MashCards/MainCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+const MainCard = ({ filter, mashReferrals }: Props): React.ReactElement => {
```
JohnFarrellDev, line 12: Just need to write some tests for this.
I'm happy to take this on.
Can do as a future PR.

components/MashCards/MainCard.tsx
JohnFarrellDev: ah I know why, it should be MashReferral[] not MashReferral.

That will mean we are stating the type of mashReferrals is an array of type MashReferral.
It needs to be an array so we can loop over it (the .map call inside the JSX).
I expect the | any got rid of this error but in general we want to avoid | any. Would be easier to explain on a call if you want an explanation for why we want to avoid any.

pages/team-assignments.tsx
```diff
@@ -1,75 +1,20 @@
...
 import { GetServerSideProps } from 'next';
 import { isAuthorised } from 'utils/auth';
-import { MashReferral } from 'types';
 import { getAllMashReferrals } from 'lib/mashReferral';
+import MashDashboard from 'components/MashDashboard/MashDashboard';
+import { MashReferral } from 'types';
 
 interface Props {
-  referrals: MashReferral[];
+  referrals: [];
```
JohnFarrellDev, line 9: this is just saying that referrals is an array, we're not saying that is is an array of MashReferrals.

We also don't do anything with the mashReferral prop?

JohnFarrellDev:
```
interface Props {
  referrals: MashReferral[];
}
```

would make more sense

components/MashDashboard/MashDashboard.tsx
```diff
@@ -0,0 +1,60 @@
+import st from 'components/Tabs/Tabs.module.scss';
```
JohnFarrellDev, line 1: This is good, just making a note for some tests that are required

pages/team-assignments.tsx
```diff
@@ -1,75 +1,20 @@
...
+  mashReferral: MashReferral[];
 }
 
 export const TeamAssignments = ({ referrals }: Props): React.ReactElement => {
```
JohnFarrellDev, line 13: we can drop this export, we already have a default export at the bottom of the file

components/MashCards/ScreeningCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <li className={s.row}>
+        <div>
+          <p className="lbh-body-m govuk-!-margin-bottom-3">
+            <span className="govuk-tag lbh-tag lbh-tag--green">
+              {timeleft} left
+            </span>{' '}
+            {datetime}
+            <span
+              className="lbh-body-l lbh-!-font-weight-bold"
+              style={{ float: 'right', display: 'block' }}
```
JohnFarrellDev, line 28: move in-line style into sensible css classes

components/MashCards/ScreeningCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+            >
+              <a href="#">Action </a>
```
JohnFarrellDev, line 30: Instead of a tags let's use Link components

components/MashCards/MainCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  if (filter === 'contact') {
+    return (
+      <div>
+        {mashReferrals.map((referral: any) => (
```
JohnFarrellDev, line 16: for all of these maps we will get rid of the :any once the any is removed from the incoming prop.

Typescript should then correctly infer that the referral is of type MashReferral

components/MashCards/InitialDecisionCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <li className={s.row}>
+        <div>
+          <p className="lbh-body-m govuk-!-margin-bottom-3">
+            <span className="govuk-tag lbh-tag lbh-tag--green">
+              {timeleft} left
+            </span>{' '}
+            {datetime}
+            <span
+              className="lbh-body-l lbh-!-font-weight-bold"
+              style={{ float: 'right', display: 'block' }}
```
JohnFarrellDev, line 29: Like other comment move this into css class

components/MashCards/InitialDecisionCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+            >
+              <a href="#">Action </a>
+            </span>
+          </p>
+          <hr className={s.line} />
+          <dl className={s.stats}>
+            <div>
+              <dt>Name of client</dt>
+              <dd>
+                <a href="#">{clientname} (referral)</a>
```
JohnFarrellDev, line 39: Like other comment utilise Link component

components/MashCards/FinalDecisionCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+            >
+              <a href="#">Action </a>
+            </span>
+          </p>
+          <hr className={s.line} />
+          <dl className={s.stats}>
+            <div>
+              <dt>Name of client</dt>
+              <dd>
+                <a href="#">{clientname} (referral)</a>
```
JohnFarrellDev, line 38: Like other comment utilise Link component

components/MashCards/ContactCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+            >
+              <a href="#">Action </a>
+            </span>
+          </p>
+          <hr className={s.line} />
+          <dl className={s.stats}>
+            <div>
+              <dt>Name of client</dt>
+              <dd>
+                <a href="#">{clientName} (referral)</a>
```
JohnFarrellDev, line 38: Like other comment utilise Link component

components/MashCards/ContactCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+  referrer: string;
+  requestedSupport: string;
+}
+const ContactCard = ({
+  clientName,
+  timeLeft,
+  dateTime,
+  referrer,
+  requestedSupport,
+}: Props): React.ReactElement | null => {
```
JohnFarrellDev, line 16: Why are we saying this can return null? To me it seems that it would always return a React.ReactElement.

Same applies to our other card components.

rs2416: Not actually sure tbh - it might have happened if I accidentally hit tab?

What does ' | null ' mean?

components/MashCards/MainCard.tsx
rs2416: Sounds good!

components/MashCards/ContactCard.tsx
JohnFarrellDev: | represents or. so : React.ReactElement | null means we are stating function returns a react component or null.

However, looking at the code I think it only ever returns a react component so we can remove the | null.

components/MashCards/MainCard.tsx
rs2416: For some reason when I remove 'any' on the second map then error messages arise so I need to investigate this further

components/MashCards/InitialDecisionCard.tsx
rs2416: Turned this into a class called 'action' in 'MashCard.module.scss'

components/MashCards/ScreeningCard.tsx
rs2416: Turned this into a class called 'action' in 'MashCard.module.scss'

components/MashCards/FinalDecisionCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <li className={s.row}>
+        <div>
+          <p className="lbh-body-m govuk-!-margin-bottom-3">
+            <span className="govuk-tag lbh-tag lbh-tag--green">
+              {timeleft} left
+            </span>{' '}
+            {datetime}
+            <span
+              className="lbh-body-l lbh-!-font-weight-bold"
+              style={{ float: 'right', display: 'block' }}
```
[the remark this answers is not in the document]
rs2416: Turned this into a class called 'action' in 'MashCard.module.scss'

components/MashCards/ContactCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <li className={s.row}>
+        <div>
+          <p className="lbh-body-m govuk-!-margin-bottom-3">
+            <span className="govuk-tag lbh-tag lbh-tag--green">
+              {timeLeft} left
+            </span>{' '}
+            {dateTime}
+            <span
+              className="lbh-body-l lbh-!-font-weight-bold"
+              style={{ float: 'right', display: 'block' }}
```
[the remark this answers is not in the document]
rs2416: Turned this into a class called 'action' in 'MashCard.module.scss'

JohnFarrellDev (approved):
:partying_face:

PR #737, opened 2021-11-03
feat(SCT-1393): Date picker validation

mapleleafjack:
**What**  
Date pickers during the flow don't have correct validation:
For LAC for example you can set a end date before the start date.
Also when editing a LAC case status, the date field should be prefilled with the active case status values (as it's the one that's been currently edited)

**Why**  
Date pickers during the flow don't have correct validation


components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
```diff
@@ -50,6 +60,10 @@ const EditCaseStatusForm: React.FC<{
           }
           if (field.id === 'endDate') {
             field.startDate = format(new Date(status.startDate), 'yyyy-MM-dd');
+
+            if (status.endDate) {
```
tony-griffin, line 64: Just wondering, would this code ever be triggered considering that when we end a case status, then the case status wouldn't be shown on the Details page. 
We wouldn't be able to get to the Edit page in this case.

components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
```diff
@@ -61,6 +75,52 @@ const EditCaseStatusForm: React.FC<{
...
+          field.default = format(
+            new Date(currentStartDateValidation),
+            'yyyy-MM-dd'
+          );
+        }
+      });
+    }
+  }
+
+  if (caseStatusType == 'LAC' && action == 'edit') {
```
tony-griffin, line 107: Rachel and I were wondering that if in the scenario **where we have a current case status & a past case status:**
if the validation date should be the past case status start date + 1
& also
If the date picker default date should be the current case status start date, as this may be clearer that we are editing this date value by doing this action.

```
if (caseStatusType == 'LAC' && action == 'edit') {
    if (pastCaseStatusStartDate && pastCaseStatusStartDate !== 'undefined' && currentCaseStatusStartDate? && currentCaseStatusStartDate !== 'undefined') {
      const defaultStartDate = new Date(currentCaseStatusStartDate);
      const pastStatusStartDate = new Date(pastCaseStatusStartDate);

      pastStatusStartDate.setDate(pastStatusStartDate.getDate() + 1);

      form_fields.map((field: any) => {
        if (field.id === 'startDate') {
          field.startDate = format(pastStatusStartDate, 'yyyy-MM-dd');

          field.default = format(defaultStartDate, 'yyyy-MM-dd');
        }
      });
    }
  }
```

 And then in the scenario **where we have a current case status only:**
 
```
const defaultStartDate = new Date(currentCaseStatusStartDate);
 field.default = format(defaultStartDate, 'yyyy-MM-dd');
```

components/CaseStatus/CaseStatusDetails.spec.tsx
```diff
@@ -124,7 +124,7 @@ describe('CaseStatusDetail component', () => {
             mockedStatusField.build({
               option: 'category',
               value: 'C2',
-              startDate: '2021-10-10',
+              startDate: '2021-09-09',
```
tony-griffin, line 127: Does this file belong in the PR for the UI date changes?

tony-griffin (commented):
Just a few comments around the Edit functionality and validation dates

components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
mapleleafjack: Hey Tony, this code in particular is to set the default value for the end date picker!

It won't change much if not setting the pre-filled value.

components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
mapleleafjack: I don't think it should be the past case status start date +1, as you may want to edit only the two dropdowns while keeping the same date (so no +1 involved)

I'll update anyways the code to have as default the case status start date, sounds good to me :)

tony-griffin (approved):

PR #738, opened 2021-11-03
feat(SCT-1433): Case status start/end date UI

mapleleafjack:
**What**  

- Adjusted the visualisation of Case Status (answers) start and end date
- Added the episode reason to the UI in case it's present

**Why**  
The UI does not always shows the correct combination of start/end date
This pr tidies up the code that displays the start date and end date of each case status (+ answers in case of LAC) so they can have the correct date displayed (and reason for episode ending for LAC) in the UI.


rachel-hutchinson (approved):
Reviewed with Tony, we're both happy after making some logic tweaks

PR #741, opened 2021-11-04
feat(SCT-1378): add EDIT announcement for changing case status start date consequences

tony-griffin:
**What**  
If changing the start date of the current case status, will affect the end date of the previous case status. 
Adding an Announcement box to alert the user of this.

**Why**  
This announcement box is to make this fact more explicit to the user.

**Anything else?**
Separate but related, I added an E2E Cypress test for displaying an announcement box on the LAC Update page when making a scheduled case status and there is a pre-existing scheduled case status

### DIFF
--- a/pages/people/[id]/case-status/[caseStatusId]/edit/edit.tsx
+++ b/pages/people/[id]/case-status/[caseStatusId]/edit/edit.tsx
@@ -34,12 +34,14 @@ const EditCaseStatus = (): React.ReactElement => {
 
   return (
     <>
-      {pastCaseStatusStartDate && action == 'edit' && (
-        <AnnouncementMessage
-          title="Changes made here may affect previous status items"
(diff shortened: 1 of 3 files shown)


pages/people/[id]/case-status/[caseStatusId]/edit/edit.tsx
```diff
@@ -34,7 +34,13 @@ const EditCaseStatus = (): React.ReactElement => {
 
   return (
     <>
-      {isScheduledCaseStatus && (
+      {pastCaseStatusStartDate && action == 'edit' && (
```
Comment (rachel-hutchinson) [inline pages/people/[id]/case-status/[caseStatusId]/edit/edit.tsx:37]: Tiny suggestion to also add `&& pastCaseStatusStartDate !== 'undefined'` to make sure that pastCaseStatusStartDate isn't undefined

rachel-hutchinson (approved):
One small suggestion for improvement but approving as this is minor

PR #746, opened 2021-11-05
fix(SCT-1444): intial decision card to match designs

JohnFarrellDev:
**What**  
Minor changes to the initial decision card so it matches the designs from figma

https://www.figma.com/file/Mw7tNmwNUGzpZDBXMaGhAj/MASH?node-id=5%3A2

**Anything else?**

### DIFF
--- a/components/MashCards/InitialDecisionCard.tsx
+++ b/components/MashCards/InitialDecisionCard.tsx
@@ -29,9 +29,9 @@ const InitialDecisionCard = ({ mashReferral }: Props): React.ReactElement => {
               <dd>
                 <Link href={mashReferral.referralDocumentURI}>
                   <a>
-                    {mashReferral.clients[0]}
+                    {mashReferral.clients[0]} +{' '}
                     {mashReferral.clients.length > 1 &&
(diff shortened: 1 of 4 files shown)


components/MashCards/InitialDecisionCard.tsx
```diff
@@ -29,21 +29,21 @@ const InitialDecisionCard = ({ mashReferral }: Props): React.ReactElement => {
...
-              <dt>Screening decision</dt>
-              <dd>{mashReferral.screeningDecision}</dd>
+              <dt>Requested support</dt>
+              <dd>{mashReferral.requestedSupport}</dd>
             </div>
             <div>
-              <dt>Referral category</dt>
-              <dd>{mashReferral.referralCategory}</dd>
+              <dt></dt>
+              <dd></dd>
```
JohnFarrellDev, line 46: I don't like this but I'm fine with it for now, currently all the cards use the same css and utilise a grid template expecting 4 items, adding an empty item maintains alignment (assign button being all the way to the right)

rs2416 (approved):

PR #747, opened 2021-11-05
feat(SCT-1441): Importing the selector options for referral categories and decisions

rs2416:
**What**  
Imported the list of options for referral categories, initial decisions, screening decisions and final decisions for the action page selectors. As well as setting up the states for the final decision page.

**Why**  
So the options in the selectors are more representative and functional. 

**Anything else?**

- There is a blank option in the selector which will be worked on as a separate PR.


pages/mash-referral/[id]/final-decision.tsx
```diff
@@ -2,131 +2,146 @@ import Button from 'components/Button/Button';
...
+                    className="govuk-label govuk-radios__label"
+                    htmlFor="yes-input"
+                  >
+                    Yes
+                  </label>
+                </div>
+                <div className="govuk-radios__conditional" id="hint-email">
+                  <label className="govuk-label" htmlFor="hint">
+                    Please email your MASH manager about the urgent case.
+                  </label>
```
Comment (JohnFarrellDev) [inline pages/mash-referral/[id]/final-decision.tsx:117]: Make this dynamically appear only when urgent is true

JohnFarrellDev (approved):
looks great :+1:

PR #749, opened 2021-11-07
feat(SCT-1431): Add case status validation

rachel-hutchinson:
**What**  
* Validation has been added to the 'add a case status' form to ensure that any new case status does not overlap with previously closed status' (in the screenshot below a previous case status ended on 11/01/2000 so it is not possible to start a new case status on 10/01/2000 as this would mean the two status' overlap
** To achieve this a flag has been added to the GET endpoint to confirm if the GET end point should return closed cases or not (be default closed cases are not returned)
* The same validation has been added to the edit case status screen to make sure a case status couldn't be edited to overlap with a previously closed status
* The validators file has been updated so that dates can be validated to ensure they are not before a specified 'start date' and also cannot be in the future
* The end to end tests have been updated so they now use a different child record when testing CIN, CP and LAC
* The end to end tests also now check that a case status can't be added on an adults record
* Before running CIN, CP or LAC tests an initial tests checks if there is already a case status, and if so it is ended
* End to end tests have also been added to check the new validation on the add a case status form as well 
* Some extra assertions have also been added to ensure the system is working as expected
* Some extra validation tests in general have been added to make sure the validation is correct for all case stauts actions

image

**Why**  

- It should not be allowed to have two case status' that overlap. This validation was added in the backend and this change was added so that users were warned this was not feasible before they tried to submit the form
- The end to end tests have been updated as they were causing issues when multiple end to end tests were run at the same time

**Anything else?**
The cypress json environment variables have been updated, these have also been updated in circleCI


components/CaseStatus/EditCaseStatusForm/EditCaseStatusForm.tsx
```diff
@@ -29,7 +30,13 @@ const EditCaseStatusForm: React.FC<{
...
 }) => {
   const router = useRouter();
-  const { data: caseStatuses } = useCaseStatuses(personId);
+  const { data: caseStatuses } = useCaseStatuses(personId, 'false');
+  const { data: caseStatusesIncClosed } = useCaseStatuses(personId, 'true');
+
+  const latestEndedStatusEndDate = getLatestEndedStatusEndDate(
+    caseStatusesIncClosed
+  );
+  console.log('latestEndDate', latestEndedStatusEndDate);
```
tony-griffin, line 39: console log to remove

components/CaseStatus/caseStatusHelper.ts
```diff
@@ -134,3 +134,28 @@ const groupAnswersByGroupId = (
...
+      (status) => status.endDate
+    );
+    if (endedCaseData && endedCaseData.length > 0) {
+      latestEndDate = endedCaseData[0].endDate;
+      endedCaseData.forEach((status) => {
+        if (Date.parse(status.endDate) > Date.parse(latestEndDate)) {
+          latestEndDate = status.endDate;
+        }
+      });
+      console.log('latestEndDate', latestEndDate);
```
tony-griffin, line 153: console log to remove

lib/validators.ts
```diff
@@ -101,6 +101,27 @@ export const generateFlexibleSchema = (
...
+      field.startDate &&
+      field.isfutureDateValid === false
+    ) {
+      shape[field.id] = Yup.string().test(
+        'Validate date is before date and past or present',
+        'Date cannot be before start date or in the future',
+
+        (dateValue) => {
+          const dateTimeToValidate = new Date(String(dateValue));
+          console.log(
```
tony-griffin, line 115: console log to remove

tony-griffin (approved):
Looks great, just a few console logs to remove

PR #753, opened 2021-11-08
feat(SCT-1446): Tests for the MASH cards and Headings

rs2416:
**What**  
Created 6 tests to check the correct text renders on the four different types of MASH cards, as well as rendering the correct card based on the tab selected on the Mash dashboard and assessing if the MashHeading is rendered properly. The test files can be found here on this path 'components/MashCards/' and 'components/MashHeading'. 

**Why**  
In order to make sure we are pushing bug free code.


JohnFarrellDev (approved):
:+1:

components/MashHeading/Heading.spec.tsx
```diff
@@ -0,0 +1,26 @@
...
+describe('Heading', () => {
+  it('renders a title', () => {
+    render(
+      <Heading
+        clientname={mockedMashReferral.clients[0]}
+        timeleft="4 hours"
+        datetime={mockedMashReferral.createdAt}
+      />
+    );
+    expect(screen.getByText('4 hours'));
```
JohnFarrellDev, line 15: 4 hours left

components/MashCards/InitialDecisionCard.spec.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { mockedMashReferral } from 'factories/mashReferral';
+import InitialDecisionCard from './InitialDecisionCard';
+
+describe('InitialDecisionCard', () => {
+  it('renders the right info from the mash referral', () => {
+    render(<InitialDecisionCard mashReferral={mockedMashReferral} />);
+
+    expect(screen.getByText('4 hours left'));
+    expect(screen.getByText(`${mockedMashReferral.clients[0]} (referral)`));
+    expect(screen.getByText(mockedMashReferral.screeningDecision as string));
```
JohnFarrellDev, line 11: an initial decision (and by extension initial decision card) does not have a screening decision.

It should have a requested support that we want to check for.

JohnFarrellDev (changes requested):

components/MashCards/InitialDecisionCard.spec.tsx
rs2416: Good point - didn't see this!

JohnFarrellDev (approved):
looks good

PR #760, opened 2021-11-11
feat(SCT-1454):  e2e test updates

tony-griffin:
**What**  
Attempting to make e2e test more robust by:
- splitting out the case status types into their own cypress files
- adding a check at the start of each test file for an existing case status for that type. If one is found we create a new resident and use that personId to run the tests with.

**Why**  
As the e2e tests are run as part of CI/CD pipeline then we have encountered problems when the tests are running in parallel. This is because they use a personId per case status type & should developer A be running tests at the same time as developer B, then developer B may encounter tests failures.
This solution should assign a unique personId to run the tests with if its detected that the default personId value for that case status type is already in use.


mapleleafjack (approved):
🚀

PR #762, opened 2021-11-12
feat(SCT-1437): Making time tags dynamic

rs2416:
**What**  
This code turns relates to the time tag component used on the MASH designs. The following code makes the time tags (that appear on the referral cards and action pages) dynamic by making them based on the time the referral was received (mashReferral.createdAt) and also based on things such as what stage the referral is at alongside what the rag rating of the referral is (green - 72 hours, amber - 24 hours and red - 4 hours).

**Why**  
To allow the MASH team to prioritise cases based on the types of cases they have and the statutory obligations.


components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,90 @@
+import moment from 'moment';
+
+interface Props {
+  filter: string;
+  createdAt: string;
+  initialDecision: string | undefined;
+}
```
JohnFarrellDev, line 7: Could we just pass in the whole referral (MashReferral object)?

We don't need filter because MashReferral has a property stage

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,90 @@
...
+}: Props): React.ReactElement => {
+  const currenttime = new Date().toISOString();
+  const hoursPassed = moment(currenttime, 'YYYY-MM-DD[T]HH:mm:ss. SSS[Z]').diff(
+    moment(createdAt, 'YYYY-MM-DD[T]HH:mm:ss. SSS[Z]'),
+    'hours'
+  );
+  const minsPassed = moment(currenttime, 'YYYY-MM-DD[T]HH:mm:ss. SSS[Z]').diff(
+    moment(createdAt, 'YYYY-MM-DD[T]HH:mm:ss. SSS[Z]'),
+    'minutes'
+  );
```
JohnFarrellDev, line 22:
```
const currentTime = new Date();
  const createdAtTime = new Date(createdAt);

  const hoursPassed = currentTime.getHours() - createdAtTime.getHours();
  const minsPassed = currentTime.getMinutes() - createdAtTime.getMinutes();
```

I'm actually slightly confused how using moment was working considering we don't have it as an installed dependency

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,90 @@
...
+            {redTimer * 60} mins ago
+          </div>
+        );
+      if (4 > hoursPassed)
+        return (
+          <div className="govuk-tag lbh-tag lbh-tag--green">
+            {redTimer} {redTimer == 1 ? 'hour' : 'hours'} left
+          </div>
+        );
+    }
```
JohnFarrellDev, line 85: I think it might be easier if we hop on a call to talk about this if-else block and ways we could make it neater

components/MashHeading/Heading.tsx
```diff
@@ -1,26 +1,33 @@
...
+import MashTag from 'components/MashTags/MashTags';
 
 interface Props {
   clientname: string;
-  timeleft: string;
-  datetime: string;
+  createdAt: string;
+  filter: string;
+  initialDecision: string | undefined;
 }
```
JohnFarrellDev, line 10: easier to just pass the whole referral (MashReferral object)

components/MashTags/MashTags.tsx
JohnFarrellDev: In my opinion calling moment and having it parse iso formatted date strings is adding complexity that we don't need to find out how many hours/minutes have passed. (explanation for the above comment with suggested code change.

components/MashForms/ScreeningDecisionForm.tsx
```diff
@@ -52,8 +54,9 @@ const ScreeningDecisionForm = ({
       <h1>Make screening decision</h1>
       <Heading
         clientname={referral.clients.join(', ')}
-        timeleft="3 hours"
-        datetime={referral.createdAt}
+        filter={filter}
+        createdAt={referral.createdAt}
+        initialDecision={referral.initialDecision}
       />
```
JohnFarrellDev, line 60: we can just pass the whole MashReferral object, let the component then decide how it wants to format clientName/s for example

JohnFarrellDev (changes requested):
Mostly relatively simple changes but we should talk through the MashTags component and see if we can find a way to reduce the if-else blocks required

components/MashCards/ContactCard.tsx
```diff
@@ -2,20 +2,24 @@ import s from './MashCard.module.scss';
...
         <div>
           <p className={`lbh-body-s govuk-!-margin-bottom-3 ${s.datetime}`}>
-            <span className="govuk-tag lbh-tag lbh-tag--green">
-              4 hours left
-            </span>{' '}
+            <MashTag
+              filter={filter}
+              createdAt={mashReferral.createdAt}
+              initialDecision={mashReferral.initialDecision}
+            />{' '}
```
JohnFarrellDev, line 22: We should have margin here instead of empty space if we are using this for spacing (styling purposes)

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+    output(mashReferral);
+
+  if (mashReferral.stage === ReferralStage.CONTACT) {
+    if (minsPassed < 60)
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {minsPassed} mins ago
+        </div>
+      );
+    if (hoursPassed > 1 || hoursPassed == 1)
```
JohnFarrellDev, line 83: watch out for ==

Also this could just an else block, if minsPassed is not less than 60 then an hour or more must have passed, no need for the conditional

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {hoursPassed} {hoursPassed == 1 ? 'hour' : 'hours'} ago
+        </div>
+      );
+  }
+  if (isOverdue != true) {
+    if (
+      timeLeftinMilliseconds / 60 / 1000 < 1 &&
+      timeLeftinMilliseconds / 60 / 1000 > 0
```
JohnFarrellDev, line 93: might be nice to have explicit function that converts our ms value to hours.

Also is the `&& timeLeftinMilliseconds / 60 / 1000 > 0` required? If isOvdue is false than timeleft must be greater than 0.

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </div>
+      );
+    if (hoursPassed > 1 || hoursPassed == 1)
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {hoursPassed} {hoursPassed == 1 ? 'hour' : 'hours'} ago
+        </div>
+      );
+  }
+  if (isOverdue != true) {
```
JohnFarrellDev, line 90: !==

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+    if (minsPassed < 60)
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {minsPassed} mins ago
+        </div>
+      );
+    if (hoursPassed > 1 || hoursPassed == 1)
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {hoursPassed} {hoursPassed == 1 ? 'hour' : 'hours'} ago
```
JohnFarrellDev, line 86: ===

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {timeLeftinMilliseconds / 60 / 1000} mins ago
+        </div>
+      );
+    if (timeLeftinMilliseconds > 0)
+      return (
+        <div className="govuk-tag lbh-tag lbh-tag--green">
+          {timeLeftinMilliseconds / 60 / 60 / 1000}
+          {timeLeftinMilliseconds / 60 / 60 / 1000 == 1 ? 'hour' : 'hours'} left
```
JohnFarrellDev, line 104: ===

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+const initialLowRatings = new Set([
+  'E3 REFERRAL',
+  'EH SCREENING REQUIRED IN MASH',
+]);
+
+const initialMedRatings = new Set(['CSC SCREENING REQUIRED IN MASH']);
+
+const initialHighRatings = new Set(['PROGRESS STRAIGHT TO CSC ALLOCATION']);
+
+const output = (mashReferral: MashReferral) => {
```
JohnFarrellDev, line 20: more descriptive name required, what does the function do, what does it return?

components/MashTags/MashTags.tsx
```diff
@@ -0,0 +1,111 @@
...
+  if (
+    mashReferral.stage === ReferralStage.SCREENING ||
+    mashReferral.stage === ReferralStage.FINAL
+  ) {
+    const initialhoursPassed = moment(
+      currentTime,
+      'YYYY-MM-DD[T]HH:mm:ss. SSS[Z]'
+    ).diff(
+      moment(mashReferral.initialCreatedAt, 'YYYY-MM-DD[T]HH:mm:ss. SSS[Z]'),
+      'hours'
```
JohnFarrellDev, line 41: could we do something like 

```
convertMsToHours(Number(currentTime) - Number(new Date(mashRefeferral.initialCreatedAt))
```

JohnFarrellDev: apply same logic to 2 other places we use moment

components/MashTags/MashTags.tsx
rs2416: Changes it to isOverdue === false

JohnFarrellDev (approved):
:+1:

PR #763, opened 2021-11-15
feat(SCT-1469): "Multiple" type + error handling

mapleleafjack:
**What**  
Added support for type `Multiple` for case status type CP
Added error handling in the FE for failed operations

**Why**  
The data imported from Mosaic contained this case status type for CP, we need to support it.
Some bits of the FE were not receiving the correct error message to be handled.


rachel-hutchinson (approved):
Looks good to me

PR #764, opened 2021-11-17
feat: workflow pilot on timeline

jhackett1:
modify timeline events to correctly show workflows:
- detect whether a case represents a workflow by looking for a `workflowId` string in `caseFormData`
- use a link `href` that correctly links to the pilot app
- show the correct name of the assessment/form type

Capture

### DIFF
--- a/components/NewPersonView/EventLink.tsx
+++ b/components/NewPersonView/EventLink.tsx
@@ -28,7 +28,7 @@ const EventLink = ({ event }: Props): React.ReactElement => {
   }
 
   // 2. handle workflows
-  if (event.caseFormData.workflowId)
+  if (event?.caseFormData?.workflowId)
     return (
       <a
         href={`${getConfigValue('workflowsPilotUrl') as string}/workflows/${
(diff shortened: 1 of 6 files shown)


components/NewPersonView/EventLink.tsx
```diff
@@ -4,13 +4,14 @@ import { Case } from 'types';
 import { generateInternalLink as generateLegacyUrl } from 'utils/urls';
 import s from './index.module.scss';
 
+const workflowsPilotUrl = `https://workflows-social-care-service.hackney.gov.uk`;
```
chubberlisk, line 7: Instead of hardcoding the URL for each environment we should use the hook we created for this, see https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/82cad6b9e00359b61e5a8f82a661691e7df677d3/components/Dashboard/DashboardWrapper.tsx#L68 as an example for the code and https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/82cad6b9e00359b61e5a8f82a661691e7df677d3/components/Dashboard/DashboardWrapper.spec.tsx#L37-L61 for test example.

acodeninja (approved):
Looks good to me 👍, great to see some good tests in there. The only feedback I'd give is to make future commit messages a little more descriptive.

PR #766, opened 2021-11-17
feat(SCT-1500): screening decision loading and error state

JohnFarrellDev:
**What**  
When submitting a screening decision the submit button is now disabled.
If the API returns an error the error message is shown to the user.
We now have tests for ScreeningDecisionForm.

**Why**  
Tests for code reliability.
Rest to improve UX.

**Anything else?**


components/MashForms/ScreeningDecisionForm.spec.tsx
```diff
@@ -0,0 +1,98 @@
...
+    ).toBeNull();
+
+    fireEvent.click(screen.getByText('Yes'));
+
+    expect(
+      screen.getByText('Please email your MASH manager about the urgent case.')
+    );
+  });
+
+  it('should disable submit button when submitScreeningDecision Submit button is clicked', () => {
```
Renny4Real, line 50: Does mean that the button is disabled after it is clicked? Why do we need to do this?

Renny4Real (approved):
Looks good. Just a question. 🎉

components/MashForms/ScreeningDecisionForm.spec.tsx
JohnFarrellDev: We don't want a user clicking the button multiple times sending multiple API calls. 

It also indicates to the user that their submission is being sent as disabling the button modifies it's appearance.

Before a user could click, see no indication that clicking did anything,  click again. It wouldn't break anything but this is better UX and stops the API potentially getting spammed.

Renny4Real: I see 🙌🏾 . Do we show a notification/banner on the page? Might also help the user understand what's going on too?

PR #767, opened 2021-11-17
feat(SCT-1505): final decision form to submit answers to api

JohnFarrellDev:
**What**  
Pass in required props to our FinalDecisionForm
Hitting submit on final decision form now submits answers to the API. API is not setup to handle the incoming data yet and update the referral.
Handle error and submitting state.

**Why**  
The form actually needs to be able to submit data to the backend...

**Anything else?**

### DIFF
--- a/components/MashForms/FinalDecisionForm.spec.tsx
+++ b/components/MashForms/FinalDecisionForm.spec.tsx
@@ -1,0 +1,102 @@
+import { render, fireEvent, screen, waitFor } from '@testing-library/react';
+import { mockedMashReferral } from 'factories/mashReferral';
+import { submitFinalDecision } from 'utils/api/mashReferrals';
+import { mockedWorker } from 'factories/workers';
+import FinalDecisionForm from './FinalDecisionForm';
+
+jest.mock('utils/api/mashReferrals');
+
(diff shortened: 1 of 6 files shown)


components/MashForms/FinalDecisionForm.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+    (submitFinalDecision as jest.Mock).mockResolvedValue(true);
+
+    fireEvent.click(screen.getByText('Submit'));
+
+    expect(submitFinalDecision).toBeCalledWith(
+      mockedMashReferral.id,
+      mockedWorker.email,
+      'NFA',
+      'Abuse linked to faith or belief',
+      false
```
Renny4Real, line 69: Where do these values come from in the test?

Renny4Real (approved):
🎉

components/MashForms/FinalDecisionForm.spec.tsx
JohnFarrellDev: FinalDecisionForm keeps track of the form answers in useState, these answers are the useState defaulted values for our form fields.

PR #768, opened 2021-11-18
feat(SCT-1490): Feature flag user permissions update

rachel-hutchinson:
**What**  
An update has been made to the case-status feature flag to allow admin users to view case status
Additional tests have also been added to ensure this works

**Why**  
We need to import case status data into production and an admin needs to be able to view and check this data

**Anything else?**
The server-side function `isFeatureFlagActive` has been set up to have an undefined user by default
This should be updated to take in a user object.
For this PR this isn't possible as the backend doesn't have the user context. Additionally `isFeatureFlagActive` is currently only being used in tests within the codebase 
@chrishutchinson we might need more of your input for this but wanted to get the minimum requirements out to allow us to go live with case status'


mapleleafjack (approved):
Looks good to me!

PR #773, opened 2021-11-22
fix(SCT-1523): removed hardcoded URL from case status flag

mapleleafjack:
**What**  
The URL now uses a relative path so it doesn't matter if we're in dev/staging/prod, the link will always work

**Why**  
The Case Status flag had the development link hardcoded in the link, making it useless on production.


chubberlisk (approved):
Approving as I know that this is a bug in production now but typically I would expect a test change as well - something like maybe https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/fbfdba4e679914dc0b3a3719920085277c707855/components/NewPersonView/Layout.spec.tsx#L247-L250.

PR #776, opened 2021-11-29
feat(SCT-1498): Case notes deletion UI

mapleleafjack:
[Link to ticket](https://hackney.atlassian.net/browse/SCT-1498)

Everything is behind the feature flag "case-notes-deletion" (available only on development/staging as of now)

**What**  

- Added "delete note" link and deletion form in the case note view (only for Admin)
- Added "hide/show deleted records" filter in main timeline (right sidebar)
- Deleted records in the timeline will have "Deleted record" words in the title + will not have a link if the user is not an Admin
- Opening a deleted record via URL will redirect back to the timeline if the user is not an Admin
- Added confirmation banner when a case note has been deleted
- Added necessary API communication bits
- Updated factories with new parameters
- Added tests

**Why**  
To allow case notes deletion (only for Admin)

| Confirmation message                                         | Modal                                                        | Delete link                                                  | Deleted record                                               | Show/hide deleted records                                    |
| ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
| Screenshot 2021-11-29 at 14 08 25 | Screenshot 2021-11-29 at 14 08 05| Screenshot 2021-11-29 at 14 08 17| Screenshot 2021-11-29 at 14 07 36|  Screenshot 2021-11-29 at 14 07 30 |


components/NewPersonView/EventLink.tsx
```diff
@@ -17,14 +17,29 @@ const EventLink = ({ event }: Props): React.ReactElement => {
...
+    } else {
+      return (
+        <>
+          <span>
+            {formName} {formTitle}{' '}
+          </span>
+          <Link
+            href={`/people/${event.personId}/submissions/${event.recordId}`}
+          >
+            <a>(deleted record)</a>
```
rachel-hutchinson, line 38: Should this have the same class name as the existing code `className={`lbh-link ${s.eventLink}`}`?

mapleleafjack: Good shout, I'll add the classname!

components/NewPersonView/Layout.tsx
```diff
@@ -136,6 +141,17 @@ const Layout = ({ person, children }: Props): React.ReactElement => {
...
+        <AnnouncementMessage
+          title="Case note deleted"
+          content={`Case note "${String(
+            router.query.case_note_name
+          )}" has been deleted correctly`}
+        />
+      ) : (
+        <></>
+      )}
+
```
rachel-hutchinson, line 154: This looks good, there is also a `confirmation banner` I'm not sure if this would be better to use? Although I'm happy with either

components/Cases/CasesTable.tsx
```diff
@@ -70,21 +70,29 @@ const CaseNoteAction = ({
...
+        formName={formName}
+        personId={personId}
+        formType={formType}
+        title={title}
+        isImported={false}
+        deleted={false}
+      />
+    ) : (
+      ''
+    )}
```
rachel-hutchinson: We should show the link to the deleted note if the user is an admin user

rachel-hutchinson (commented):
Overall this looks good 
There is one change to add a link to the 'my-work' page for deleted notes for admin users
When I talked this over with Jack we also spotted that deleted notes will need to be filtered out on the 'my-work' page for non-admin users but we agreed this will be picked up separately on ticket SCT-1544

components/Cases/CasesTable.tsx
mapleleafjack: Now the My Work page should display the deleted case notes entries (together with the link) only if the user is an Admin (or dev)

rachel-hutchinson (approved):
Approved, changes look good

PR #777, opened 2021-11-29
feat(SCT-1530, SCT-1532): Case Status pre-launch flow fix UI

mapleleafjack:
[Link to ticket 1](https://hackney.atlassian.net/browse/SCT-1530)
[LInk to ticket 2](https://hackney.atlassian.net/browse/SCT-1532)

**What**  
- CP now can have multiple episodes in the same fashion as LAC
- Removed `X1` from the possible ending reason for LAC

**Why**  
CP needs to record episodes in the same way LAC does
X1 for LAC is not selectable by the user, but se automatically by the system when the child circumstance changes


rachel-hutchinson (approved):
Reviewed all changes and these look good
Is it worth waiting to deploy this until the backend changes go out to make sure things don't break on staging? (backend changes are on [this PR](https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/537) )

PR #779, opened 2021-11-30
feat(SCT-1448): Add tests for mash dashboard component

Miles-Alford:
**What**  
This PR adds a couple of tests for the `MashDashboard` component

**Why**  
To ensure the `MashDashboard` component works as intended

### DIFF
--- a/components/MashDashboard/MashDashboard.spec.tsx
+++ b/components/MashDashboard/MashDashboard.spec.tsx
@@ -1,0 +1,74 @@
+import { render, screen, fireEvent } from '@testing-library/react';
+import { mockedMashReferrals } from 'factories/mashReferral';
+import MashDashboard from './MashDashboard';
+
+let tabType: string | undefined;
+let confirmationType: string | undefined;
+const mockedRouter = {
+  query: { confirmation: confirmationType, tab: tabType },
+  push: jest.fn(),
+};
+
+jest.mock('next/router', () => ({
+  useRouter: () => mockedRouter,
+}));
+
+describe('MashDashboard', () => {
+  beforeEach(() => {
+    mockedRouter.query.tab = undefined;
+    mockedRouter.query.confirmation = undefined;
+  });
+
+  it('renders the dashboard default tab to be the contact tab', () => {
+    render(<MashDashboard referrals={mockedMashReferrals} />);
+
+    expect(screen.getByText(`Contact (0)`).parentElement?.className).toContain(
+      'active'
+    );
+  });
+
+  it('selects the correct tab if the url query is updated', () => {
+    mockedRouter.query.tab = 'initial-decision';
+    render(<MashDashboard referrals={mockedMashReferrals} />);
+
+    expect(
+      screen.getByText(`Initial decision (${mockedMashReferrals.length})`)
+        .parentElement?.className
+    ).toContain('active');
+  });
+
+  it('updates the selected tab when a tab is clicked on', async () => {
+    mockedRouter.query.confirmation =
+      '{"title":"test-title","test-key":"test-body"}';
+    render(<MashDashboard referrals={mockedMashReferrals} />);
+
+    expect(screen.getByText(`Contact (0)`).parentElement?.className).toContain(
+      'active'
+    );
+
+    fireEvent.click(
+      screen.getByText(`Initial decision (${mockedMashReferrals.length})`)
+    );
+
+    expect(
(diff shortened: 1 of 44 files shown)


components/MashDashboard/MashDashboard.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+      'active'
+    );
+  });
+
+  it('selects the correct tab if the url query is updated', () => {
+    mockedRouter.query.tab = 'initial-decision';
+    render(<MashDashboard referrals={mockedMashReferrals} />);
+
+    expect(
+      screen.getByText(`Initial decision (${mockedMashReferrals.length})`)
```
Renny4Real, line 35: We get the number from the mocked object here but hard code zero for the contacts. Does mockedMashReferrals contain only initial decision objects. I don't think it's very clear why we call `.length` here and then hard code `0` above.

Is it possible to create the number of contacts, initial decisions, etc in the test setup and then check their numbers match in the expectation?

components/MashDashboard/MashDashboard.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    fireEvent.click(
+      screen.getByText(`Initial decision (${mockedMashReferrals.length})`)
+    );
+
+    expect(
+      screen.getByText(`Initial decision (${mockedMashReferrals.length})`)
+        .parentElement?.className
+    ).toContain('active');
+    expect(mockedRouter.push).toHaveBeenCalledWith(
```
Renny4Real, line 57: Does this check that the tab switch was not done via the URL?

Renny4Real (commented):
Looks good 🎉 
Couple of questions. 🙏🏾

components/MashDashboard/MashDashboard.spec.tsx
Miles-Alford: Yeah, it checks that it changes through clicking a different tab rather then through the URL

components/MashDashboard/MashDashboard.spec.tsx
Miles-Alford: Added some setup for getting the correct number of referral for each tab in `f111abe` & `c8e8736`

components/MashDashboard/MashDashboard.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+  push: jest.fn(),
+};
+
+jest.mock('next/router', () => ({
+  useRouter: () => mockedRouter,
+}));
+
+describe('MashDashboard', () => {
+  const contactMashReferral = mockedMashReferral;
+  contactMashReferral.stage = ReferralStage.CONTACT;
```
JohnFarrellDev, line 22: could we call the factory method with .build here?

components/MashDashboard/MashDashboard.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  beforeEach(() => {
+    mockedRouter.query.tab = undefined;
+    mockedRouter.query.confirmation = undefined;
+  });
+
+  it('renders the dashboard default tab to be the contact tab', () => {
+    render(<MashDashboard referrals={mockedMashReferrals} />);
+
+    expect(contact.length).toBe(1);
```
JohnFarrellDev, line 41: I#m not sure what the purpose of this expectation is, doesn't seem to test what the component is rendering but the logic within our test component (line 24-31).

Same applies to initial.length on line 51.

components/MashDashboard/MashDashboard.spec.tsx
Miles-Alford: Done in 8b5297d

components/MashDashboard/MashDashboard.spec.tsx
Miles-Alford: Removed in 8b5297d

components/MashDashboard/MashDashboard.spec.tsx
Renny4Real: Do we still need line 24 since we've called the build function with the stage we want above?

Renny4Real (approved):
Looks good 🎉

components/MashDashboard/MashDashboard.spec.tsx
Renny4Real: Do we still need line 24 i.e `contactMashReferral.stage = ReferralStage.CONTACT;` since we've called the build function with the stage we want above?

PR #781, opened 2021-12-01
feat(SCT-1550): Case notes deletion: set include_deleted_count flag in the frontend based on the user

mapleleafjack:
**What**  
- Added `include_deleted_count`  flag in the request for case_notes (set to true only if the user is an Admin or Dev)
- Added count in the UI 
- Added type definition
- Added tests

**Why**  
We don't want to add overhead on the backend server, so the system will load the deleted rows count only if needed (if the user is an Admin or Dev)

Screenshot 2021-12-01 at 10 59 38

### DIFF
--- a/components/NewPersonView/PersonHistory.tsx
+++ b/components/NewPersonView/PersonHistory.tsx
@@ -43,8 +43,8 @@ const PersonHistory = ({ personId }: Props): React.ReactElement => {
   for (let i = 0; casesData !== undefined && i < casesData.length; i++) {
     events.push(...casesData[i].cases);
     if (includeDeletedCount) {
-      deletedCount = casesData[i].deletedCount
-        ? Number(casesData[i].deletedCount)
+      deletedCount = casesData[i].deletedRecordsCount
+        ? Number(casesData[i].deletedRecordsCount)
         : 0;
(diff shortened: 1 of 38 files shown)


components/NewPersonView/PersonHistory.tsx
```diff
@@ -31,9 +37,16 @@ const PersonHistory = ({ personId }: Props): React.ReactElement => {
...
   }
 
+  let deletedCount = 0;
+
   const events = [] as Case[];
   for (let i = 0; casesData !== undefined && i < casesData.length; i++) {
     events.push(...casesData[i].cases);
+    if (includeDeletedCount) {
+      deletedCount = casesData[i].deletedCount
+        ? Number(casesData[i].deletedCount)
```
rachel-hutchinson, line 47: I think based on [this PR](https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/538/files) the `deletedCount` will be returned as `DeletedRecordsCount` from the backend

mapleleafjack: updated!

rachel-hutchinson (approved):
PR looks good, approved

PR #783, opened 2021-12-01
fix: Case note parameter name change

mapleleafjack:
**What**  
This PR standardise two parameter sent to the /case endpoint

**Why**  
The parameter sent by the frontend are not standard.


rachel-hutchinson (approved):
Looks good

PR #784, opened 2021-12-01
feat(SCT-1348): Added description to deleted case notes in timeline

mapleleafjack:
[Link to ticket ](https://hackney.atlassian.net/browse/SCT-1348)

**What**  

- Added extra information to a deleted case notes in the timeline
- Added tests
- Enabled some Case Note deletion test that were disabled by mistake

**Why**  
We need to display the deletion extra information in the timeline (requested by / date / reason for deletion)

image

### DIFF
--- a/components/NewPersonView/Event.tsx
+++ b/components/NewPersonView/Event.tsx
@@ -48,15 +48,11 @@ const Event = ({ event }: Props): React.ReactElement => {
         {safelyFormat(displayDate)}
         {event.caseFormUrl?.includes('google') && ` · Google document`}
         {event.officerEmail && ` · ${event.officerEmail}`}
-        {event.deleted && event.deletedFields
-          ? ` · deleted ${safelyFormat(event.deletedFields?.dateOfDeletion)}`
-          : null}
-        {event.deleted && event.deletedFields
-          ? ` · ${event.deletedFields?.deleteReason}`
-          : null}
-        {event.deleted && event.deletedFields
-          ? ` · requested by ${event.deletedFields?.deleteRequestedBy}`
-          : null}
+        {event.deleted &&
+          event.deletedFields &&
+          ` · deleted ${safelyFormat(event.deletedFields?.dateOfDeletion)}
+          · ${event.deletedFields?.deleteReason}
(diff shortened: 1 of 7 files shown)


components/NewPersonView/Event.tsx
```diff
@@ -48,6 +48,15 @@ const Event = ({ event }: Props): React.ReactElement => {
         {safelyFormat(displayDate)}
         {event.caseFormUrl?.includes('google') && ` · Google document`}
         {event.officerEmail && ` · ${event.officerEmail}`}
+        {event.deleted && event.deletedFields
```
JohnFarrellDev, line 51: Why do we have the same conditional check 3 times?

Could we just do 

```

{event.deleted && event.deletedFields && (
    ` · deleted ${safelyFormat(event.deletedFields?.dateOfDeletion)}
      · ${event.deletedFields?.deleteReason}
      · requested by ${event.deletedFields?.deleteRequestedBy}
    `
)}

```

factories/cases.ts
```diff
@@ -57,6 +57,12 @@ export const mockedDeletedCaseNote = caseFactory.build({
   },
   formName: 'adult-case-note',
   deleted: true,
+  deletedFields: {
+    deletedBy: 'jack.musajo@hackney.gov.uk',
+    dateOfDeletion: '2020-12-31',
```
JohnFarrellDev, line 62: Are we definitely returning in this format and not an ISO datetime?

Probably fine if we are but just checking

rachel-hutchinson: I checked and we are expecting ISO format so I've updated this

JohnFarrellDev (approved):
LGTM :+1:

rachel-hutchinson:
Holding off merging this into main until the backend is ready

JohnFarrellDev:
> Holding off merging this into main until the backend is ready

could be worth making this a draft PR for now

rachel-hutchinson:
> > Holding off merging this into main until the backend is ready
> 
> could be worth making this a draft PR for now

I took another look and realised that the backend was working as expected but the frontend was just had inconsistent naming. I also added in e2e tests to confirm it all works as expected

PR #785, opened 2021-12-03
fix(SCT-1346): Extend feature flags

rachel-hutchinson:
**What**  
Some of the features flags had expiry dates that had passed or were close. These dates have been extended to 31/01/2022.

**Why**  
When we pass the expiry date on the feature flags, the main pipeline fails and it is not possible to deploy to production


chubberlisk (approved):
Thanks for picking this up and sorry it blocked your deployment! I'll create a card on our backlog to remove the workflows pilot feature flag.

PR #786, opened 2021-12-03
feat(SCT-1552): Add deleted information to case note history

rachel-hutchinson:
**What**  
Have added in a 'Deleted by' history point to deleted case note records

**Why**  
This keeps the interface consistent and provides additional information when on a specific case note
image

**Anything else?**
Have added an additional e2e test to confirm this works as expected


acodeninja (commented):
The unit tests for this module are a little thin on the ground, any way of getting some of this in there? Not a blocker, but would be nice to have them in addition to the cypress ones.

tony-griffin (approved):
LGTM

PR #789, opened 2021-12-06
fix: Revision Timeline tests

mapleleafjack:
Better test for Revision Timeline deleted record

**What**  
-Edited a test so it's cleaner / neater

**Why**  
The `queryByText` function with some parameters were not testing the correct items, but tests were passing anyways.

**Anything else?**


tony-griffin:
Revisions look good

tony-griffin (approved):
Tests look good

PR #790, opened 2021-12-08
chore(SCT-928): Update pipeline to deploy to mosaic-production account only

LBHTKarki:
**What**  
Remove deployment workflow for old production account

**Why**  
Now that the front end Lambda migration is complete there's no longer need to deploy to the old production account. This update simplifies the pipeline and enables us to proceed with clearing up the old recources.

**Anything else?**

Updated `generate-github-release` step with correct dependency

Serverless config changes to remove the old production resources will be handled in separate PRs.

`mosaic-production` is still used in the file for the new account name to avoid confusion. That can be changed later if required.


chubberlisk (approved):
👍

mapleleafjack (approved):

PR #797, opened 2021-12-13
fix(SCT-1452): dependency audit

acodeninja:
**What**  
Clean up dependencies to update to non-vulnerable versions of current deps.

**Why**  
Dependabot couldn't do this on its own.

**Anything else?**

- Removed storybook, took a poll in the dev channel and no-one had use this in a while.
- Set cypress to version 9
- Set non-breaking upgrades on nextjs.
- Fixed type errors for AxiosError in endpoints.


mapleleafjack (approved):
Looks good to me! Good one!

PR #799, opened 2021-12-13
feat(SCT-1279): sentry setup & Content Security Policy for sentry

tony-griffin:
What
Add Sentry.io to the Content Security Policy to allow our app to access content from that domain

Why
Helps us control the domains that can deliver content to the application so as to mitigate certain types of attacks, including Cross-Site Scripting (XSS) and data injection attacks

Anything else?

Files to review:

                      next.config.js
                      pages/_error.js
                      pages/api/throwError.ts
                      sentry.client.config.js
                      sentry.server.config.js
                      utils/contentSecurity.ts
                      utils/contentSecurity.spec.ts
                      pages/throw-error.tsx
                      pages/api/throw-error.ts


acodeninja (approved):
Good first round, I would recommend a follow up to create a wrapper for all api methods that wraps the handlers in withSentry, the current setup will only cover the endpoint  `/api/case-notes/[id]` AFAIK

PR #800, opened 2021-12-14
feat(SCT-1572): update the CircleCI/Node image being used in config.yml

tony-griffin:
**What**  
The current CircleCI/Node image is about to be deprecated.
We use this to run many of our executors, a n ode environment in which to run our code on CircleCI

We should update the image to the new **cimg/node** format and relevant release number.


acodeninja (approved):
LGTM

PR #803, opened 2021-12-15
feat(SCT-1534): permissions checks for Case Statuses

mapleleafjack:
[Link to ticket](https://hackney.atlassian.net/browse/SCT-1534)

**What**  
Added permissions check for adding/editing/ending case status as per ticket:

-Only people in the saml-socialcare-safeguardingandreviewing should be able to add and update Child Protection Statuses. 
-Only people in the saml-socialcare-placementmanagementunit should be able to update Looked After Children Statuses. 
-Everyone with access to the CFS side of the system should be able to add and update Child in Need Statuses. 
-Everyone with access to the CFS side of the system should be able to see all statuses. 

**Why**  
To let users with different permissions work on different part of the Case Status journey


tony-griffin (approved):
Looks good Jack!

PR #804, opened 2021-12-15
fix: Serverless ssm parameters

mapleleafjack:
**What**  
-Removed "~true" from variable ssm link
-added "variablesResolutionMode" parameter set to `20210326`

**Why**  
Serverless recently returned this error:
```
Serverless: Checking Stack update progress...
....... 
 Serverless Error ----------------------------------------
 
  The security token included in the request is expired
 
  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com
 
  Your Environment Information ---------------------------
     Operating System:          linux
     Node Version:              14.17.5
     Framework Version:         2.69.1
     Plugin Version:            5.5.1
     SDK Version:               4.3.0
     Components Version:        3.18.1
 
Serverless: Deprecation warning: Syntax for referencing SSM parameters was upgraded with automatic type detection and there's no need to add "~true" or "~split" postfixes to variable references.
            Drop those postfixes and set "variablesResolutionMode: 20210326" in your service config to adapt to a new behavior.
            Starting with next major release, this will be communicated with a thrown error.
            
            More Info: https://www.serverless.com/framework/docs/deprecations/#NEW_VARIABLES_RESOLVER

Exited with code exit status 1
```

**Anything else?**

[Documentation for variablesResolutionMode parameter](https://www.serverless.com/framework/docs/providers/aws/guide/variables)

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -1,8 +1,9 @@
 service: lbh-social-care
 
+variablesResolutionMode: 20210326
+
 provider:
   name: aws
-  variablesResolutionMode: 20210326
   runtime: nodejs14.x
   versionFunctions: false
   region: eu-west-2
--- a/serverless.yml
+++ b/serverless.yml
@@ -37,7 +37,7 @@ functions:
       subnetIds: ${self:custom.subnets.${self:provider.stage}}
     environment:
       NEXT_PUBLIC_ENV: ${opt:stage}
-      AWS_KEY: ${ssm:/lbh-social-care/${self:provider.stage}/aws-key~true}
+      AWS_KEY: ${ssm:/lbh-social-care/${self:provider.stage}/aws-key}
       ENDPOINT_API: ${ssm:/lbh-social-care/${self:provider.stage}/endpoint-case-viewer}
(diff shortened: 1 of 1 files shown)


serverless.yml
```diff
@@ -2,6 +2,7 @@ service: lbh-social-care
 
 provider:
   name: aws
+  variablesResolutionMode: 20210326
```
acodeninja, line 5: I'm pretty sure this should be at the root...

https://www.serverless.com/framework/docs/providers/aws/guide/variables#reference-variables-using-the-ssm-parameter-store

https://github.com/LBHackney-IT/lbh-core-pathway-pilot/blob/main/serverless.yml#L3

acodeninja (changes requested):
Looks good, though I think the `variablesResolutionMode` should be set in the root of the file rather than in the provider, details are in a separate comment.

serverless.yml
mapleleafjack: My bad, fixed!

acodeninja (approved):
Great, thank you, look like it should do the trick.

PR #807, opened 2021-12-17
feat(SCT-1569): address dropdown filtered by building number & postcode

tony-griffin:
**What**  

Problem:
Some postcodes can return up to 6000 results for a postcode (though that is not the normal size)
We display these results in a drop down which is limited to returning only 50 results due to the address api limits
There’s no guarantee that the desired address will be in the result set returned.

Solution:
Update the postcode search to return all the matching addresses and implement an optional ability to add a building number to the search query to help filter out non matching results
This would be totally optional and a user would not need to know the building number in order to do an address search.

**Why**  

Allow users to find the address more easily.


utils/api/postcodeAPI.ts
```diff
@@ -24,7 +24,14 @@ export const normalizeAddress = (address: AddressAPI): Address => {
...
-export const lookupPostcode = async (postcode: string): Promise<Address[]> => {
-  const { data } = await axios.get(`/api/postcode/${postcode}`);
-  return data.address.map(normalizeAddress);
+export const lookupPostcode = async (
+  postcode: string,
+  page_number = 1,
+  building_number?: string
+): Promise<AddressWrapper> => {
+  const { data } = await axios.get(
+    `/api/postcode/${postcode}?page=${page_number}&buildingNumber=${building_number}`
```
chubberlisk, line 33: Could we add a test for this change? See https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/a582468826dcc24ee9880d31fc50cfcf5b9e688e/utils/api/relationships.spec.ts#L149-L160 for an example.

utils/allNumericInputs.ts
```diff
@@ -0,0 +1,8 @@
+export const allNumeric = (inputTxt: string) => {
```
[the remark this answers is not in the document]
chubberlisk: Could we add some unit tests for this function please?

cypress/integration/search_address.ts
```diff
@@ -0,0 +1,82 @@
...
+    cy.get(`input[id=postcode]`).click().clear().type('SE17 1JJ');
+    cy.get(`button[id=lookup-button]`).click();
+    cy.get(`select[id=address]`).should('be.visible');
+    cy.get(`select[id=address]`)
+      .select('194 WALWORTH ROAD')
+      .should('have.text', '194 WALWORTH ROAD');
+  });
+});
+
+describe('Validates user input', () => {
```
chubberlisk, line 45: These are good tests but they feel more like unit tests for the `AddressLookup` component. See https://github.com/LBHackney-IT/lbh-core-pathway-pilot/blob/ea9c13c6196bc0a2bcdfa77d2060e27d2deadbdd/components/ManagerApprovalDialog.test.tsx#L121-L137 as an example. Let me know if you want to 🍐 on dis.

components/Form/AddressLookup/AddressLookup.tsx
```diff
@@ -130,16 +185,41 @@ const AddressLookup = ({
           {hint}
         </span>
       )}
-      <div className="govuk-grid-row">
-        <div className="govuk-grid-column-one-third">
+
+      <div>
+        <div className={styles.inputContainer}>
```
chubberlisk, line 190: By adding this custom styling, we lose the normal styling for the focus state of the inputs, see [Understanding focus state styles](https://design-system.service.gov.uk/get-started/focus-states/).

Normal focus state styling | Custom styling
-- | --
image  | Screenshot 2021-12-22 at 15 10 26

I'd recommend refactoring to use the `TextInput` component so custom styling isn't required:

```jsx
<TextInput
  label="Building number"
  name="buildingNumber"
  width={3}
  id="buildingNumber"
  onChange={(e) => setBuildingNumber(e.target.value)}
/>
```

components/Form/AddressLookup/AddressLookup.spec.tsx
```diff
@@ -30,6 +34,16 @@ describe('AddressLookup', () => {
...
+      expect(validate.buildingNumber()).toBe(
+        'Building number must use valid characters (0-9)'
+      );
+      expect(validate.buildingNumber({ buildingNumber: '123' })).toBe(true);
+      expect(validate.buildingNumber({ buildingNumber: '123a' })).toBe(
+        'Building number must use valid characters (0-9)'
+      );
+      expect(validate.buildingNumber({ buildingNumber: '' })).toBe(
+        'Building number must use valid characters (0-9)'
+      );
```
chubberlisk, line 46: I think these tests warrant a refactoring as the test descriptions don't document the function very well and each test tests different behaviours but for this PR maybe we could separate out the building number related behaviours?

chubberlisk (commented):
Had a looksies as I know you've got a number of your teammates off! Tested locally and seems to work as expected, I've mainly added suggestions/comments around tests.

utils/allNumericInputs.ts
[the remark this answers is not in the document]
tony-griffin: Unit tests added 👍

utils/api/postcodeAPI.ts
tony-griffin: Test added which required some changes in the implementation code to cover response being undefined.

components/Form/AddressLookup/AddressLookup.tsx
tony-griffin: Great, didn't know this component existed. I am having an issue with a ref prop not being part of the type though. In the code for TextInput it looks like it spreads any remaining props so thought it would be ok to keep the pre existing `ref={inputRef}` as it's being used to set focus

components/Form/AddressLookup/AddressLookup.spec.tsx
tony-griffin: Agreed, have split them out into their own describe blocks and made them more descriptive

cypress/integration/search_address.ts
rachel-hutchinson: Thanks Ting, we've moved the validation tests into the AddressLookup.spec unit test file. 

We also tried moving the address search tests to the unit test file, however, this didn't work as we wanted to test the real response from the postcodeAPI and we were unable to make this axios.get() call in the unit tests without mocking the response, I hope this all makes sense

components/Form/AddressLookup/AddressLookup.spec.tsx
```diff
@@ -1,35 +1,294 @@
-import { defaultValidation } from './AddressLookup';
+import { defaultValidation, AddressBox } from './AddressLookup';
```
chubberlisk, line 1: The changes you've made in this test file is such an improvement, you made broken up a big test and made it clearer to understand and you've added missing coverage. Thank you so much! 💯

components/Form/AddressLookup/AddressLookup.spec.tsx
```diff
@@ -1,35 +1,294 @@
...
+          arg0?: Partial<AddressBox['value']>
+        ) => true | 'You must enter a valid postcode';
+      };
+
+      beforeEach(() => {
+        validate = defaultValidation();
+        return validate;
+      });
+
+      it('when validation is not required should return true when address has no value', () => {
```
chubberlisk, line 27: Might be a personal preference thing but I'd prefer to organise these tests like so describe blocks are:

- "for address"
  - "when validation is not required"
  - "when validation is required"

and then remove the need to have "when validation is not required/ required" in each of the test description as it makes it long to read in the terminal.

components/Form/AddressLookup/AddressLookup.tsx
```diff
@@ -95,27 +99,64 @@ const AddressLookup = ({
...
+        );
+
+        address.length === 0
+          ? (errorMessage = 'There was a problem with the postcode.')
+          : matchingAddresses?.push(...address);
+        page_number === page_count ? (isLastPage = true) : page_number++;
+      } catch {
+        errorMessage = 'There was a problem with the postcode.';
+      }
+    } while (!errorMessage && !isLastPage);
```
chubberlisk, line 151: Not 100% sure this is the most performant way of doing what we need, possibly making use of [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) might be what we need but this is good enough for now! Make it work, make it right then make it fast. Amirite?!

utils/api/postcodeAPI.spec.ts
```diff
@@ -1,10 +1,15 @@
-import { normalizeAddress } from './postcodeAPI';
+import { formatAddress, lookupPostcode } from 'utils/api/postcodeAPI';
+import axios from 'axios';
+import { addressAPIWrapperFactory } from 'factories/postcode';
+
+jest.mock('axios');
+const mockedAxios = axios as jest.Mocked<typeof axios>;
 
 describe('postcodeAPI', () => {
   describe('normalizeAddresses', () => {
```
chubberlisk, line 9:
```suggestion
  describe('formatAddress', () => {
```

Since we've renamed the function.

utils/api/postcodeAPI.spec.ts
```diff
@@ -1,10 +1,15 @@
...
-    it('should work properly', () => {
+    it('should format address data', () => {
```
chubberlisk, line 10:
```suggestion
    it('combines the address lines into single line', () => {
```

Definitely an improvement on "should work properly" but maybe this suggestion or something similar is more explicit to what we mean by formatting?

chubberlisk (approved):
Thank you for **_addressing_** my feedback! (Get it?!) I appreciate this might have taken a while but you've all made some great improvements. 💯

components/Form/AddressLookup/AddressLookup.spec.tsx
rachel-hutchinson: All thanks to @tony-griffin for his hard work on this one :)

components/Form/AddressLookup/AddressLookup.spec.tsx
rachel-hutchinson: Thanks, that makes sense, I've rearranged it as you've suggested

components/Form/AddressLookup/AddressLookup.tsx
rachel-hutchinson: oh, interesting, I hadn't thought of a Promise.all - I'll leave it as is for now but might revisit it with the Promise.all in mind

mapleleafjack (approved):

PR #809, opened 2022-01-04
Adding functionality to 'cancel' button on MASH forms

rs2416:
**What**  
Added the 'back/cancel' functionality to the cancel button on the MASH forms. Along with additions to the tests for the forms.

**Why**  
This allows practitioners to go back/cancel whist on a form

### DIFF
--- a/components/MashForms/FinalDecisionForm.tsx
+++ b/components/MashForms/FinalDecisionForm.tsx
@@ -177,6 +177,7 @@ const FinalDecisionForm = ({
           <a
             href="#"
             className={`lbh-link lbh-link--no-visited-state govuk-!-margin-left-3`}
+            onClick={() => window.history.back()}
           >
             Cancel
           </a>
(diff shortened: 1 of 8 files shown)


martapederiva (commented):
Looks good! Only thing are some tests failing in the pipeline (aside from the PR title), not sure it's because of your code though.

components/MashForms/FinalDecisionForm.tsx
```diff
@@ -177,6 +177,7 @@ const FinalDecisionForm = ({
           <a
             href="#"
             className={`lbh-link lbh-link--no-visited-state govuk-!-margin-left-3`}
+            onClick={() => window.history.back()}
```
chrishutchinson, line 180: This is a bit of a nitpick, but I _think_ we should use Next.js' `router.back()` method here, instead of `window.history.back()`. It's not a massive deal but might be nicer. Docs for it are here: https://nextjs.org/docs/api-reference/next/router#routerback

rs2416: Will take a look! Thanks

rs2416: Have made the change on the form pages - just need to figure out how to alter the tests

chrishutchinson (approved):

PR #810, opened 2022-01-07
fix: resolve issue with date in review warning note tests

chrishutchinson:
We had a bug with a test that used a date of Jan 2nd 2022, which worked until the world ticked past Jan 2nd 2022, then it started failing.

This switches that test to use a dynamically generated date instead!


martapederiva (approved):
✨ go dynamic dates!

PR #812, opened 2022-01-10
feat(SCT-1279): add lambda layer & sentry setup branch

tony-griffin:
**What**  

This PR covers the sentry set up that we previously attempted but had to roll back due to the unexpected hurdle of the deployment package exceeding the lamdba function size. 

Though the Sentry code was previously reviewed and approved, we had to wait to be able to produce a smaller deployment package in order to move forward.

We decided to roll back and pursue adding the node modules folder as a lambda layer to get around the function limits.

This now appears to produce a layer and lambda function deployment package of the correct sizes. I have combined the sentry code and lambda layer code into the PR as we needed both to be working to move ahead with Sentry.
The PR that covers that layer work is [here](https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/811)

The layer has been introduced into the servreless.yml file that takes a copy of the current node modules folder. The lambda function will reference this layer when zipping up the deployment package.

The two variables `NEXT_PUBLIC_SENTRY_TARGET` and `NEXT_PUBLIC_SENTRY_DSN` must be set for this functionality to work.

`NEXT_PUBLIC_SENTRY_TARGET` has value `local` in the development environment


.circleci/config.yml
```diff
@@ -74,6 +74,14 @@ jobs:
...
           paths:
             - project/build
+            
+  create-layer-directory:
+    executor: node-executor
+    steps:
+      - *attach_workspace
+      - run:
+          name: create layer directory
+          command: yarn copy-node-modules
```
acodeninja, line 84: Just a note on the naming of this file, I'm not sure it really describes what the command does, maybe something like create or prepare -lambda-layer? Just thinking that it doesn't actually copy anything.

.circleci/config.yml
```diff
@@ -186,12 +193,34 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Create release and notify Sentry of deploy
+          command: |
+            curl -sL https://sentry.io/get-cli/ | bash
+            export SENTRY_RELEASE=$(sentry-cli releases propose-version)
+            sentry-cli releases new -p $SENTRY_PROJECT $SENTRY_RELEASE
+            sentry-cli releases set-commits $SENTRY_RELEASE --auto
+            sentry-cli releases files $SENTRY_RELEASE upload-sourcemaps path-to-sourcemaps-if-applicable
```
acodeninja, line 212: A question here, mostly because I'm not familiar with circle ci, after the checkout step is the nextjs build diirectoy available for the source maps to be uploaded? 

Also shouldn't `path-to-sourcemaps-if-applicable` be a directoy path?

pages/_error.js
```diff
@@ -0,0 +1,69 @@
+import NextErrorComponent from 'next/error';
+import PropTypes from 'prop-types';
+
+import * as Sentry from '@sentry/nextjs';
+
+const MyError = ({ statusCode, hasGetInitialPropsRun, err }) => {
```
acodeninja, line 6: `MyError` maybe we can change this to ErrorPage or something similar?

tsconfig.json
```diff
@@ -25,7 +25,7 @@
     "next-env.d.ts",
     "**/*.ts",
     "**/*.tsx"
-  ],
+, "pages/_error.js"  ],
```
acodeninja, line 28: Syntax

acodeninja (changes requested):
I've made a few small points on naming and a question on circle ci.

.circleci/config.yml
tony-griffin: Fair point, will rename this.

tsconfig.json
tony-griffin: Linted

pages/_error.js
tony-griffin: Changed to ErrorPage

.circleci/config.yml
```diff
@@ -259,4 +269,4 @@ workflows:
             - build-deploy-mosaic-production
           filters:
             branches:
-              only: main
+              only: main
```
acodeninja, line 272: Doesn't matter too much but missing new line at the end of this file.

tsconfig.json
```diff
@@ -19,15 +15,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "baseUrl": "."
+    "baseUrl": ".",
```
acodeninja, line 18: Invalid JSON (typescript still compiles with this because of the way it loads it).

serverless.yml
```diff
@@ -55,10 +66,13 @@ functions:
       POSTCODE_LOOKUP_APIKEY: ${ssm:/lbh-social-care/${self:provider.stage}/postcode-lookup-apikey}
       NEXT_PUBLIC_FEEDBACK_LINK: ${ssm:/lbh-social-care/${self:provider.stage}/next-public-feedback-link}
       REDIRECT_URL: ${ssm:/lbh-social-care/${self:provider.stage}/redirect_url}
+      SENTRY_RELEASE: ${env:SENTRY_RELEASE}
       NEXT_PUBLIC_GOOGLE_ANALYTICS_ID: ${ssm:/lbh-social-care/${self:provider.stage}/next_public_google_analytics_id}
       NOTIFY_API_KEY: ${ssm:/lbh-social-care/${self:provider.stage}/notify-api-key}
       NOTIFY_APPROVER_TEMPLATE_ID: ${ssm:/lbh-social-care/${self:provider.stage}/notify-approver-template-id}
       NOTIFY_RETURN_FOR_EDITS_TEMPLATE_ID: ${ssm:/lbh-social-care/${self:provider.stage}/notify-return-for-edits-template-id}
+      NEXT_PUBLIC_SENTRY_DSN: ${ssm:/lbh-social-care/${self:provider.stage}/next-public-sentry-dsn}
```
acodeninja, line 74: I'm not sure about polluting the environment with NEXT_PUBLIC_ env vars (as these should only be needed at build time) personal preference that I'm happy to be pushed back on but feels like using them at backend runtime muddies the waters if you need to debug something.

acodeninja (commented):
A few comments, and maybe a discussion to be had around the `NEXT_PUBLIC_` env vars.

.circleci/config.yml
tony-griffin: Good spot, have added this in.

tsconfig.json
tony-griffin: Removed the comma 👍

serverless.yml
tony-griffin: Decided we will rework what  NEXT_PUBLIC env vars there are in a new tech debt ticket.

acodeninja (approved):
Looks good to me now, lets get it on staging and test it with the error pages.

PR #819, opened 2022-01-20
feat(SCT-1279): remove sentry_release var from serverless.yml

tony-griffin:
**What**  
remove sentry_release var from serverless.yml

**Why**  
To see if we can get around the following error:

```
Serverless Error ----------------------------------------
 
  Cannot resolve serverless.yml: Variables resolution errored with:
    - Cannot resolve variable at "functions.lbh-social-care.environment.SENTRY_RELEASE": Value not found at "env" source
```


rachel-hutchinson (approved):
LGTM

PR #820, opened 2022-01-20
feat(SCT-1279): update lambda layer files included

tony-griffin:
**What**  
Update the files that the lambda layer can include when building serverless configuration


rachel-hutchinson (approved):
Looks good

PR #822, opened 2022-01-20
feat(SCT-1279): remove explicit exclude from lambda function

tony-griffin:
**What**  
Test include/exclude statements

**Why** 
To correct the error where these statements can't find the relevant files.


rachel-hutchinson (approved):
Looks good

PR #829, opened 2022-01-24
chore: Revert sentry

mapleleafjack:
**What**  
Reverted changes (removed Sentry) so the pipeline works while we work on a solution


rachel-hutchinson (approved):
Looks good

PR #830, opened 2022-01-25
chores: setup sentry and lambda layer

mapleleafjack:
**What**  
This PR adds a Layer to the Lambda that holds the main application, which will hold the node_modules.
It also add Sentry to the platform


tony-griffin (approved):
LGTM

PR #831, opened 2022-01-25
chore: Config change so we don't need a Lambda layer

mapleleafjack:
**What**  
Removed the unnecessary / nonworking bits around the Lambda layer.

**Why**  

**Anything else?**


rachel-hutchinson (approved):
Looks good

PR #832, opened 2022-01-25
chore: Serverless config switch from include to patterns

mapleleafjack:
**What**  
Changed the way we include/exclude files in the serverless configuration.

**Why**  
We cannot use include/exclude together with patterns, so we opted for using only patterns


acodeninja (approved):
LGTM

PR #835, opened 2022-01-26
chore(SCT-1658): Upgrade next and cross fetch

rachel-hutchinson:
**What**  
Upgraded next and cross-fetch dependencies

**Why**  
This was done to resolve a dependabot alert for the node-fetch package


package.json
```diff
@@ -36,7 +36,7 @@
     "lbh-frontend": "^3.5.3",
     "lodash.debounce": "^4.0.8",
     "lodash.get": "^4.4.2",
-    "next": "^11.1.1",
+    "next": "^12.0.8",
```
acodeninja, line 39: Just a little note, have you gone over the upgrade guide for 11 to 12 to make sure everything will work? https://nextjs.org/docs/upgrading#upgrading-from-11-to-12

rachel-hutchinson: Thanks Lawrence, I hadn't spotted that documentation
I've dropped the next version to 11.1.3 and created ticket SCT-1666 to handle the upgrade to version 12

PR #836, opened 2022-01-26
fix(SCT-1659): update sentry env vars to use correct variable types

tony-griffin:
## **What**  

Remove **NEXT_PUBLIC** type env vars from sentry.server.config and next.config.js files & replace them with appropriate type variables.

In this case, remove  NEXT_PUBLIC_SENTRY_DSN  and  NEXT_PUBLIC_SENTRY_TARGET 
to be replaced with:
SENTRY_DSN  and  SENTRY_TARGET.

These new variables would have the same value as the previous NEXT_PUBLIC variables but we would be able to change the value of the new variables at runtime.
Currently, this would mean adding these new variables to our local **.env** file also.

## **Why**  

NEXT_PUBLIC variable types are set at build time and cannot be changed after.
We were using these variable types in files that may need to change variable values at runtime and this was inappropriate.

Currently we are using NEXT_PUBLIC type env vars in:
- sentry.server.config
- sentry.client.config
- next.config.js
- In our .env file

We should only be using **NEXT_PUBLIC** type env vars for processes that run at build time & for variables whose value we do not expect to change at runtime.
**NEXT_PUBLIC** type variables should not be used for variables whose value may change at runtime

Also, we wish to set the right environment value for assigning issues against Sentry reporting.
This means we need to align the environments referenced in the application to those used by Hackney, which are:
**”dev”,  “stg“,  “prod“**

- When these are aligned then we will be able to assign application issues against these environments in 
   Sentry.
- Currently, all Sentry issues are being logged against the “dev“ environment which is incorrect.
- The SENTRY_TARGET variable used in the sentry.server.config file is what is used to assign issues against  
   an environment in Sentry.
   
## **AWS**

We have added references to these new variables to Staging in the parameter store:
```
/lbh-social-care/staging/sentry-dsn
/lbh-social-care/staging/sentry-target
```
We are unsure if we still need the existing AWS parameter store variables that have the NEXT_PUBLIC prefix:
```
/lbh-social-care/staging/next-public-sentry-dsn
/lbh-social-care/staging/next-public-sentry-target
```
We will need to get someone with permissions who can either edit or delete these variables as needed.
   

### - Add :
`/lbh-social-care/mosaic-production/sentry-dsn `
& 
`/lbh-social-care/mosaic-production/sentry-target`
to production on AWS parameter store with the relevant values. 
SENTRY_TARGET should be “prod“

### - Edit:
We must also edit the values for the equivalent 
` /lbh-social-care/staging/next-public-sentry-target`, 
env var in the Staging account, which should be “stg“

We will need to get someone with permission to edit the value for
` /lbh-social-care/mosaic-production/next-public-sentry-target`, 
which should be “prod“


acodeninja (approved):
LGTM

PR #842, opened 2022-02-04
fix: Expired session token error

mapleleafjack:
**What**  
This PR fixes the Internal Server Error when the token expires

**Why**  
Screenshot 2022-02-04 at 15 35 48


utils/auth.ts
```diff
@@ -10,8 +10,6 @@ export const AUTH_WHITELIST = ['/login', '/access-denied'];
 
 const GSSO_TOKEN_NAME = process.env.GSSO_TOKEN_NAME;
 
-export const SESSION_EXPIRY = 14400;
```
acodeninja, line 13: I'm not sure this is right @mapleleafjack we should still be checking the expiry.

mapleleafjack: Re-added!

martapederiva (approved):
💡

PR #846, opened 2022-02-07
fix: (SCT-1668) Cannot read properties of undefined (reading 'slice') - TypeError call(lib/validators)

tony-griffin:
**What**  
Sentry issue - [SCT-1668](https://hackney.atlassian.net/browse/SCT-1668)

**Why**  
When a date was not set we had no value on which to call the slice() method.


acodeninja (approved):
LGTM

PR #859, opened 2022-02-24
fix: unit test errors

acodeninja:
**What**  

Changes to reduce the number of errors on the console in unit tests. This also introduces fixes to a number of tests with issues.

**Why**  
Because seeing warnings and errors in tests is a code smell. This also meant I found missing and incomplete tests.

**Anything else?**

### DIFF
--- a/data/flexibleForms/caseStatus/caseStatus.ts
+++ b/data/flexibleForms/caseStatus/caseStatus.ts
@@ -57,12 +57,22 @@ const form: Form = {
           question: 'Start Date',
           type: 'date',
           required: true,
-          conditions: [
-            {
-              id: 'type',
-              value: 'CIN',
-            },
-          ],
+          conditions: {
+            OR: [
+              {
+                id: 'type',
+                value: 'CIN',
+              },
+              {
+                id: 'type',
+                value: 'CP',
+              },
+              {
+                id: 'type',
+                value: 'LAC',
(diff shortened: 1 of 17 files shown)


data/flexibleForms/caseStatus/caseStatus.ts
```diff
@@ -57,12 +57,22 @@ const form: Form = {
...
           type: 'date',
           required: true,
-          conditions: [
-            {
-              id: 'type',
-              value: 'CIN',
-            },
-          ],
+          conditions: {
+            OR: [
```
mapleleafjack, line 61: Didn't know you could use OR, good to know!

mapleleafjack (approved):

data/flexibleForms/caseStatus/caseStatus.ts
acodeninja: You couldn't, I added this here: https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/859/files#diff-9893684b9b99f4716e2adf2b28835c8e6f78037ecb1346a1e7a091712dc3bb75

mapleleafjack: Yup, noticed that :) Will definitely use this approach next time we have a similar problem to solve :)

PR #863, opened 2022-02-28
fix: fix font in new NextJS version

mapleleafjack:
**What**  
The new NextJS version breaks the font import.
We added them back manually


tony-griffin (approved):
LGTM

PR #864, opened 2022-03-01
fix: Added nonce to stylesheet link

mapleleafjack:
**What**  
The login page is not loading the Google CDN for the font stylesheet.
This is because there's a CSP preventing the script for loading.
This PR adds the nonce to the script so it can be executed correctly.

**Why**  
The login page is not loading the Google CDN for the font stylesheet.


rachel-hutchinson (approved):
Looks good to me

PR #868, opened 2022-03-02
chore: new resident view fixes

jhackett1:
assorted small fixes to the resident view

- [ ] fixes text overflow bug (pictured)
- [ ] fixes issue where resident does not have a creator recorded, and therefore always fails when trying to update it

### DIFF
--- a/pages/api/residents/[id]/index.ts
+++ b/pages/api/residents/[id]/index.ts
@@ -45,7 +45,11 @@ const endpoint: NextApiHandler = async (
 
     case 'PATCH':
       try {
-        const data = await updateResident({ id, ...req.body });
+        const data = await updateResident({
+          id,
+          ...req.body,
+          createdBy: req.body.createdBy || user.email,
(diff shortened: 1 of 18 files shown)


pages/api/residents/[id]/index.ts
```diff
@@ -45,7 +45,11 @@ const endpoint: NextApiHandler = async (
 
     case 'PATCH':
       try {
-        const data = await updateResident({ id, ...req.body });
+        const data = await updateResident({
+          id,
+          ...req.body,
+          createdBy: req.body.createdBy || user.email,
```
Comment (moarpheus) [inline pages/api/residents/[id]/index.ts:51]: It feels like this deserves a test.

mapleleafjack (approved):

moarpheus (approved):

pages/api/residents/[id]/index.ts
Comment (jhackett1) [reply pages/api/residents/[id]/index.ts:51]: i agree

PR #872, opened 2022-03-04
chore: bring back resident view v3 preview banner

jhackett1:
bring back preview banner linking to the new resident view


acodeninja (approved):
Should do the trick. Lets get it to staging and go from there.

PR #873, opened 2022-03-04
feat(SCT-1715): team pages and allocations

jhackett1:
we are overhauling the way our app treats:

- teams and the users who belong to them
- allocations, particularly allocations to a team but not a particular user

this adds:

- [x] team pages at a `/teams/:id` path inside the app, which will show that team's waiting list (i.e. people allocated to the team but not to an individual user within it)
- [x] subpages under each team for `/members` and `/allocations`
- [x] the main user page is now the new worker allocation list
- [x] changes the way the worker allocation request is done so to add sorting parameters
- [x] and more!

Screenshot 2022-04-11 at 11 49 37

Screenshot 2022-04-11 at 11 49 45

Screenshot 2022-04-11 at 11 52 44


martapederiva (approved):
👍 🔥

PR #874, opened 2022-03-04
chore: better empty and initial states for case notes, workflows and maps

jhackett1:
- [x] better null states for case notes/records and workflows
- [x] better null state for map+street view module
- [x] improve the spacing between status tags when they wrap onto a new line
- [x] show full, rather than abbreviated, primary support reason in the read-only (rather than edit) state of the personal details dialog

Screenshot 2022-03-04 at 14 36 00

Screenshot 2022-03-04 at 14 44 10

Screenshot 2022-03-04 at 15 03 04

### DIFF
--- a/components/ResidentPage/Mapping.tsx
+++ b/components/ResidentPage/Mapping.tsx
@@ -51,6 +51,8 @@ const Mapping = ({ socialCareId }: Props): React.ReactElement | null => {
 
   if (!data) return null;
 
+  if (!data.address?.postcode && !data.address?.address) return null; // handle resident with no address set
+
   const mapUrl = `https://maps.googleapis.com/maps/api/staticmap?size=360x360&return_error_code=true&markers=${prettyAddress(
     data
   )}&zoom=15&key=${process.env.NEXT_PUBLIC_GOOGLE_CLIENT_KEY}`;
--- a/components/ResidentPage/WorkflowOverview.tsx
+++ b/components/ResidentPage/WorkflowOverview.tsx
@@ -20,19 +20,29 @@ const WorkflowOverview = ({
     ?.filter((w) => !w.submittedAt)
     .filter((w) => w.id !== mostRecent?.id); // exclude most recent workflow
 
+  if (workflows?.length === 0)
+    return <p className="lbh-body-s">This resident has no workflows yet.</p>;
+
   return (
     <>
       <div className="govuk-grid-row">
         {inProgress && (
           <div className="govuk-grid-column-one-half">
             <h3 className="lbh-heading-h5">In progress</h3>
-            {inProgress.slice(0, 3)?.map((w) => (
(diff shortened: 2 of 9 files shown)


components/ResidentPage/Mapping.tsx
```diff
@@ -51,6 +51,8 @@ const Mapping = ({ socialCareId }: Props): React.ReactElement | null => {
 
   if (!data) return null;
 
+  if (!data.address?.postcode && !data.address?.address) return null; // handle resident with no address set
```
acodeninja, line 54: Not seeing an accompanying test in `./components/ResidentPage/Mapping.spec.tsx` around this change.

components/ResidentPage/WorkflowOverview.tsx
```diff
@@ -20,19 +20,29 @@ const WorkflowOverview = ({
     ?.filter((w) => !w.submittedAt)
     .filter((w) => w.id !== mostRecent?.id); // exclude most recent workflow
 
+  if (workflows?.length === 0)
```
acodeninja, line 23: Not seeing an accompanying test in `./components/ResidentPage/WorkflowOverview.spec.tsx` around this change.

components/ResidentPage/WorkflowOverview.tsx
```diff
@@ -20,19 +20,29 @@ const WorkflowOverview = ({
...
+                  <WorkflowChunk workflow={w} key={w.id} />
+                ))}
+                {inProgress.length > 3 && (
+                  <p className="lbh-body-xs govuk-!-margin-top-2">
+                    and {inProgress.length - 3} more
+                  </p>
+                )}
+              </>
+            ) : (
+              <p className="lbh-body-s">Nothing is in progress right now.</p>
```
acodeninja, line 45: Needs some tests for when there are workflows but no in progress to check this shows correctly.

pages/residents/[id]/index.tsx
```diff
@@ -486,22 +490,25 @@ const ResidentPage = ({ resident }: Props): React.ReactElement => {
...
-              {canManage && (
-                <Link href={`/residents/${resident.id}/case-notes`}>
-                  <a className="lbh-link lbh-link--muted lbh-body-xs govuk-!-margin-top-2">
-                    See all {totalCount} case notes & records
-                  </a>
-                </Link>
-              )}
-            </>
-          )}
+          {cases &&
```
Comment (acodeninja) [inline pages/residents/[id]/index.tsx:493]: As there is quite a bit of code in the frontend pages here, it would be good to see some cypress tests covering this, or some unit/acceptance tests if you prefer.

acodeninja (changes requested):
A few tests to add, otherwise looks good 👍

AlexHerbertMadeTech (approved):
Seems good to me

PR #875, opened 2022-03-09
feat(SCT-1737, SCT-1738): Refactor frontend to use new search endpoint 

rachel-hutchinson:
**What**  
* Combined the first name and last name search boxes, so there is now just one person name search box when looking for a person or adding a relationship
* Updated tests to input information into this new search box
* Added new search endpoint that will be used when searching for a person or adding a relationship

image

**Why**  
This change was made to allow for fuzzy searching on a persons name. This means that users could misspell a person's name and still be able to find the person they are looking for.

**Anything else?**
This depends on [this PR](https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/595) which adds the new endpoint in the backend

### DIFF
--- a/components/Search/Search.spec.tsx
+++ b/components/Search/Search.spec.tsx
@@ -1,12 +1,18 @@
-import { act, fireEvent, render } from '@testing-library/react';
+import {
+  act,
+  fireEvent,
+  render,
+  screen,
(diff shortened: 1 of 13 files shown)


acodeninja (approved):
LGTM

PR #876, opened 2022-03-11
fix(SCT-1737): Change from using mosaic_id to person_id in the front end search

rachel-hutchinson:
**What**  
Update the frontend to have the name person_id for the field used to search for a specific id

**Why**  
The frontend was passing mosaic_id to the new search endpoint, however the end point was expecting person_id

**Anything else?**
Altered the query that is passed when switching from person search to record search. Now only the person_id will be passed between the two searches


mapleleafjack (approved):

acodeninja (approved):
LGTM

PR #879, opened 2022-03-15
fix(SCT-1752): UI refactor to split first and last name on search

acodeninja:
**What**  
Split the first name and last name in the search by name UI. The fields are then combined and sent as a single string in the query params.

**Why**  
Users are averse to the change that combines both fields, this change merely makes the UI more comfortable for them.

**Anything else?**


mapleleafjack (approved):
🚀

PR #880, opened 2022-03-16
feat(SCT-1756): stop combining first and last name on search

acodeninja:
**What**  
Send first and last name separately to search endpoint, API updated under https://github.com/LBHackney-IT/social-care-case-viewer-api/pull/601

**Why**  
Details discussed on ticket, but primarily to allow for more accurate results when giving first and last names.

**Anything else?**


mapleleafjack (approved):
I guess the search doesn't return anything until the API is deployed.

Anyways looks good to me :)

PR #884, opened 2022-03-18
feat(SCT-1769): Add team allocation

mapleleafjack:
**What**  
Screenshot 2022-03-18 at 12 16 07

This PR adds a new "Add an allocation" page.
This page is not currently linked in the UI, you can access it by going to the url 
`http://dev.hackney.gov.uk:3000/residents/34/allocations/add?`

**Why**  
This form replace the old one, as it lets the user allocate a person to a team without specifying the worker

**Anything else?**


components/Allocations/SelectWorker/SelectWorker.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+          }),
+        ]}
+        callback={(value: any) => console.log(value)}
+      />
+    );
+    expect(queryAllByText('Roby Robinson')[1]).toBeInTheDocument();
+    expect(queryAllByText('Mary Johnson')[1]).toBeInTheDocument();
+    expect(getByText('90')).toBeInTheDocument();
+    expect(getByText('0')).toBeInTheDocument();
+  });
```
acodeninja, line 61: I'd like to see something like this added to the test suite:

```typescript
 it('should execute the callback on selection', () => {
    const callback = jest.fn();

    const { queryAllByText } = render(
      <SelectWorker
        records={[
          userFactory.build({
            id: 999,
            firstName: 'Roby',
            lastName: 'Robinson',
            allocationCount: 90,
          }),
          userFactory.build({
            firstName: 'Mary',
            lastName: 'Johnson',
            allocationCount: 0,
          }),
        ]}
        callback={callback}
      />
    );

    act(() => {
      fireEvent.click(queryAllByText('Roby Robinson')[0]);
    });

    expect(callback).toHaveBeenCalledWith('999');
  });
```

components/Allocations/SelectWorker/SelectWorker.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <label data-testid="allocationCount" htmlFor={String(id)}>
+            {allocationCount}
+          </label>
+        </td>
+      </tr>
+    </>
+  );
+};
+
+const SelectWorker = ({
```
acodeninja, line 51: Little tip here, not something you must update, but you can make typing functions a lot nicer like this:

```typescript
const SelectWorker: React.FC<{
  records: Worker[];
  callback: any;
}> = ({ records, callback }) => (
  <table className="govuk-table lbh-table" data-testid="residents-table">
```

acodeninja, line 51: Little tip here, not something you must update, but you can make typing functions a lot nicer like this:

```typescript
const SelectWorker: React.FC<{
  records: Worker[];
  callback: (id: string) => void;
}> = ({ records, callback }) => (
  <table className="govuk-table lbh-table" data-testid="residents-table">
```

Also, I'd just specify the function signature rather than passing `any`.

acodeninja (changes requested):
Just a test to add but otherwise looks good. One thing to maybe double-check is what happens with the existing people page for allocations? Not sure if backward compatible, but worth checking.

components/Allocations/SelectWorker/SelectWorker.spec.tsx
mapleleafjack: Test added!

components/Allocations/SelectWorker/SelectWorker.tsx
mapleleafjack: Added function signature

acodeninja (approved):
Looking good!

PR #890, opened 2022-03-25
fix: historic notes

jhackett1:
> and, following on from the issue beverley spotted just moments ago: this is related to "historic" case notes, and we're investigating. we'll post here again as soon as we have a solution in place. in the meantime, you can still access the content of the notes via the "printable version" link in the popup.

> I know that there was some user acceptance testing of the resident view but I need to feedback that the casenotes/records are not displayed in a very easy to navigate way. You still have to go back and forth in a record to try and obtain information and some of the records are not visible (until clicking on the printable option).

https://hackit-lbh.slack.com/archives/C021Q88C953/p1648140239855699


acodeninja (approved):
Good work!

PR #892, opened 2022-03-25
fix: Fix parameter names allocations

mapleleafjack:
**What**  
This PR fixes the values sent by the frontend for the `ragRating` parameter

**Why**  
I'm sending colours, BE is expecting keywords

**Anything else?**


acodeninja (approved):
All good

PR #893, opened 2022-03-28
fix: Fixes for allocations

mapleleafjack:
**What**  

- Fix redirect to people view on deallocation
- Fix redirect to old person view when editing rag rating.
- Fix setting teamID and allocation date in the url when allocating a worker to an existing team allocation
- If worker not set for allocation do not make get request to get worker details (the request is for the Workers in the team)
- Links in blue 
- Alignment of “deallocate worker“

**Why**  
To complete the Worker Allocation flow


rachel-hutchinson (approved):
Looks good to me

PR #894, opened 2022-03-28
chore: only proper people can change restrictions

jhackett1:
prevents a simple privilege escalation attack where a non-admin user could change a restricted resident to unrestricted.

now, the restricted control is read-only unless the user is sufficiently powerful

### DIFF
--- a/pagesTest/residents/[id]/index.spec.tsx
+++ b/pagesTest/residents/[id]/index.spec.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { render, screen, queryByRole } from '@testing-library/react';
 import { mockedResident } from 'factories/residents';
 import ResidentPage from 'pages/residents/[id]';
 import { useCases } from 'utils/api/cases';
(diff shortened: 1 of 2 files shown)


acodeninja (approved):
LGTM, I'd avoid merging until we have team allocations out of the way though.

pagesTest/residents/[id]/index.spec.tsx
```diff
@@ -66,4 +66,19 @@ describe('ResidentPage', () => {
 
     expect(screen.getByText('This resident has no case notes yet.'));
   });
+
+  it("doesn't let normal users change restricted status", () => {
```
Comment (AlexHerbertMadeTech) [inline pagesTest/residents/[id]/index.spec.tsx:70]: "Doesn't let unauthorised users change restricted status"

Also I feel that there should be a test showing the opposite case, that a user with permission is able to make a change

AlexHerbertMadeTech (commented):
Just one small comment on the tests for this one.

Also I love the implication of the name of this change - you're not a Proper Person until you can change unrestrict case notes in Hackney Adult Social Care. Anyone else is legally a child.

PR #895, opened 2022-03-28
chore: strip html tags from historic notes

jhackett1:
historic case notes have some html tags in their content.

they're no longer needed, so we strip them out here.

### DIFF
--- a/components/ResidentPage/CaseNoteDialog.spec.tsx
+++ b/components/ResidentPage/CaseNoteDialog.spec.tsx
@@ -21,7 +21,10 @@ jest.mock('utils/api/submissions');
   data: mockedCaseNote,
 });
 (useHistoricCaseNote as jest.Mock).mockReturnValue({
-  data: mockedHistoricCaseNote,
+  data: {
+    ...mockedHistoricCaseNote,
+    content: '<h1>foo historic</h1>',
+  },
 });
 (useSubmission as jest.Mock).mockReturnValue({
   data: {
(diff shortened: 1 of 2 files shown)


components/ResidentPage/CaseNoteDialog.spec.tsx
```diff
@@ -138,6 +141,33 @@ describe('CaseNoteDialog', () => {
...
+              ...mockedCaseNote.caseFormData,
+              is_historical: true,
+            },
+          },
+        ]}
+      />
+    );
+
+    expect(screen.getByText('foo historic'));
+    expect(screen.queryByText('<h1>foo historic</h1>')).toBeNull;
```
acodeninja, line 168: `.toBeNull()` ?

acodeninja: Also just wondering if this actually works? Thinking that react will render it as `&lt;h1&gt;foo historic&lt;/h1&gt;` on the back?

components/ResidentPage/CaseNoteDialog.spec.tsx
```diff
@@ -138,6 +141,33 @@ describe('CaseNoteDialog', () => {
...
+            caseFormData: {
+              ...mockedCaseNote.caseFormData,
+              is_historical: true,
+            },
+          },
+        ]}
+      />
+    );
+
+    expect(screen.getByText('foo historic'));
```
acodeninja, line 167: Maybe add `.toBeVisible()`?

acodeninja (commented):
Small alteration to the test, but otherwise good.

components/ResidentPage/CaseNoteDialog.spec.tsx
jhackett1: pretty sure they're functionally identical since getByText [throws an error](https://testing-library.com/docs/react-testing-library/cheatsheet/) if the thing isn't found

components/ResidentPage/CaseNoteDialog.spec.tsx
jhackett1: fixed typo

AlexHerbertMadeTech (approved):
Might be nice if you put some other HTML tags in for the test just to be sure, but this still looks fine.

PR #899, opened 2022-03-28
fix: Removed FE filter for closed allocations

mapleleafjack:
**What**  
This PR removes an unnecessary Allocations at FE API level.
The correct URL parameter is now sent as an URL parameter in the request

**Why**  
The Allocations loaded in the FE are filtered by "closed" status and "allocationEndDate".
This is an issue because the BE is sending only 20 rows (not all) and the FE is doing the filtering, so it's not entirely clear what's going on "under the hood"

**Anything else?**
This filter should be in the BE, and the 20 rows limit removed.
This PR should be merged after that happens.


acodeninja (approved):
LGTM

PR #900, opened 2022-03-28
feat: RagRating for legacy allocations

mapleleafjack:
**What**  
This PR lets us add a priority level for legacy allocations (created without ragRating in the previous versions)

Screenshot 2022-03-28 at 16 25 17

**Why**  
Before this there was no way on editing a ragRating for legacy applications


acodeninja (approved):
Looking good

PR #901, opened 2022-03-28
fix: Disable flaky e2e tests

mapleleafjack:
**What**  
E2E tests for Warning Notes are flaky and prone to error.
In particular those tests are looking for warning note creation/update/end strings in the timeline, but as time passes, those strings are buried in the timeline and the tests fails.

Tests has been disabled.

**Why**  
This is because they test only for the presence of the Warning Note strings.
E2E tests should test the whole flow, not only the presence of a string.


acodeninja (approved):
Get it outta here!

PR #902, opened 2022-03-28
fix: fix arrow nav

jhackett1:
https://trello.com/c/Dwn7mQiF/895-case-notes-arrows-get-stuck-on-non-case-notes

> this happens because workflows don't render a big chunk of the content inside the dialog, which is what the user was focused on. therefore focus is no longer "inside" the dialog when moving to a workflow, therefore the keyboard events aren't caught.
> we can prove this by clicking into the dialog of a workflow directly, or clicking on it after sliding into it with the keyboard. the arrows will then work again.

fixes it by moving the event listener off the dialog and onto the document element

### DIFF
--- a/components/ResidentPage/CaseNoteDialog.tsx
+++ b/components/ResidentPage/CaseNoteDialog.tsx
@@ -13,7 +13,7 @@ import { useCase, useCases, useHistoricCaseNote } from 'utils/api/cases';
 import { useSubmission } from 'utils/api/submissions';
 import FlexibleAnswers from 'components/FlexibleAnswers/FlexibleAnswers';
 import Link from 'next/link';
-import { KeyboardEventHandler, useState } from 'react';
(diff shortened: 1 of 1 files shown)


components/ResidentPage/CaseNoteDialog.tsx
```diff
@@ -136,21 +136,31 @@ const CaseNoteDialog = ({
...
+        newId = caseNotes?.[i - 1]?.recordId; // previous/newer note
       }
-    }
-    if (newId)
-      replace(`${window.location.pathname}?case_note=${newId}`, undefined, {
-        scroll: false,
-      });
-  };
+      if (e.key === 'ArrowRight') {
+        console.log('->');
```
AlexHerbertMadeTech, line 147: Is this console log needed?

AlexHerbertMadeTech (approved):
Looks good to me

PR #904, opened 2022-03-29
fix: allocation summary in header

mapleleafjack:
**What**  
This PR fixes the displayed allocation in the header.
Code has been optimised as it was copy/pasted multiple time
Tests added

Screenshot 2022-03-29 at 10 37 20

**Why**  
It was not supporting team-only allocation, so it was saying "Allocated to undefined" 

**Anything else?**


acodeninja (approved):
🪂

PR #905, opened 2022-03-29
feat: Added deallocationScope to dealllocation request object

mapleleafjack:
**What**  
This PR adds a `deallocationScope` parameter in the deallocateAllocation request.
This parameter can be either `worker` or `team` and it's placed in the body of the request

**Why**  
The BE will differentiate what to deallocate based on this parameter

**Anything else?**


acodeninja (approved):
🥼

PR #907, opened 2022-03-29
fix(SCT-1795): Added necessary parameters for children allocations/deallocations

mapleleafjack:
**What**  
This PR adds the necessary parameters for Children to add and remove allocations from the /people/{id}/allocations page

**Why**  
The structure of the (de)Allocation request in the BE has changed, but the /people/{id}/allocations form doesn't work as it uses the old request object.

**Anything else?**


acodeninja (approved):
🪴

PR #908, opened 2022-03-30
feat: Allocations E2E tests

mapleleafjack:
**What**  
Added E2E tests for the new Allocation flow
The tests will:
-Add a Team / Worker allocation
-Change the priority level
-Remove the Worker allocation
-Remove the Team allocation

The user on which the test will point is the ID `7`, the team will be `testing-team`

+ Added styles and ID for easier testing

**Why**  
This ensures that the feature works as intended

**Anything else?**


acodeninja (approved):
Looks great, thanks for sorting this.

PR #909, opened 2022-03-30
fix: Updating inline editor to use new PATCH endpoint

rachel-hutchinson:
**What**  
Updated inline editor to use a new PATCH endpoint
This will now only send edited information to the PATCH endpoint

**Why**  
We want to make use of a new PATCH endpoint that will update only edited information so that invalid fields do not prevent updating details


jhackett1 (approved):

acodeninja (changes requested):
Type errors need solving.

acodeninja (approved):
LGTM

PR #912, opened 2022-03-30
chore: turns google doc urls into links

jhackett1:
satisfies https://trello.com/c/vJzhNSpb/910-link-text-to-google-docs-in-case-note-popup-should-be-actual-links

### DIFF
--- a/components/ResidentPage/CaseNoteDialog.tsx
+++ b/components/ResidentPage/CaseNoteDialog.tsx
@@ -56,18 +56,12 @@ interface HistoricContentProps {
 const prettyKey = (key: string): string =>
   key?.replace(/_/g, ' ')?.replace(/^\w/, (char) => char.toUpperCase());
 
-const PrettyValue = ({ value }: { value: string }): React.ReactElement => {
-  if (
-    // detect urls
-    // eslint-disable-next-line no-useless-escape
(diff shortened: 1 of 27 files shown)


components/ResidentPage/CaseNoteDialog.tsx
```diff
@@ -56,6 +56,19 @@ interface HistoricContentProps {
 const prettyKey = (key: string): string =>
   key?.replace(/_/g, ' ')?.replace(/^\w/, (char) => char.toUpperCase());
 
+const PrettyValue = ({ value }: { value: string }): React.ReactElement => {
+  if (
+    // detect urls
+    // eslint-disable-next-line no-useless-escape
+    /((?:(?:http?|ftp)[s]*:\/\/)?[a-z0-9-%\/\&=?\.]+\.[a-z]{2,4}\/?([^\s<>\#%"\,\{\}\\|\\\^\[\]`]+)?)/gi.test(
```
AlexHerbertMadeTech, line 63: Not sure about this regex, I see FTP in there. Will this turn every case note link into a clickable?

jhackett1: i copied it from a highly starred stackoverflow answer. it's supposed to turn anything in the case form data that could be a url into one

AlexHerbertMadeTech: Yes, but is turning _any_ URL into something clickable the best th ing for this? As a security minded person, I have a feeling that it's a bad idea to allow users to put whatever in here and also make any link clickabl;e without caring what it's linking to, especially when there are a lot of less computer literate users about.

AlexHerbertMadeTech (approved):
LGTM

PR #915, opened 2022-03-30
feature: children, stop!

jhackett1:
sticks an obstinate dialog in front of all child residents that users have to dismiss.

very annoying, my worst work yet.

### DIFF
--- a/components/StopDialog.tsx
+++ b/components/StopDialog.tsx
@@ -1,0 +1,45 @@
+import { useEffect, useState } from 'react';
+import Dialog from 'components/Dialog/Dialog';
+import s from 'stylesheets/StopDialog.module.scss';
+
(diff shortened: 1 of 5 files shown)


components/StopDialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+        title="You shouldn't make edits here any more"
+        isOpen={true}
+        onDismiss={() => setOpen(false)}
+        showCloseButton={false}
+        className={s.stopDialog}
+      >
+        <>
+          <p>
+            This is a child resident, so you must only use Mosaic for them from
+            now on.
```
acodeninja, line 26: Should we maybe also include in the copy that the record may be out of date and should not be relied upon?

acodeninja (approved):
Looks good, though we might want to address my comment before rolling.

PR #916, opened 2022-03-31
chore: don't warn on unsaved changes

jhackett1:
it's annoying people more than it's helping, so let's not

only affects new resident view + inline editing


acodeninja (commented):
That seems reasonable, I take it there were no tests covering this. Do you mind linking to the ticket this covers? If you get that in I'll approve.

martapederiva (approved):
🔥

PR #921, opened 2022-04-08
fix: Fix case note tags

rachel-hutchinson:
**What**  
Refactored the way that replacing special characters in case notes work. There is a new formatter function that has been added and this is then used in the various case note scenarios.

**Why**  
Previous work only covered removing special characters from historical case notes, this should work for all case notes


mapleleafjack (approved):
tidyText(approved)

PR #922, opened 2022-04-12
feat(SCT-1813): Allocation link in header

mapleleafjack:
This PR needs to be merged AFTER https://github.com/LBHackney-IT/lbh-social-care-frontend/pull/873, as some of the new links points to these pages

**What**  
This PR adds a component that displays links to the team and allocation pages.

Screenshot 2022-04-12 at 16 25 10
Screenshot 2022-04-12 at 16 24 49
Screenshot 2022-04-12 at 16 24 41

**Why**  
To complete the UI link/loop for the allocations

**Anything else?**


components/ResidentPage/AllocationHeader.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+            allocatedWorker: 'Jon Doe',
+            allocatedWorkerTeam: undefined,
+          }),
+          allocationFactory.build({ allocatedWorker: 'Foo Bar' }),
+          allocationFactory.build({ allocatedWorker: 'Agent Smith' }),
+        ]}
+      />
+    );
+    expect(screen.getByTestId('allocationHeader')).toHaveTextContent(
+      'Allocated to Jon Doe and 2 other'
```
acodeninja, line 73: Just a tiny language thing, should be 2 others? Looking at the code in the component it should be rendering it as others anyway?

components/ResidentPage/AllocationHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+    </Link>
+  );
+
+  const moreLink = allocations?.length > 1 && (
+    <Link href={`/residents/${allocations[0].personId}/allocations`}>
+      <a
+        data-testid="allocationLink"
+        className={`lbh-link lbh-link--no-visited-state`}
+      >
+        {allocations?.length === 2
```
acodeninja, line 28: Should this be `allocations?.length =< 2`?

acodeninja (changes requested):
Small change around pluralisation to do.

components/ResidentPage/AllocationHeader.tsx
mapleleafjack: It should be correct, as if the allocations are 2 we display the first one with long description + "one other"

If it's more than 2 (the "else") it says the length -1 (so, not counting the one with long description)

components/ResidentPage/AllocationHeader.spec.tsx
mapleleafjack: changed from other -> others

acodeninja (approved):
Sweet! Thanks for sorting that.

PR #924, opened 2022-04-13
fix: Fixes for Worker and Team allocation

mapleleafjack:
**What**  

- When an allocation is created the same day, the time difference was calculated from midnight. Now it displays "Today"
- The links now points to the allocation page, rather than the main resident
- Adjusted the text in Worker Allocation so it displays "Date allocated" instead of the redundant worker name
- Colour of small text is now  #585d60;
- Teams link is now always in non-visited state
- Added a bottom margin of 35px to the team name
- Added a top margin to the first element of the worker list

Now displays "today"
Screenshot 2022-04-13 at 15 38 33

Displays "Date allocated" instead of the worker name in Worker Allocation view
Colour is now #585d60
Screenshot 2022-04-13 at 15 37 19

Teams link is now always in non-visited state
Bottom margin of 35px to the team name
Screenshot 2022-04-13 at 15 44 41

Added a top margin to the first element of the worker list
Screenshot 2022-04-13 at 15 47 58

**Anything else?**

### DIFF
--- a/components/TeamPage/Layout.tsx
+++ b/components/TeamPage/Layout.tsx
@@ -60,13 +60,18 @@ const TeamLayout = ({ team, children }: Props): React.ReactElement => {
     <div className="govuk-breadcrumbs lbh-breadcrumbs lbh-container">
       <ol className="govuk-breadcrumbs__list">
         <li className="govuk-breadcrumbs__list-item">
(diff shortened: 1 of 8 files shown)


martapederiva (approved):
LGTM

PR #925, opened 2022-04-14
feat: unfinished case notes

AlexHerbertMadeTech:
**What**  
Adds unfinished case notes to the new Resident view

**Why**  
This replicates functionality that existed in the old People view

**Anything else?**

Also adds a line to prettier config. As a Windows user my code is checked out with windows style line endings but gets pushed as normal-style line endings. Prettier kept getting annoyed because it doesn't like Windows style endings.
The change to the prettier file stops Windows users from having prettier errors at the end of every line because it doesn't like the CRLF line endings that Windows defaults to.


acodeninja (approved):
All seems very reasonable.

rachel-hutchinson (approved):
Looks good to me

PR #931, opened 2022-04-20
fix: Redirect people view to residents view

jpmaHub:
**What**  
Redirecting the app to residents page from the people page

**Why**  
people page is deprecated


rachel-hutchinson (commented):
These changes look good, would it be possible to add in some unit tests to confirm they updated buttons are routing to the new resident view?

Miles-Alford (approved):
👍

PR #933, opened 2022-04-21
fix(SCT-1826): Activity list bugs/tweaks

mapleleafjack:
**What**  
API changes:

-Adjust the FE to load the allocation by ID when loading a timeline allocation element
-Changes the status parameter sent  when showOnlyOpen is set to false from "closed" to undefined, as the meaning of the boolean is "load all" and not "load only the closed ones"

UI changes:

-Check nextCursor existence and hide the button in case it’s null
-Add a spinner when loading data
-Center the Load More button

**Why**  
The Activity list stopped working properly because of the new Allocation system.
The UI was also not behaving properly

**Anything else?**

### DIFF
--- a/lib/allocatedWorkers.spec.ts
+++ b/lib/allocatedWorkers.spec.ts
@@ -42,13 +42,13 @@ describe('allocatedWorkersAPI', () => {
           ],
         },
       });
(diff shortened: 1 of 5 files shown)


rachel-hutchinson (approved):
Looks good to me

PR #936, opened 2022-04-25
fix(SCT-1828): Sorting the teams in alphabetical order

martapederiva:
**What**  
In the Manage Workers tool, the teams were presented without sorting.
This PR introduces sorting in alphabetical order.

**Why**  
For clarity and because it's nicer this way ✨ 

**Anything else?**


pages/workers/[id]/edit/[[...stepId]].tsx
```diff
@@ -33,6 +33,13 @@ const UpdateWorker = (): React.ReactElement => {
   if (!ATeams || !CTeams || !data) {
     return <Spinner />;
   }
+  ATeams.sort(function (a, b) {
+    return a.name.localeCompare(b.name);
+  });
```
Comment (rachel-hutchinson) [inline pages/workers/[id]/edit/[[...stepId]].tsx:38]: This looks good, one suggestion that I have is that the sorting could be done in one line, using an arrow function:
`ATeams.sort((a, b) => a.name.localeCompare(b.name));`

mapleleafjack (approved):
I've added a test to make sure the pages does it's job properly - feel free to have a look at it before merging :)

PR #940, opened 2022-05-03
fix(SCT-1848): Allocation count

mapleleafjack:
**What**  
This PR fixes the bug in which the allocation count is limited to the paginated response.
This goes together with SCT-1849 which adds the parameter to the response from the backend

**Anything else?**


acodeninja (approved):
LGTM

PR #947, opened 2022-05-05
fix(SCT-1851): Don't display the date if it's the dummy imported one

mapleleafjack:
**What**  
While importing data, we set some of the dates to 01-01-0001 as they were required by the system.
This PR hides that date from the UI as it doesn't make sense to display it.

Screenshot 2022-05-05 at 09 56 47

**Anything else?**


acodeninja (approved):
Looks good.

PR #949, opened 2022-05-05
chore: Stop opening of case-note dialog on resident details page

Miles-Alford:
**What**  
This PR stops the keyboard navigation from opening the case notes dialog on the resident details by only updating the path to open a case note on the case-notes page on the resident view.
**Why**  
This is to stop users from being able to opening the case note dialog on any other page beside the case-notes page on the resident view.

**Anything else?**
Any suggestions to improve this would be helpful, I am not very knowledgeable on how events work.


acodeninja (approved):
Sweet!

PR #950, opened 2022-05-06
fix: ERR_STREAM_WRITE_AFTER_END thrown on unauthenticated API requests

acodeninja:
**What**  
Removes calls to `.end()` from returned streams on unauthenticated requests.

**Why**  
This should fix the issue where errors are being posted after a 40X response from the API.

**Anything else?**


mapleleafjack (approved):
.approved()

acodeninja:
> .approved()

Don't you mean `.approved()`?

PR #951, opened 2022-05-09
feat(SCT-1857): CSRF token

mapleleafjack:
**What**  
This PR adds the necessary bits for setting the CSRF token in requests.
In particular:

- Added npm package CSRF
- Implemented csrfToken.ts
- CSRF token set it in the header and in the meta tag in _document.tsx
- AXIOS interceptors set in _app.tsx
- Every API request wrapped with apiHandler(csrfMiddleware(handler))
- Every fetch request changed to csrfFetch
- added CRSF_TOKEN definition in serverless.yaml

**Why**  
To make the app more secure

**Anything else?**

- Created /lbh-social-care/staging/csrf-secret in AWS Staging Parameter Store
- Created /lbh-social-care/mosaic-prod/csrf-secret in AWS Prod Parameter Store
- Added csrf-secret (staging) to CircleCI
- added CSRF_TOKEN in .env file


lib/csrfToken.test.ts
```diff
@@ -0,0 +1,109 @@
+import { CSRFValidationError, init } from './csrfToken';
```
acodeninja, line 1: Does this need to be a `.spec.ts` file? I'm not sure if `.test.ts` files are picked up?

mapleleafjack: Refactored :)

lib/auth/test-functions.tsx
```diff
@@ -0,0 +1,92 @@
...
+  body?: unknown;
+  headers?: {
+    [key: string]: string;
+  };
+  cookies?: {
+    [key: string]: string;
+  };
+}
+
+export const makeNextApiRequest = ({
```
acodeninja, line 53: I think this is the only method needed from this file, might be a good refactor to move it into the spec file for csrf tokens and get rid of the rest of the file.

acodeninja (approved):
Sweet!

PR #952, opened 2022-05-10
fix: AppConfigProvider errors on static pages

acodeninja:
**What**  
Only throw an `AppConfigProvider` error when getting a config value, rather than when calling `useAppConfig()`.

**Why**  
Errors are thrown when static pages are loaded, even though they never actually pull data from the `AppConfigProvider`.

**Anything else?**


mapleleafjack (approved):
🚀

PR #956, opened 2022-05-12
fix: Csrf token for Case Notes

mapleleafjack:
**What**  
This PR adds the XSRF-TOKEN for all the calls in Case Notes that don't use the standard API routes.

**Why**  
The offended rows were not implementing the token, so an error was returned

**Anything else?**


acodeninja (approved):
Looks good.

